PR #124, opened 2023-10-23
Adds scaffolding for publishing

ncordon:
This PR brings a lot of changes to make the package ready for publishing:

* Adds a LICENSE file to each subproject, as that will be needed.
* Fixes the VSCode extension debugging. This has been done by:
   * Using a different language server (a non bundled one that loads the language support package code when used) for debugging and a bundled one for run.
   * Enabling source maps in the `.vscode/launch` file and giving them the paths of all source maps inside language support and schema-poller.
* Enables TS incremental compilation. This enables to recompile projects very quickly after the first run with `npm run build`, unless you run a clean.
* Bundles antlr4 and antlr4-c3 inside the language-support package and uses esbuild whenever possible or necessary with other packages.
* Adds the typings for react-codemirror.
* Configures the exports so all the projects work both with esm and commonjs. 
   * I've tried publishing the language-support package locally, installing it into a separate project with npm link and using it both as commonjs and esm, it works ✅ .
* Cleans up the libraries and aligns the projects on Typescript 4.9.5. We have 0 library vulnerabilities now, we used to have some.

* Installs `changeset` and aligns all versions on a 2.0.0-next.0 release.
* Changes the VSCode extension settings from `CypherLSP.neo4j` to just `neo4j` namespace, which looks cleaner:
```
    "neo4j.connect": true,
    "neo4j.password": "pass12345",
    "neo4j.user": "neo4j"
```

## Caveats
* I don't know why but the `package-lock.json` has a lot of changes. I upgraded my npm locally to 9.5.0 so maybe that is caused a lot of them.


.vscode/launch.json
```diff
@@ -16,12 +16,13 @@
...
+        "${workspaceRoot}/packages/vscode-extension/**/*.js",
+        "${workspaceRoot}/packages/language-support/**/*.cjs",
+        "${workspaceRoot}/packages/schema-poller/**/*.cjs"
+      ],
       "autoAttachChildProcesses": true,
-      "preLaunchTask": {
-        "type": "npm",
-        "script": "watch"
-      }
+      "sourceMaps": true
```
ncordon, line 25: This fixes the VSCode debugging

package.json
```diff
@@ -34,6 +35,6 @@
     "prettier-plugin-organize-imports": "^3.2.1",
     "ts-jest": "^29.1.1",
     "turbo": "^1.10.15",
-    "typescript": "^4.8.4"
+    "typescript": "4.9.5"
```
ncordon, line 38: Aligns the same typescript version everywhere

.changeset/pre.json
```diff
@@ -0,0 +1,15 @@
+{
+  "mode": "pre",
+  "tag": "next",
```
ncordon, line 3: I don't know how it works but my intuition tells me we should be using this next tag instead of the alpha one? At least that's what they even describe in the changesets documentation

packages/language-support/package.json
```diff
@@ -14,11 +22,11 @@
   },
   "scripts": {
     "gen-parser": "antlr4 -Dlanguage=TypeScript -visitor src/antlr-grammar/CypherParser.g4 src/antlr-grammar/CypherLexer.g4 -o src/generated-parser/ -Xexact-output-dir",
-    "postinstall": "npm run gen-parser",
```
ncordon, line 17: We don't need this as far as I know. The generated parser should be included in the package

packages/language-support/package.json
```diff
@@ -1,11 +1,19 @@
...
   "description": "Cypher language support",
-  "author": "neo4j",
+  "author": "Neo4j",
   "license": "Apache-2.0",
-  "version": "0.0.1",
-  "main": "./dist/index.js",
-  "module": "./esm/index.js",
+  "version": "2.0.0-next.0",
+  "main": "./dist/cjs/index.cjs",
+  "moudle": "./dist/esm/index.mjs",
```
ncordon, line 8: mispelt

vendor/antlr4-c3/deps
```diff
@@ -0,0 +1,1606 @@
+@neo4j-cypher@1.0.0 /Users/ncordon/neo4j/intellisense/cypher-lsp
```
ncordon, line 1: Remove this file

vendor/antlr4-c3/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
     "declaration": true,
-    "target": "ES2022",
+    "target": "ESNext",
```
ncordon, line 4: Research whether we should use ESNext vs es2020 for example

packages/vscode-extension/src/extension.ts
```diff
@@ -12,7 +12,7 @@ let client: LanguageClient;
 export function activate(context: ExtensionContext) {
   // The server is implemented in node
   const serverModule = context.asAbsolutePath(
-    path.join('..', 'language-server', 'dist', 'cypher-language-server.js'),
+    path.join('..', 'language-server', 'dist', 'server.js'),
```
ncordon, line 15: If we use the bundled one here we won't be able to debug anything.

packages/language-server/package.json
```diff
@@ -1,23 +1,23 @@
 {
   "name": "@neo4j-cypher/language-server",
   "description": "Cypher Language Server",
-  "author": "neo4j",
+  "author": "Neo4j",
   "license": "Apache-2.0",
-  "version": "0.0.1",
-  "main": "dist/cypher-language-server.js",
+  "version": "2.0.0-next.0",
+  "main": "./dist/server.js",
```
ncordon, line 7: This is for VSCode debugging to work

.vscode/launch.json
```diff
@@ -16,12 +16,13 @@
...
+      "outFiles": [
+        "${workspaceRoot}/packages/vscode-extension/**/*.js",
+        "${workspaceRoot}/packages/language-support/**/*.cjs",
+        "${workspaceRoot}/packages/schema-poller/**/*.cjs"
+      ],
       "autoAttachChildProcesses": true,
-      "preLaunchTask": {
-        "type": "npm",
-        "script": "watch"
-      }
```
ncordon, line 24: Not sure whether we need the watch but it seems broken to me locally

packages/vscode-extension/src/extension.ts
```diff
@@ -11,16 +11,19 @@ let client: LanguageClient;
 
 export function activate(context: ExtensionContext) {
   // The server is implemented in node
-  const serverModule = context.asAbsolutePath(
-    path.join('..', 'language-server', 'dist', 'cypher-language-server.js'),
+  const runServer = context.asAbsolutePath(
+    path.join('dist', 'cypher-language-server.js'),
+  );
+  const debugServer = context.asAbsolutePath(
+    path.join('..', 'language-server', 'dist', 'server.js'),
```
ncordon, line 18: Use a different configuration for debug and for running (once the extension is compiled, the language server has to be living inside it, we cannot use a relative path). Also we want that language server to be minified.

tsconfig.base.json
```diff
@@ -9,7 +9,8 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "target": "ESNext",
-    "outDir": "dist"
+    "outDir": "dist",
+    "incremental": true
```
ncordon, line 13: Enables incremental compilation

packages/vscode-extension/package.json
```diff
@@ -1,17 +1,18 @@
 {
-  "name": "@neo4j-cypher/vscode-extension",
-  "description": "Neo4j's Cypher VSCode extension",
-  "author": "neo4j",
+  "name": "neo4j-cypher-vscode-extension",
```
ncordon, line 2: The command to publish the VSCode extension doesn't like this name either

OskarDamkjaer (approved):

PR #136, opened 2023-11-07
Make editor history controlled and `onExecute` reactive

OskarDamkjaer:
We need more control of what is in the editor history in query, so I've moved it to be externally controlled. If we see the need for uncontrolled history, we can add an `uncontrolledHistory` prop to make `onExecute` push to history again.

### DIFF
--- a/.changeset/quick-berries-report.md
+++ b/.changeset/quick-berries-report.md
@@ -1,0 +1,5 @@
+---
+'@neo4j-cypher/react-codemirror': patch
+---
+
+Editor history is now a controlled prop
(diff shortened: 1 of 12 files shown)


.changeset/quick-berries-report.md
```diff
@@ -0,0 +1,5 @@
+---
+'@neo4j-cypher/react-codemirror': patch
+---
+
+Editor history is now a controlled prop
```
ncordon, line 5: Can we also push the package.json bump alongside this please?

ncordon (approved):
LGTM. You are just missing doing `npx changeset version` I think

PR #138, opened 2023-11-10
Fix incorrect completions when identifier overlaps with keyword

OskarDamkjaer:
We have logic for knowing when to complete the next vs current word. Our logic only moved the cursor back for identifier, so it didn't work when there was overlap between identfiers and keywords. For example:

`SHOW DAT` correctly suggests `DATABASE`
but
`SHOW DATA` would not.

### DIFF
--- a/.changeset/stale-donkeys-grow.md
+++ b/.changeset/stale-donkeys-grow.md
@@ -1,0 +1,5 @@
+---
+'@neo4j-cypher/language-support': patch
+---
+
+Fix incorrect completions when identifier overlaps with keyword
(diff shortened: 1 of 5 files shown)


ncordon (approved):

.changeset/stale-donkeys-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@neo4j-cypher/language-support': patch
+---
+
+Fix incorrect completions when identifier overlaps with keyword
```
ncordon, line 5: Only comment is this should not be included in the pr I think and we should bunch them up before doing a release? Rather than creating this patch.

OskarDamkjaer: This will not trigger a release, only add a changeset right? It's nice to add the changeset relating to the commit at the same time we do the commit IMO

PR #139, opened 2023-11-10
Fix issue where syntax highlighting crashes on create constraint query

OskarDamkjaer:


ncordon (approved):

.changeset/purple-suns-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'@neo4j-cypher/language-support': patch
+---
+
+Fix issue where syntax highlighting crashes on create constraint query
```
ncordon, line 5: I think we should leave this file out of the PR and same for #138. Because we have a patch with a different file name in both so I don't know what's the expected behaviour in that case

PR #144, opened 2023-11-22
improve history prop change detection

OskarDamkjaer:
The shortcut of comparing lengths works fine until you hit the maximum history length in workspace -> every new entry pushes an old one out, meaning the length of history entries is constant. I now also compare the newest history entry

### DIFF
--- a/packages/react-codemirror/src/CypherEditor.tsx
+++ b/packages/react-codemirror/src/CypherEditor.tsx
@@ -302,7 +302,13 @@ export class CypherEditor extends Component<CypherEditorProps> {
       });
     }
 
-    if (prevProps.history?.length !== this.props.history?.length) {
+    // This component rerenders on every keystroke and comparing the
+    // full lists of editor strings on every render could be expensive.
+    const didChangeHistoryEstimate =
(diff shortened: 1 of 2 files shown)


ncordon (approved):

packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -302,7 +302,13 @@ export class CypherEditor extends Component<CypherEditorProps> {
       });
     }
 
-    if (prevProps.history?.length !== this.props.history?.length) {
+    // This component rerenders on every keystroke and comparing the
+    // full lists of editor strings on every render could be expensive.
+    const didChangeHistoryEstimate =
+      prevProps.history?.length !== this.props.history?.length ||
```
ncordon, line 308: One question. Can this array grow indefinitely?

OskarDamkjaer: It's up to the consumer, in workspace we currently limit it to 300. In browser it's configurable by the user. Some of the query strings can also be quite large

PR #145, opened 2023-11-22
Adds instructions on how to maintain the semantic analysis

ncordon:


OskarDamkjaer (approved):
The git patch is a neat way to re-apply the same change, good idea!

PR #149, opened 2023-11-29
Adds notifications to the semantic analysis

ncordon:
## What

* Adds notifications to the semantic analysis. All of the notifications will be surfaced as warnings.
* Provides a better surface to access the semantic analysis.

### DIFF
--- a/packages/language-support/src/highlighting/syntaxValidation/syntaxValidation.ts
+++ b/packages/language-support/src/highlighting/syntaxValidation/syntaxValidation.ts
@@ -8,7 +8,10 @@ import {
   LabelType,
   parserWrapper,
 } from '../../parserWrapper';
-import { doSemanticAnalysis } from './semanticAnalysisWrapper';
+import {
+  doSemanticAnalysis,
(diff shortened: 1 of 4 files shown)


packages/language-support/src/highlighting/syntaxValidation/syntaxValidation.ts
```diff
@@ -132,33 +169,26 @@ export function validateSyntax(
...
+      );
     }
 
     const warnings = warnOnUndeclaredLabels(parsingResult, dbSchema);
-    diagnostics = errors.concat(warnings);
+    diagnostics = diagnostics.concat(warnings);
   }
 
-  return diagnostics;
+  return sortByPosition(diagnostics);
```
ncordon, line 193: This is to provide a stable ordering. It's not strictly needed but it would be good for testing purposes to have always the same ordering, especially if we change things in the semantic analysis.

OskarDamkjaer (commented):
Good stuff!

OskarDamkjaer (approved):

PR #150, opened 2023-11-30
Codemirror benchmarks

OskarDamkjaer:
This PR includes two new github actions, one running on merged PR that records benchmarkresults to a grafana cloud instance and one job that compares main / pr benchmarks to warn if performance is decreased bellow a set threshold.

Note that the comparative github action will fail on this PR because there's no benchmarking script to compare with yet (as it's not merged yet). I tested the action comparing the PR to PR, which succeeded


.github/workflows/performance-regression.yaml
```diff
@@ -0,0 +1,67 @@
...
+        working-directory: main
+
+      - name: Install dependencies with frozen lock file
+        run: npm ci
+        working-directory: pr
+
+      - name: Build benchmarks
+        env:
+          NODE_OPTIONS: '--max_old_space_size=4096'
+        run: npm run build-benchmark
```
OskarDamkjaer, line 36: this is an outdated script, I'll need to give this another look

packages/language-support/src/parserWrapper.ts
```diff
@@ -220,7 +220,8 @@ class ParserWrapper {
   parse(query: string): EnrichedParsingResult {
     if (
       this.parsingResult !== undefined &&
-      this.parsingResult.query === query
+      this.parsingResult.query === query &&
+      process.env.BENCHMARKING !== 'true'
```
ncordon, line 224: Why do we need this?

ncordon, line 224: I'm not a fan of this cause if someone had this env variable defined (unlikely but still possible) then the behaviour of the language server changes.

I'd rather probably we had a method to clean the cache and we call that after every test of the benchmark?

packages/react-codemirror/src/e2e_tests/performance-test.spec.tsx
```diff
@@ -68,4 +90,58 @@ test('performance test session ', async ({ mount, page }) => {
...
+        },
+      },
+    ).then((res) => {
+      if (res.ok) {
+        // eslint-disable-next-line no-console
+        console.log('Metrics pushed to grafana successfully');
+      } else {
+        throw new Error(`Failed to push metrics to grafana: ${res.statusText}`);
+      }
+    });
```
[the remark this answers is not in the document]
OskarDamkjaer: Since they live in different packages I'd need to either export the helper from `language-support` publicly (which doesn't seem to make a lot of sense IMO) or create some new util package that we don't publish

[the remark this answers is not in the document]
OskarDamkjaer: I can create a util package for it if you like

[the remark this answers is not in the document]
ncordon: We can leave it like this then

packages/language-support/src/parserWrapper.ts
ncordon: I'm flexible, so if you prefer to do that on another PR, we can iterate on this

OskarDamkjaer: I think it'd be nice to do in a follow up PR, just to see that the Github actions work

packages/language-support/src/highlighting/syntaxColouring/syntaxColouring.ts
```diff
@@ -300,3 +300,7 @@ export function applySyntaxColouring(
 
   return result;
 }
+
+export function clearCache() {
+  delete parserWrapper.parsingResult;
```
ncordon, line 305: Shouldn't this call parserWrapper.clearCache instead?

OskarDamkjaer: It got in by mistake, thanks for catching it!

PR #152, opened 2023-12-08
Fixes signature help edge case

ncordon:
## What
Fixes the case of getting the overlay with the information about the first argument for functions.

## How

`RETURN apoc.do.when(` gets parsed as:

```
      statements
     /    |    \     
statement (    EOF
    /  \
RETURN  expression
```

rather than:

```
      statements
     /    |    \     
statement (    EOF
    /  \
RETURN  functionInvocation
```
so we need to treat that case differently because we cannot modify the relative priority of functionInvocation vs expression

`RETURN apoc.do.when(x` gets parsed correctly in contrast (when we've started the first argument), because the parser has enough information to recognize it is a function invocation.

When our final character is `(`, just go to the preceeding statement and traverse it to find whether it ends on a expression. If it does, that's our function name.

### DIFF
--- a/packages/language-support/src/signatureHelp.ts
+++ b/packages/language-support/src/signatureHelp.ts
@@ -62,11 +62,13 @@ RETURN  expression
 rather than:
 
 
-      statements
-     /    |    \     
-statement (    EOF
-    /  \
+        statements
+      /            \     
+statement          EOF
+    /  \ 
 RETURN  functionInvocation
+             /         \ 
+        functionName    (
 
 
 so we need to treat that case differently because we cannot modify the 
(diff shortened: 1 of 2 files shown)


packages/language-support/src/tests/signatureHelp.test.ts
```diff
@@ -164,7 +164,7 @@ describe('Functions signature help', () => {
   //   variable, property, property, spurious (
```
OskarDamkjaer, line 164: we can remove this todo comment now right?

packages/language-support/src/signatureHelp.ts
```diff
@@ -59,13 +118,22 @@ function tryParseFunction(
...
     const numMethodArgs = ctx.functionArgument_list().length;
 
-    return {
+    result = {
       methodName: methodName,
       numProcedureArgs: numMethodArgs,
     };
-  } else {
-    return undefined;
+  } else if (currentNode.getText() === '(') {
```
OskarDamkjaer, line 125: I think it'd be nice if we could move the comment from line 170 here instead to explain why we're looking for the `(` . Ideally it'd be nice to check that the current node is a child of statements here as well but it's fine to leave that check inside the helper fn if you prefer

OskarDamkjaer (approved):
I worry a little bit about false positives with this approach (since expressions are so flexible), but I can't think of any compelling examples or any other solution barring looking into changing the error recovery so I think it makes sense to go ahead

PR #155, opened 2023-12-15
Fixes grafana push (hopefully)

ncordon:


.github/workflows/benchmark.yaml
```diff
@@ -25,6 +25,6 @@ jobs:
         run: npm run build
 
       - name: Benchmark
-        run: npm run benchmark --concurrency 1
+        run: npm run benchmark
```
OskarDamkjaer, line 28:
```suggestion
        run: npm run benchmark --concurrency=1
```
will solve the error. We need them to run one after the other to get more stable performance metrics

OskarDamkjaer (approved):

PR #158, opened 2023-12-21
Decrease input-lag for large queries

OskarDamkjaer:
# Overview
This ended up being quite a bit larger than I had anticipated/hoped for, but the main changes are:
1. Use `workerpool` pkg to create a pool of semantic analysis workers in both codemirror & the server itself. It will cancel outdated and unfinished semanatic analsysis runs when new ones get triggered. I also debounced it for the language server so it fires when the user has stopped typing (chose 600ms somewhat arbitrarily, codemirror has a built in debounce of 750ms for comparison)
2. I detect slow parses in codemirror and switch to using [prismjs](https://prismjs.com/) for tokenization. Prism is a syntax highlighting library that has cypher support and can be used as a crude parser via it's `tokenize` method. In my testing it is about ~20x our parser.

A few future tasks:
[ ] Figure out a smarter way to bundle (see note on it below)
[ ] Reduce duplication of the worker code
[x] Improve react perf by debouncing the on change there as well

# Outdated description below
Since the semantic analysis takes so much longer than the normal errors, I split it to it's own call.

I started by trying to get the same `runSemanticAnalysis` webworker implementation working everywhere with the `web-worker` node package, but it turned out to be really tricky to get it to work in all combinations of environment (jest/web/plain node) and targets (esm/cjs). I ended up leaving the function more or less as it was and then I do an esm only webworker in the react-codemirror package and then I use nodes built in `worker_threads` to get it to work in the server!

## Future improvement - smarter debouncing
When the document is very big, it's easy to queue up a few semantic analysis calls in the webworker (even when debounced) since they can take so long. I think it'd make sense to make the calls into a queue and when one call is finished, all but the latest call are discarded. I carded this here: https://trello.com/c/EvujJSED/219-dont-start-new-semantic-analysis-while-the-old-one-is-still-running-debounce-while-running

## Bundling concerns
We should consider how we can split the bundle like this:
CleanShot 2023-12-21 at 12 07 54

Right now only the worker loads the semantic analysis but it's still in the main bundle. My attempt at splitting failed because it was tricky to configure the worker to do the dynamic import. The other option is if we don't bundle, the client application will use their dead code elimination to make sure it doesn't get in the final bundle - i think this would be easier. I've added a card: worker-import-meta-url


.vscode/settings.json
```diff
@@ -2,7 +2,7 @@
   "typescript.tsc.autoDetect": "off",
   "typescript.preferences.quoteStyle": "single",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
OskarDamkjaer, line 5: the "true" value has been deprecated, this was automatically updated by vscode

packages/language-support/src/highlighting/syntaxValidation/semanticAnalysisWrapper.ts
```diff
@@ -29,7 +31,27 @@ export function doSemanticAnalysis(query: string): SemanticAnalysisResult {
...
       semanticErrorsResult.$notifications.data;
-    return { errors: errors, notifications: notifications };
+
+    return {
+      errors: errors.map(({ message, position }) => ({
+        severity: DiagnosticSeverity.Error,
+        message,
+        position: {
+          column: position.column,
+          line: position.line,
```
OskarDamkjaer, line 41: This gets rid of some $$position objects that could not be passed over the `webworker` <-> `main thread` boundary. I also add the severity here, so we can make the mapping later simpler

packages/language-server/src/worker.ts
```diff
@@ -0,0 +1,9 @@
+import { doSemanticAnalysis } from '@neo4j-cypher/language-support';
+import { parentPort } from 'worker_threads';
+
+parentPort.on('message', (value: { query: string; port: MessagePort }) => {
+  const result = doSemanticAnalysis(value.query);
```
OskarDamkjaer, line 5: It's not a great name, do you have any other suggestions? Maybe something like `getSemanticDiagnostics` or simliar ?

packages/react-codemirror/package.json
```diff
@@ -18,22 +18,15 @@
...
-  },
+  "type": "module",
   "sideEffects": false,
   "scripts": {
-    "build": "concurrently 'npm:build-types' 'npm:build-esm' 'npm:build-commonjs'",
+    "build": "concurrently 'npm:build-types' 'npm:build-esm' 'npm:build-worker'",
     "build-types": "tsc --emitDeclarationOnly --outDir dist/types",
     "build-esm": "esbuild ./src/index.ts --bundle --format=esm --sourcemap --packages=external --outfile=dist/esm/index.mjs",
-    "build-commonjs": "esbuild ./src/index.ts --bundle --format=cjs --sourcemap --packages=external --outfile=dist/cjs/index.cjs",
+    "build-worker": "esbuild ./src/lang-cypher/semantic-analysis-worker.ts --bundle --format=esm --sourcemap --packages=external --outfile=dist/esm/semantic-analysis-worker.mjs",
```
OskarDamkjaer, line 29: note: i removed the cjs build here to get the webworkers to work in a convenient way for the consumer of the package

packages/language-server/package.json
```diff
@@ -35,6 +35,7 @@
   },
   "dependencies": {
     "@neo4j-cypher/language-support": "2.0.0-next.2",
+    "@neo4j-cypher/schema-poller": "2.0.0-next.2",
```
ncordon, line 38: Why is this needed? I don't think this would work with the publishing since this package is not in npm

packages/language-server/src/server.ts
```diff
@@ -86,6 +92,22 @@ documents.onDidChangeContent((change) => {
...
+  port2.on('message', (diagnostics: Diagnostic[]) => {
+    void connection.sendDiagnostics({
+      uri: document.uri,
+      diagnostics: diagnostics,
+    });
+  });
+
+  semanticAnalysisWorker.postMessage({ query, port: port1 }, [port1]);
+});
+
```
ncordon, line 110: I think you cannot have two `onDidChangeContent` because that overwrites it. In other words right now we are not surfacing syntactic errors

packages/language-server/src/worker.ts
```diff
@@ -0,0 +1,9 @@
+import { doSemanticAnalysis } from '@neo4j-cypher/language-support';
+import { parentPort } from 'worker_threads';
+
+parentPort.on('message', (value: { query: string; port: MessagePort }) => {
```
ncordon, line 4: Dummy question. Why do we have a parentPort here and a `self.onmessage` in the react-codemirror package?

Is it due to node vs browser approach?

OskarDamkjaer: yes

packages/language-server/package.json
OskarDamkjaer: my mistake

packages/language-server/src/server.ts
OskarDamkjaer: I merged them back as per our offline discussion

package-lock.json
```diff
@@ -78,18 +74,16 @@
...
+      "license": "MIT",
       "engines": {
         "node": ">=6.9.0"
       }
     },
     "node_modules/@babel/core": {
       "version": "7.23.2",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.23.2.tgz",
-      "integrity": "sha512-n7s51eWdaWZ3vGT2tD4T7J6eJs3QoBXydv7vkUM06Bf1cbVD2Kc2UrkzhiQwobfV7NwOnQXYL7UBJ5VPU+RGoQ==",
       "dev": true,
```
ncordon, line 85: I don't think these changes are needed? Is it because we are using different versions of npm?

ncordon (commented):
I see some files are renamed. It would be good to be consistent in the project in general cause we have camel case ones and hyphened ones

ncordon (approved):
LGTM, just left a couple of non important comments. 

We could also revert this to semantic analysis and syntactic errors separated if you think it worked snappier in that case.

packages/language-server/src/lint-worker.ts
```diff
@@ -0,0 +1,14 @@
+import { validateSemantics } from '@neo4j-cypher/language-support';
```
OskarDamkjaer, line 1: I'm aware these lint workers are now completely duplicated. It would make sense to build the worker in the language support pkg, but then we need some mechanism to copy it out to the other applications I think. I would prefer to that work at a later stage

packages/language-server/src/linting.ts
```diff
@@ -0,0 +1,62 @@
...
+  sendDiagnostics(syntaxErrors);
+
+  if (syntaxErrors.length === 0) {
+    try {
+      if (lastSemanticJob !== undefined && !lastSemanticJob.resolved) {
+        void lastSemanticJob.cancel();
+      }
+
+      const proxyWorker = (await pool.proxy()) as unknown as LintWorker;
+      lastSemanticJob = proxyWorker.validateSemantics(query);
```
OskarDamkjaer, line 41: could pass positions here to avoid doing the the findposition in the main thread

packages/react-codemirror/package.json
```diff
@@ -18,22 +18,14 @@
...
-      "import": "./dist/esm/index.mjs",
-      "default": "./dist/cjs/index.cjs"
-    }
-  },
+  "main": "./src/index.ts",
+  "type": "module",
   "sideEffects": false,
   "scripts": {
-    "build": "concurrently 'npm:build-types' 'npm:build-esm' 'npm:build-commonjs'",
+    "build": "echo done",
```
OskarDamkjaer, line 25: whoops

packages/language-server/src/lint-worker.ts
ncordon: Can you make a cleanup card about this?

packages/react-codemirror/package.json
ncordon: Is this correct?

packages/react-codemirror/src/e2e_tests/syntax-validation.spec.tsx
```diff
@@ -90,6 +90,12 @@ test('Semantic errors are surfaced when there are no syntactic errors', async ({
   page,
   mount,
 }) => {
+  page.on('worker', (worker) => {
+    console.log('Worker created: ' + worker.url());
+    worker.on('close', (worker) =>
+      console.log('Worker destroyed: ' + worker.url()),
+    );
+  });
```
ncordon, line 98: Why do we need this?

packages/react-codemirror/src/lang-cypher/parser-adapter.ts
```diff
@@ -0,0 +1,145 @@
...
+    });
+  }
+
+  private antlrParse(document: string) {
+    const startTime = performance.now();
+    const tokens = applySyntaxColouring(document);
+    const timeTaken = performance.now() - startTime;
+    if (timeTaken > 300) {
+      this.config.setUseLightVersion?.(true);
+    }
```
ncordon, line 79: This sets the light version to true in case the parsing takes more than 300 milliseconds, but we are never doing the opposite right? How do we revert back to the antlr parser (if we do want to ever do that)

packages/language-server/src/linting.ts
```diff
@@ -0,0 +1,63 @@
...
+} from '@neo4j-cypher/language-support';
+import debounce from 'lodash.debounce';
+import { join } from 'path';
+import { Diagnostic, TextDocumentChangeEvent } from 'vscode-languageserver';
+import { TextDocument } from 'vscode-languageserver-textdocument';
+import workerpool from 'workerpool';
+import { LinterTask, LintWorker } from './lint-worker';
+
+const pool = workerpool.pool(join(__dirname, 'lint-worker.js'), {
+  minWorkers: 2,
```
ncordon, line 14: Why do we need 2 workers? I think you explained me this already.

I tried including a `console.log` of the `pool.stats()` at the beginning of the linting here and at the beginning of the semantic linting and I can see 1 occupied worker, 1 idle one at most.


```
{totalWorkers: 2, busyWorkers: 1, idleWorkers: 1, pendingTasks: 0, activeTasks: 1}
```

packages/react-codemirror/package.json
OskarDamkjaer: no, this I need to correct. I only got the workerpool working in codemirror by not bundling. But we still need to build for non-typescript consumers

packages/react-codemirror/src/lang-cypher/parser-adapter.ts
```diff
@@ -0,0 +1,145 @@
...
+    return { type: 'antlr' as const, tokens };
+  }
+
+  private prismParse(document: string) {
+    if (document.length === 0) {
+      this.config.setUseLightVersion?.(false);
+    }
```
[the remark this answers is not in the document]
OskarDamkjaer: Yes. I was originally considering a parse time threshold (if prism parses under Xms it's probably fine with antlr), but that risks some really odd behaviour for outliers. Some queries might be fast with antlr and slow with prism for unexpected reasons and we risk rapid flipping between parsers if the thresholds are poorly set. 

Flipping back when the document is simple to understand, and in the worst case scenario (paste large -> empty all -> paste large again), we're still avoiding the main issue (delay between inputs) as we switch back to prism after the first parse

packages/language-server/src/linting.ts
OskarDamkjaer: The intention is to always have an idle worker available, so that we can avoid the startup times of spawning a new thread. My reasoning is that when one thread is busy, we can use the idle one without startup cost and then the old busy thread will be terminated and respawned so it's ready for the next call

I've not measured the start up time/performance gains though

packages/language-server/src/lint-worker.ts
OskarDamkjaer: https://trello.com/c/s3BckR3v/235-lint-worker-code-is-now-fully-duplicated

ncordon (approved):

packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -183,6 +190,10 @@ export class CypherEditor extends Component<CypherEditorProps> {
     theme: 'light',
   };
 
+  private debouncedOnChange = this.props.onChange
+    ? debounce(this.props.onChange, 200)
+    : undefined;
+
```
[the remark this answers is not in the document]
OskarDamkjaer: Ah, yes - my bad

packages/language-support/src/highlighting/syntaxValidation/semanticAnalysisWrapper.ts
```diff
@@ -29,7 +31,27 @@ export function doSemanticAnalysis(query: string): SemanticAnalysisResult {
     const errors: SemanticAnalysisElement[] = semanticErrorsResult.$errors.data;
```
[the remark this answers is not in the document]
OskarDamkjaer: Ah right, thanks! updated now :+1:

PR #167, opened 2024-01-22
Cut down on benchmarking metrics to improve build speeds

OskarDamkjaer:
Let's set something up in grafana for alerts, rather than build twice on the PR itself :+1:


packages/language-support/src/tests/benchmarks/benchmark.ts
```diff
@@ -18,21 +16,6 @@ Benchmark.options.minSamples = 20;
...
-    applySyntaxColouring(simpleQuery);
-  })
-  .add('simple - validate syntax', function () {
-    parserWrapper.clearCache();
-    validateSyntax(simpleQuery, benchmarkingMediumSizeSchema);
-  })
-  .add('simple - autocomplete next statement', function () {
-    parserWrapper.clearCache();
-    autocomplete(simpleQuery, benchmarkingMediumSizeSchema);
-  })
```
ncordon, line 35: Why are we removing benchmarks? Wouldn't it be better to leave them but just run them on merge to `main`?

OskarDamkjaer: I was in the "reduce gh minute" mindset, so I removed some that didn't feel like they were super impactful, to see if the build got noticeably faster. But you're right, I think we're better off keeping them - i added them back now

PR #169, opened 2024-01-31
Adds more benchmarks

ncordon:
## Why

To test out more improvements on auto-completion and signature help, especially when we are moving the cursor in the query


packages/language-support/src/tests/benchmarks/benchmark.ts
```diff
@@ -16,7 +17,11 @@ import {
   tictactoe,
 } from './benchmark-queries';
 
-Benchmark.options.minSamples = 20;
+const periodicIterate = 'CALL apoc.periodic.iterate(';
+const periodicIterateFirstArg = '"MATCH (p:Person) RETURN id(p) as personId", ';
+
+Benchmark.options.minSamples = 10;
+Benchmark.options.maxTime = 1;
```
ncordon, line 24: This doesn't have much effect in the results of the benchmark, except on the parsing time, which I believe is cause the first iteration is always going to be really slow

OskarDamkjaer: I think the semantic analysis for the really large query takes more than a second per cycle, so won't setting the maxTime effectively make that benchmark useless?

OskarDamkjaer (approved):

packages/language-support/src/tests/benchmarks/benchmark.ts
ncordon: With the 1s limit:

```
simple - validate syntax x 5,016 ops/sec ±4.13% (26 runs sampled)
movies - validate syntax x 0.38 ops/sec ±2.84% (10 runs sampled)
tictactoe - validate syntax x 126 ops/sec ±6.65% (23 runs sampled)
```

Without:

```
simple - validate syntax x 4,893 ops/sec ±1.38% (93 runs sampled)
movies - validate syntax x 0.39 ops/sec ±2.01% (10 runs sampled)
tictactoe - validate syntax x 124 ops/sec ±3.25% (74 runs sampled)
```

I don't think that 1s is a hard limit, so if the benchmarks overrun that it detects it well

PR #171, opened 2024-02-06
Splits parsing of statements

ncordon:
## What
Parses a query with independent statements.

## Why
Because we want to have independent statements for the collected variables, which get offered in auto-completions, but also to avoid destroying and recreating the parsing several times on every keystroke.


packages/language-support/src/autocompletion/autocompletion.ts
```diff
@@ -1,15 +1,15 @@
...
 
 import { DbSchema } from '../dbSchema';
-import { findLatestStatement } from '../helpers';
 import { parserWrapper } from '../parserWrapper';
 import { completionCoreCompletion } from './completionCoreCompletions';
 
 export function autocomplete(
   textUntilPosition: string,
   dbSchema: DbSchema,
+  // TODO Add caret index here
```
ncordon, line 10: For the next pr

packages/language-support/src/helpers.ts
```diff
@@ -139,6 +110,80 @@ export function inRelationshipType(stopNode: ParserRuleContext) {
...
+class MyTokenStream extends CommonTokenStream {
+  tokens: antlrDefaultExport.Token[] = [];
+
+  constructor(lexer: CypherLexer, tokens: Token[]) {
+    super(lexer);
+    this.tokens = [...tokens];
+  }
+}
+
+export function findCaret(
```
ncordon, line 122: Just moves and reshapes this from the signature help, hopefully it can be reused for the auto-completion

packages/react-codemirror-playground/src/tree-util.ts
```diff
@@ -25,5 +26,11 @@ export function getDebugTree(cypher: string): SimpleTree {
     };
   }
 
-  return walk(statements);
+  const children = statements.map((statement) => walk(statement));
+
+  return {
+    // TODO Is this correct?
+    name: 'root',
```
ncordon, line 33: Should we name this differently?

packages/language-support/src/helpers.ts
OskarDamkjaer: I feel like there should be a good way to find where the cursor is in the tree without having to go looking for it, but I've not found anything either, a bit dissapointing

packages/language-support/src/highlighting/syntaxValidation/syntaxValidation.ts
```diff
@@ -163,57 +161,62 @@ export function sortByPosition(a: SyntaxDiagnostic, b: SyntaxDiagnostic) {
...
-        const { notifications, errors } = wrappedSemanticAnalysis(cmd.query);
-
-        return fixSemanticAnalysisPositions({
-          cmd,
-          semanticElements: notifications.concat(errors),
-          parseResult: reparse,
-        });
+/**
+ * Assumes the provided query has no parse errors
+ */
```
OskarDamkjaer, line 195: Since we do the parse here anyway, why assume that there are no parse errors when we can just check `cachedParse.diagnostics`?

packages/language-support/src/parserWrapper.ts
```diff
@@ -68,52 +91,78 @@ export type LabelOrRelType = {
...
-  tokenStream: CommonTokenStream;
-}
-
 export function createParsingScaffolding(query: string): ParsingScaffolding {
   const inputStream = CharStreams.fromString(query);
   const lexer = new CypherLexer(inputStream);
   const tokenStream = new CommonTokenStream(lexer);
-  const parser = new CypherParser(tokenStream);
-  parser.removeErrorListeners();
+  const stsTokenStreams = splitIntoStatements(tokenStream, lexer);
```
OskarDamkjaer, line 98: `sts` short for statements? you don't need change it, it just tripped me up a little since I think we've shortened it `stmts` other places in the code base

packages/language-support/src/parserWrapper.ts
```diff
@@ -68,52 +91,78 @@ export type LabelOrRelType = {
...
+
+  const statementsScaffolding: StatementParsingScaffolding[] =
+    stsTokenStreams.map((t) => {
+      const tokens = [...t.tokens];
```
OskarDamkjaer, line 102: Why doe we need to copy this list? Is there some mutation going on later or is it just to be on the safe side?

packages/language-support/src/signatureHelp.ts
```diff
@@ -133,49 +133,35 @@ class SignatureHelper extends CypherCmdParserListener {
...
-}
-
 export function signatureHelp(
   fullQuery: string,
   dbSchema: DbSchema,
   caretPosition: number,
 ): SignatureHelp {
   let result: SignatureHelp = emptyResult;
 
+  // TODO Is this check needed
```
OskarDamkjaer, line 143: are you leaving this for the next pr as well?

packages/language-support/src/tests/consoleCommands.test.ts
```diff
@@ -99,7 +107,10 @@ describe('sanity checks', () => {
   });
 
   test('accepts trailing ; ', () => {
-    expectParsedCommands(':history;', [{ type: 'history' }]);
+    expectParsedCommands(':history;', [
+      { type: 'history' },
+      { type: 'parse-error' },
```
OskarDamkjaer, line 112: My gut feeling is that this should not give a parse error here. To me it makes sense that empty string on it's own should be a parse error, but empty string as the last statement in a list should not be considered a statement at all?

packages/react-codemirror-playground/src/tree-util.ts
OskarDamkjaer: We've got a similar pattern in the `parserAdapter` (making up a new node that serves only as a root), there we call it `topNode`. I think it makes sense to call both the same, but I'll leave it to you to decide if that's `root`, `virtualRoot`, `topNode` or some other name

packages/react-codemirror/src/e2e_tests/syntax-validation.spec.tsx
```diff
@@ -16,6 +16,7 @@ test('Prop lint set to false disables syntax validation', async ({
   });
 });
 
+// TODO Should this be unskipped
```
OskarDamkjaer, line 19: do you want me to look into this? Or is it relevant to the PR somehow?

packages/react-codemirror/src/lang-cypher/syntax-validation.ts
```diff
@@ -55,7 +55,13 @@ export const semanticAnalysisLinter: (config: CypherConfig) => Extension = (
 
     // we want to avoid the ANTLR4 reparse in the worker thread, this should hit our main thread cache
     const parse = parserWrapper.parse(query);
-    if (parse.diagnostics.length !== 0) {
+    const statements = parse.statementsParsing;
+
+    const anySyntacticError =
+      statements.filter((statement) => statement.diagnostics.length !== 0)
+        .length > 0;
```
OskarDamkjaer, line 62: For convenience it could be interesting to have something aggregating helper functions on the top level parse result?  
```ts
const parse = parserWrapper.parse(query);
if(parse.getAllDiagnostics().length > 0) {
```

OskarDamkjaer (commented):
Nice work! I've mostly left some questions to understand the changes better :+1:

packages/language-support/src/helpers.ts
```diff
@@ -139,6 +110,79 @@ export function inRelationshipType(stopNode: ParserRuleContext) {
...
+    const current = tokens[i].clone();
+    current.tokenIndex -= offset;
+
+    chunk.push(current);
+
+    if (
+      current.type === CypherLexer.SEMICOLON ||
+      current.type === CypherLexer.EOF
+    ) {
+      // This does not relex since we are not calling fill on the token stream
```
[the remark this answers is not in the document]
ncordon: Relex = call the lexer again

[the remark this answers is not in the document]
OskarDamkjaer: ah of course

packages/language-support/src/autocompletion/autocompletion.ts
ncordon: https://github.com/neo4j/cypher-language-support/pull/174

packages/language-support/src/parserWrapper.ts
ncordon: It surprised me but when splitting the token stream into separate ones, the parser can sometimes modify some of them and add an extra EOF, which I would not expect, so it's just a protection against the parser modifying the tokens

packages/react-codemirror/src/lang-cypher/syntax-validation.ts
ncordon: Wouldn't that be creating extra objects (cause we need to concatenate possibly a few arrays from the internal statements)? 

Not sure there's a logical view for arrays in Typescript or a similar concept

packages/language-support/src/tests/consoleCommands.test.ts
ncordon: It should be considered a statement cause I do need the auto-completion to pick up that we are in a new statement to suggest new things.

I've changed this for `{ statement: ''", type: "cypher"}`

packages/react-codemirror-playground/src/tree-util.ts
ncordon: Done

packages/language-support/src/helpers.ts
```diff
@@ -139,6 +110,79 @@ export function inRelationshipType(stopNode: ParserRuleContext) {
...
+      result.push(new CustomTokenStream(lexer, chunk));
```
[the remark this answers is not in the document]
ncordon: You are right, I could just create one of `CommonTokenStream`s and override the tokens

packages/react-codemirror/src/e2e_tests/syntax-validation.spec.tsx
ncordon: I've uncommented the test and it works, not sure if this was an oversight and we forgot to unskip it in the past

OskarDamkjaer (approved):

PR #172, opened 2024-02-06
Corrects a bug with an empty query in VSCode

ncordon:
## What
When writing something with an error, i.e `MATCH (n)` in VSCode, and replacing it with an empty query, we were still showing the old error.

## Why
Because we were not sending empty diagnostics in the case of an empty query.


OskarDamkjaer (approved):
Good catch! :10

PR #177, opened 2024-02-21
Remove `workerpool` usage in codemirror to enable the workspace builld

OskarDamkjaer:
It's not easy getting `workerpool` and `vite` to play nice, see this issue [issue](https://github.com/josdejong/workerpool/issues/341). I thought I had a something working well, but in the end I was unable to get it to work properly in Query/Browser. 

I re-wrote the original implementation I had for a small workerpool, a "worker puddle" if you will. I used`verdaccio` to get it properly in workspace, so I know this config "should" build properly in workspace, finally.

I still think it's nice to keep `workerpool` in the vscode extension because I assume it's a better implementation than what I've done here, easier API to use and to not change something that already works. Because of the node/browser differences it'll likely not be easy to share the code regardless, but if we run into issues with workerpool or if you feel strongly about it @ncordon I'm fine with removing it in another PR


packages/react-codemirror-playground/vite.config.ts
```diff
@@ -3,5 +3,5 @@ import { defineConfig } from 'vite';
 
 export default defineConfig({
   plugins: [react()],
-  base: '/cypher-language-support',
+  base: '',
```
OskarDamkjaer, line 6: Setting this to empty string allowed me to test the built source locally, as well as host it in our demo, see docs: https://vitejs.dev/config/shared-options.html#base

OskarDamkjaer:
After fiddling with the imports/set up workspace and I tried using the workerpool release - and it worked. The main problems were not with workerpool but with a few vite specifics ([links](https://github.com/vitejs/vite/issues/11672) [for](https://github.com/vitejs/vite/issues/10047) the [curious](https://vitejs.dev/config/dep-optimization-options#optimizedeps-exclude)). Combine the above issues with the semantic analysis not working in firefox (something to look into once TeaVM releases their next update), I got thrown off track.

Debugging this has been very fiddly and I'm sure it's not the last we hear of it (I think the code in main could work less well without vite), but let's close this PR and hope we don't need it in the future :)

PR #179, opened 2024-02-27
Updates grammar

ncordon:
Updates to the 5.18 grammar from the database. Also it splits the `EXPLAIN` and `PROFILE` to a separate grammar. I'm happy to either have that pyramid of grammars or write it inside `CypherCmdParser` instead. We will most probably need a more developed preparser sooner or later because all of those `CYPHER` options you can prefix a query with.


OskarDamkjaer:
If we don't see any issues with having the preparser be separate I think it makes sense to have it be separate like you've done here, since that's how it's in the main database (?)

OskarDamkjaer (approved):

PR #184, opened 2024-03-04
implement basic snippet completions for vscode & codemirror

OskarDamkjaer:


packages/vscode-extension/package.json
```diff
@@ -7,7 +7,8 @@
   "version": "2.0.0-next.4",
   "publisher": "Neo4j Inc.",
   "activationEvents": [
-    "onLanguage:cypher"
+    "onLanguage:javascript",
+    "onLanguage:typescript"
```
ncordon, line 11: I think this is not from this pr right? Looks more like you were trying how to provide embedded support for Cypher?

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -282,6 +283,13 @@ function calculateNamespacePrefix(
   return namespacePrefix;
 }
 
+const baseTriggerCharacters = ['.', '(', ')', ','];
+export const snippetTriggerCharacters = ['<', '-'];
```
ncordon, line 287: Did you solve the problem with the `-` and `<`?

packages/vscode-extension/package.json
OskarDamkjaer: yes that's right

packages/language-support/src/autocompletion/completionCoreCompletions.ts
OskarDamkjaer: yep, will open a new PR with only the relevant changes

OskarDamkjaer:
closing in favor of #190

PR #188, opened 2024-03-08
Complete usernames and roles

OskarDamkjaer:
Polls role and user names from the database and offers auto-completions for them.

## Examples

We only allow existing users and string parameters:

```
ALTER USER 
           ^
```

We only allow non-existing users here (i.e. string parameters):

```
RENAME USER user TO 
                    ^
```

We only allow non-existing users here:

```
GRANT TERMINATE TRANSACTION (
                             ^
GRANT TERMINATE TRANSACTION (user, 
                                  ^
```

Analogously for transactions


packages/language-support/src/tests/autocompletion/databasenameCompletion.test.ts
```diff
@@ -223,230 +223,159 @@ describe('can complete database names', () => {
...
-        'RENAME SERVER ',
-        'RENAME SERVER $adb TO ',
-        'DROP SERVER ',
-        'DEALLOCATE DATABASES FROM SERVERS ',
-        'DEALLOCATE DATABASES FROM SERVERS "ab", ',
-      ];
-      const optionsCases = [
-        'ENABLE SERVER "abc" OPTIONS ',
-        'ALTER SERVER "abc" SET OPTIONS ',
-      ];
```
ncordon, line 423: I've extracted these tests to its own file

OskarDamkjaer (commented):
Nice work - thorough tests, well factored/readable implementation that fits well with the rest of the project

OskarDamkjaer:
Would be good to get a changeset as well so we release the changes

ncordon (approved):

PR #194, opened 2024-04-02
Handle accepting completion within snippets

OskarDamkjaer:
If there's a completion open, tab etc should accept the completion instead of moving to the next field.


packages/react-codemirror/src/e2e_tests/autoCompletion.spec.tsx
```diff
@@ -244,13 +244,42 @@ test('can complete pattern snippet', async ({ page, mount }) => {
...
+
+  await textField.press('Tab');
+  await expect(page.locator('.cm-tooltip-autocomplete')).not.toBeVisible();
+
+  // tab out of the snippet
+  await textField.press('Tab');
+
+  await expect(textField).toHaveText('MATCH ()-[]->()-[ ]->(:City)');
+});
+
```
ncordon, line 282: Do we have any test to check the shift + Tab traverses the snippets back or that the Escape clears them out?

ncordon (commented):
We still don't have a snippet for the pattern `()-[]-()`.

Also as some inspiration, maybe we could check the ones that were included by @adam-cowley [here](https://github.com/adam-cowley/neo4j-vscode/blob/main/cypher/cypher.snippets.json)

packages/react-codemirror/src/e2e_tests/autoCompletion.spec.tsx
OskarDamkjaer: No, I but I'll add some :+1:

OskarDamkjaer:
@ncordon I did add the `()-[]-()` snippet as you suggested in the last PR! I didn't know about the snippets from adams extensions, neat! We should show greg

packages/react-codemirror/src/e2e_tests/autoCompletion.spec.tsx
OskarDamkjaer: I broke out a `snippets.spec.tsx` file and added a test for navigating snippets

ncordon (approved):

PR #198, opened 2024-04-23
Signature Help CSS, step in the right direction

OskarDamkjaer:
Here's how it looks:
https://github.com/neo4j/cypher-language-support/assets/10564538/63660615-4d13-4584-9db4-215def412fce

Want to have greg & you try it and give feedback for the next steps.

I also added a ts-config for the turbowatch file, so we don't need to ignore it from the styling linting anymore.

### DIFF
--- a/packages/react-codemirror/src/lang-cypher/signatureHelp.ts
+++ b/packages/react-codemirror/src/lang-cypher/signatureHelp.ts
@@ -57,6 +57,10 @@ function getSignatureHelpTooltip(
               const dom = document.createElement('div');
               dom.className = 'cm-signature-help-panel';
 
+              const contents = document.createElement('div');
+              contents.className = 'cm-signature-help-panel-contents';
+              dom.appendChild(contents);
+
               const signatureLabel = document.createElement('div');
               signatureLabel.className = 'cm-signature-help-panel-name';
               signatureLabel.appendChild(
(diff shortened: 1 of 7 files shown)


packages/react-codemirror/src/lang-cypher/signatureHelp.ts
```diff
@@ -45,16 +45,7 @@ function getSignatureHelpTooltip(
...
-        let doc = signature.documentation.toString();
-
-        if (
-          activeParameter >= 0 &&
-          activeParameter <
-            (signatures[activeSignature].parameters?.length ?? 0)
-        ) {
-          doc =
-            parameters[activeParameter].documentation.toString() + '\n\n' + doc;
-        }
```
ncordon, line 57: This was a safeguard in case we return an active parameter of -1, which if I recall correctly it could happen, so please don't remove

ncordon (approved):

packages/react-codemirror/src/lang-cypher/signatureHelp.ts
OskarDamkjaer: Ah, that was not my intention, I'll add back a safe guard :+1:

OskarDamkjaer: Hmm, wait looking into it now, it's still there just moved

PR #200, opened 2024-04-24
Adds TextMate grammar and Markdown highlighting

ncordon:
Adds a TextMate grammar with support for limited constructs.

Enables Cypher highlighting inside ` ```cypher` Markdown blocks.

Visually the fixtures added would show like this:

(the comment above doesn't show up in the right colour because they are not supported in this pr)


packages/language-support/src/lexerSymbols.ts
```diff
@@ -406,3 +406,6 @@ export const tokenNames: string[] = Array.from({ length }).map(
 );
 
 export const keywordNames = new Set(lexerKeywords.map((i) => tokenNames[i]));
+export const operatorSymbols = new Set(
+  lexerOperators.map((i) => literalNames[i].replaceAll("'", '')),
```
ncordon, line 410: The lexer operators are wrapped up in `'` inside ANTLR, e.g. the vocabulary has `'*'` instead of `*`, so we neeed to clean that manually here

packages/language-support/src/textMateGrammar.ts
```diff
@@ -0,0 +1,103 @@
+import { keywordNames, operatorSymbols } from './lexerSymbols';
+
+const keywordRegex = new Array(...keywordNames.values()).join('|');
+const operatorsRegex = new Array(...operatorSymbols.values())
+  .map((v) => v.replace(/[.*+?^${}()|[\]\\/]/g, '\\$&'))
```
ncordon, line 5: Escapes reserved constructs in regexes, e.g. build the regex `\\*` instead of `*` which has the meaning of repeating the previous character indefinitely.

packages/vscode-extension/syntaxes/cypher.json
```diff
@@ -0,0 +1,89 @@
+{
```
ncordon, line 1: This is generated automatically from the `textMateGrammar.ts` using `genTextMate.js`

packages/language-support/src/lexerSymbols.ts
OskarDamkjaer: Are there any literals containing the `'` sign naturally where this is problematic?

packages/vscode-extension/package.json
```diff
@@ -46,6 +46,23 @@
         "configuration": "./cypher-language-configuration.json"
       }
     ],
+    "grammars": [
```
OskarDamkjaer, line 49: are we interested in having this highlighting by default as well (instead of/in addition to the semantic highlighting)?

OskarDamkjaer (approved):
Regexes are famously write-only so I'm sure there'll be some edge case we've not thought of, but it's worked well in my testing so far (although I miss the highlighting for comments)

packages/language-support/src/lexerSymbols.ts
ncordon: Not for the lexer operators

packages/vscode-extension/package.json
ncordon: This is already on by default. The semantic highlighting just overrides it

PR #201, opened 2024-04-25
Speeds up build and tests

ncordon:
This makes `tsc` not care about the `semanticAnalysis.js` file when emitting type declarations, but it's still bundling it with `esbuild`

Before:

After:

Also the semantic analysis tests run really fast now and the debugger works way way faster when setting up a debug point in the code.


OskarDamkjaer (approved):
Great stuff, nice find!

PR #206, opened 2024-05-08
Adds job to release a canary package to npm manually

ncordon:
Adds a `release-canary.js` script to do all the heavy lifting of releasing canary packages of the `language-support` and `react-codemirror` on every merge to master.

This means on every merge to master we will release a new version `x.y.z-canary-{git commit hash}` of those packages.

This has a few complications:
* We need to update the versions for the corresponding packages
* We have a monorepo, so if we release language-support, we need to update its version in react-codemirror to be able to use this new version.
* There's no current support in changesets for doing this, so we need to craft our own solution.
* Having a `.js` script to do this is not out of the ordinary (a bash one could have been hell to get right IMO), that's what they do in the React package to release canary:
[example](https://github.com/facebook/react/blob/main/scripts/release/build-release-locally-commands/update-version-numbers.js), [other example](https://github.com/facebook/react/blob/ec15267a001086deb4ab5412d3f8b7e13573d6a5/scripts/release/utils.js#L223).

Example of run from this same pr:


release-canary.js
```diff
@@ -0,0 +1,46 @@
+const semver = require('semver');
```
ncordon, line 1: Having a `.js` script to do this is not out of the ordinary, that's what they do in the React package to release canary:

* https://github.com/facebook/react/blob/main/scripts/release/build-release-locally-commands/update-version-numbers.js
* https://github.com/facebook/react/blob/ec15267a001086deb4ab5412d3f8b7e13573d6a5/scripts/release/utils.js#L223

release-canary.js
```diff
@@ -0,0 +1,85 @@
...
+  .toString()
+  .trim();
+
+function exec(command) {
+  try {
+    console.log(`Executing: ${command}`);
+    childProcess.execSync(command);
+  } catch (error) {
+    console.log(`Error while executing ${command}\n` + error);
+    throw new Error();
```
OskarDamkjaer, line 16: would be nice with some message here, or rethrowing the old error (i.e. `throw error`

release-canary.js
```diff
@@ -0,0 +1,85 @@
...
+  const jsonFile = `${wd}/${package}/package.json`;
+  const packageJson = require(jsonFile);
+  packageJson.dependencies[dependency] = newVersion;
+  const data = JSON.stringify(packageJson, null, '  ');
+  try {
+    fs.writeFileSync(jsonFile, data);
+  } catch (error) {
+    // logging the error
+    console.error(error);
+    throw new Error();
```
OskarDamkjaer, line 44: same here

release-canary.js
```diff
@@ -0,0 +1,85 @@
...
+    `Updating dependency ${dependency} in ${package} to ${newVersion}\n`,
+  );
+  const jsonFile = `${wd}/${package}/package.json`;
+  const packageJson = require(jsonFile);
+  packageJson.dependencies[dependency] = newVersion;
+  const data = JSON.stringify(packageJson, null, '  ');
+  try {
+    fs.writeFileSync(jsonFile, data);
+  } catch (error) {
+    // logging the error
```
OskarDamkjaer, line 42:
```suggestion
```

release-canary.js
```diff
@@ -0,0 +1,85 @@
...
+updateVersion(langSupport, langSupportVersion);
+replaceDependency(
+  reactCodemirror,
+  '@neo4j-cypher/language-support',
+  langSupportVersion,
+);
+updateVersion(reactCodemirror, reactCodemirrorVersion);
+buildProject();
+publishCanaryVersion(langSupport);
+publishCanaryVersion(reactCodemirror);
```
OskarDamkjaer, line 85: makes sense to limit the canaries to reactcodemirror & lang support :+1: 

We'll have to remember to update this script for when we break out the `codemirror` part of the `react-codemirror` to it's own package.

OskarDamkjaer (approved):

PR #209, opened 2024-05-09
Adds GH job to publish the VSCode ext to Marketplace and OpenVSX

ncordon:


.github/workflows/publish-vscode-extension.yaml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '18'
+
+      - name: Setup antlr4
+        uses: ./.github/actions/setup-antlr4
```
OskarDamkjaer, line 24: feels like it's soon time for us to consider caching builds in CI, so we only do one build per merge and just download the artifacts

OskarDamkjaer (approved):
🔥

PR #214, opened 2024-05-14
Adds signature help information on auto-completion

ncordon:
Adds signature information on auto-completion.

In codemirror we can see it this way:

In VSCode we can see it like this:

and we can toggle a wider box with the full description clicking on the `>`:


packages/language-support/src/types.ts
```diff
@@ -34,4 +36,9 @@ export type Neo4jFunction = {
   returnDescription: string;
   signature: string;
   aggregating: boolean;
+  isDeprecated: boolean;
+};
+
+export type CompletionItem = VSCodeCompletionItem & {
+  signature?: string;
 };
```
ncordon, line 44: The reason for overriding this type is that I need to copy the signature into the `details` of `CompletionItem` after the type  (e.g. `(function)`) for VSCode to show it in the right place. But for codemirror we need the type in a different place than the method information.

packages/react-codemirror/src/e2e_tests/configuration.spec.tsx
```diff
@@ -1,8 +1,6 @@
 import { expect, test } from '@playwright/experimental-ct-react';
 import { CypherEditor } from '../CypherEditor';
 
-test.use({ viewport: { width: 500, height: 500 } });
```
ncordon, line 4: This cleanup is so that things shape up in the same way that they do in the codemirror demo

packages/react-codemirror/src/lang-cypher/langCypher.ts
```diff
@@ -31,8 +32,9 @@ export function cypher(config: CypherConfig) {
   const cypherLanguage = new Language(facet, parserAdapter, [], 'cypher');
 
   return new LanguageSupport(cypherLanguage, [
-    cypherLanguage.data.of({
-      autocomplete: cypherAutocomplete(config),
+    autocompletion({
+      override: [cypherAutocomplete(config)],
+      optionClass: completionStyles,
```
ncordon, line 37: This is so we can style the deprecations as strikethrough

packages/language-support/src/tests/autocompletion/functionsCompletion.test.ts
```diff
@@ -134,16 +178,22 @@ describe('function invocations', () => {
...
-          label: 'first',
+          label: 'last',
           kind: CompletionItemKind.Function,
           detail: '(function)',
+          signature: functions['apoc.agg.last'].signature,
+          documentation: functions['apoc.agg.last'].description,
         },
         {
-          label: 'first',
+          label: 'slice',
```
ncordon, line 192: This test was not testing what it said

packages/react-codemirror/src/lang-cypher/signatureHelp.ts
```diff
@@ -20,14 +18,6 @@ function getTriggerCharacter(query: string, caretPosition: number) {
...
 }
 
-function getDocString(result: string | MarkupContent): string {
-  if (MarkupContent.is(result)) {
-    result.value;
-  } else {
-    return result;
-  }
-}
-
```
ncordon, line 30: I've moved this to a common utils file since it's needed for the auto-completions as well

packages/schema-poller/src/data-transforms/clean-type.ts
```diff
@@ -0,0 +1,13 @@
+export function cleanType(type: string): string {
```
ncordon, line 1: Copied from upx

packages/schema-poller/src/queries/functions.ts
```diff
@@ -1,8 +1,61 @@
...
+
+function cleanTypes(result: Neo4jFunction): Neo4jFunction {
+  return {
+    ...result,
+    argumentDescription: result.argumentDescription.map(cleanTypeDescription),
+    returnDescription: cleanType(result.returnDescription),
+  };
+}
+
+export const functionSchema: JSONSchemaType<Neo4jFunction> = {
```
ncordon, line 17: Copied from upx

packages/schema-poller/src/queries/procedures.ts
```diff
@@ -1,14 +1,81 @@
 import { Neo4jProcedure } from '@neo4j-cypher/language-support';
+import Ajv, { JSONSchemaType } from 'ajv';
 import { resultTransformers } from 'neo4j-driver';
+import { cleanTypeDescription } from '../data-transforms/clean-type';
 import type { ExecuteQueryArgs } from '../types/sdkTypes';
 
-type ListFunctionArgs = { executableByMe: boolean };
+type ListProcedureArgs = { executableByMe: boolean };
+
+export const procedureSchema: JSONSchemaType<Neo4jProcedure> = {
```
ncordon, line 9: Copied from upx

packages/schema-poller/src/queries/procedures.ts
```diff
@@ -1,14 +1,81 @@
 import { Neo4jProcedure } from '@neo4j-cypher/language-support';
+import Ajv, { JSONSchemaType } from 'ajv';
 import { resultTransformers } from 'neo4j-driver';
+import { cleanTypeDescription } from '../data-transforms/clean-type';
 import type { ExecuteQueryArgs } from '../types/sdkTypes';
 
-type ListFunctionArgs = { executableByMe: boolean };
+type ListProcedureArgs = { executableByMe: boolean };
```
ncordon, line 7: Just renamed because it's not functions, it's procedures

packages/react-codemirror/src/lang-cypher/langCypher.ts
```diff
@@ -19,20 +20,30 @@ const facet = defineLanguageFacet({
...
   schema?: DbSchema;
   useLightVersion: boolean;
   setUseLightVersion?: (useLightVersion: boolean) => void;
 };
 
 export function cypher(config: CypherConfig) {
-  _internalFeatureFlags.consoleCommands = true;
+  const featureFlags = config.featureFlags;
+  // We allow to override the consoleCommands feature flag
+  if (featureFlags.consoleCommands !== undefined) {
```
OskarDamkjaer, line 35: should all feature flags be handled here? `signatureInfoOnAutoCompletions` and any future ones we add

packages/react-codemirror/src/lang-cypher/autocomplete.ts
```diff
@@ -65,17 +83,70 @@ export const cypherAutocomplete: (config: CypherConfig) => CompletionSource =
...
+            const paragraph = document.createElement('p');
+            paragraph.textContent = getDocString(o.documentation);
+            newDiv.appendChild(paragraph);
+          }
+
+          maybeInfo = {
+            info: () =>
+              new Promise((resolve) => {
+                resolve(newDiv);
+              }),
```
OskarDamkjaer, line 110: Assuming this needs to a promise, the easiest way/clearest way to wrap it is to just do 
```suggestion
              Promise.resolve(newDive)
```

packages/react-codemirror/src/e2e_tests/autoCompletion.spec.tsx
```diff
@@ -234,3 +232,126 @@ test('completes allShortestPaths correctly', async ({ page, mount }) => {
...
+  const textField = page.getByRole('textbox');
+  await textField.fill('RETURN apoc.coll.');
+
+  await expect(page.locator('.cm-tooltip-autocomplete')).toBeVisible();
+
+  const infoTooltip = page.locator('.cm-completionInfo');
+
+  while (!(await infoTooltip.getByText(fnName).isVisible())) {
+    await page.keyboard.press('ArrowDown');
+    await page.waitForTimeout(100);
```
OskarDamkjaer, line 291: This pattern of a while await with a static timeout strikes me as something that can easily flake. If I understand it correctly what it does is move down the dropdown until the right signature shows? An alternative approach to avoid the waiting for a timeout could be (in pseudo code):

let selectedOption = infoTooltip.getBy(".cm-selected")
while (!selectedOption.textContent.has(fnName)) {
 arrowDown()
 // check that the selection moved
 expect (infoTooltip.getBy(".cm-selected").toNotEqual(selectedOption)
 selected = (infoTooltip.getBy(".cm-selected"))
}

This ensure that's it's not just visible but also selected. It might be fiddly to implement, so if you think it'll work well as is now you can keep it if you prefer

packages/react-codemirror/src/lang-cypher/autocomplete.ts
```diff
@@ -65,17 +83,70 @@ export const cypherAutocomplete: (config: CypherConfig) => CompletionSource =
...
+          if (o.documentation) {
+            const paragraph = document.createElement('p');
+            paragraph.textContent = getDocString(o.documentation);
+            newDiv.appendChild(paragraph);
+          }
+
+          maybeInfo = {
+            info: () =>
+              new Promise((resolve) => {
+                resolve(newDiv);
```
OskarDamkjaer, line 109: I think this always attaches the div, even if there's no info. It seems the panel will pop up on all completions. Perhaps it should resolve to `undefined` instead of the div if there's no signature/documentation? 
CleanShot 2024-05-22 at 13 12 45

packages/react-codemirror/src/lang-cypher/langCypher.ts
ncordon: It depends on whether we want to pass them down the language support or not. Meaning there are things we might want to block at codemirror level, but not necessarily at language support one. For example the signature info it's not needed to gate it at language support level for example.

packages/react-codemirror/src/lang-cypher/autocomplete.ts
ncordon: Good catch! I've added a few tests to prevent empty boxes like that

OskarDamkjaer (approved):

PR #216, opened 2024-05-22
Add an ariaLabel prop to CypherEditor

daveajrussell:
Adding an 'ariaLabel' prop to the CyperEditor props so that we can set an aria-label attribute on the codemirror editor.

This uses the [contentAttributes](https://codemirror.net/docs/ref/#view.EditorView^contentAttributes) facet, which may be extended later to add more DOM attributes, or changed to accept a kvp/Record so that DOM attributes may be fully flexible.

with the `ariaLabel` property set, we can see the attribute set in the playground:

Screenshot 2024-05-22 at 12 23 52

The property is optional, so not setting it should result in the aria-label attribute not being added to the editor:
Screenshot 2024-05-22 at 12 24 19

### DIFF
--- a/packages/react-codemirror/src/e2e_tests/configuration.spec.tsx
+++ b/packages/react-codemirror/src/e2e_tests/configuration.spec.tsx
@@ -100,10 +100,7 @@ test('aria-label is not set by default', async ({ mount, page }) => {
   await mount(<CypherEditor />);
 
   const textField = page.getByRole('textbox');
-
-  await expect(async () => {
-    expect(await textField.getAttribute('aria-label')).toBeNull();
-  }).toPass();
+  expect(await textField.getAttribute('aria-label')).toBeNull();
 });
 
 test('can set aria-label', async ({ mount, page }) => {
@@ -112,8 +109,5 @@ test('can set aria-label', async ({ mount, page }) => {
   await mount(<CypherEditor ariaLabel={ariaLabel} />);
 
   const textField = page.getByRole('textbox');
-
-  await expect(async () => {
-    expect(await textField.getAttribute('aria-label')).toEqual(ariaLabel);
-  }).toPass();
+  expect(await textField.getAttribute('aria-label')).toEqual(ariaLabel);
 });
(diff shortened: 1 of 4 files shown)


ncordon (approved):

packages/react-codemirror/src/e2e_tests/configuration.spec.tsx
```diff
@@ -95,3 +95,25 @@ test('can set/unset onFocus/onBlur', async ({ mount, page }) => {
...
 });
+
+test('aria-label is not set by default', async ({ mount, page }) => {
+  await mount(<CypherEditor />);
+
+  const textField = page.getByRole('textbox');
+
+  await expect(async () => {
+    expect(await textField.getAttribute('aria-label')).toBeNull();
+  }).toPass();
```
ncordon, line 106: I think you don't need the double wrap in an expect and this would do:

```
  expect(await textField.getAttribute('aria-label')).toBeNull();
```

packages/react-codemirror/src/e2e_tests/configuration.spec.tsx
```diff
@@ -95,3 +95,25 @@ test('can set/unset onFocus/onBlur', async ({ mount, page }) => {
...
+test('can set aria-label', async ({ mount, page }) => {
+  const ariaLabel = 'Cypher Editor';
+
+  await mount(<CypherEditor ariaLabel={ariaLabel} />);
+
+  const textField = page.getByRole('textbox');
+
+  await expect(async () => {
+    expect(await textField.getAttribute('aria-label')).toEqual(ariaLabel);
+  }).toPass();
```
ncordon, line 118: Same:

```
expect(await textField.getAttribute('aria-label')).toEqual(ariaLabel);
```

packages/react-codemirror/src/e2e_tests/configuration.spec.tsx
daveajrussell: yeh fair comment - I've removed the double wrapped assertion

packages/react-codemirror/src/e2e_tests/configuration.spec.tsx
daveajrussell: removed :)

daveajrussell:
[Add property for an aria-label in the CypherEditor](https://trello.com/c/IvKLLXtl/157-add-property-for-an-aria-label-in-the-cyphereditor)

PR #217, opened 2024-05-23
Fixes bug with canary release script

ncordon:
Builds only `language-support` and `react-codemirror` when releasing canary packages.

## Why

Because when we merged #216 the process to release to canary started to fail. That is because a new prop was added to `react-codemirror`, and the script does not update the package.json of `react-codemirror-playground` package that depends on it, so it was failing on build.

Building only the packages we intend to publish to npm as canary releases solves the issue.

### DIFF
--- a/release-canary.js
+++ b/release-canary.js
@@ -71,7 +71,7 @@ function buildProject() {
   exec(`
     cd ${wd} &&
     npm i &&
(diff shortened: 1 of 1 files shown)


daveajrussell (approved):
Makes sense 👍

PR #218, opened 2024-05-23
fix issue where database name was not completed when creating alias

OskarDamkjaer:
The database name wasn't completed in the `CREATE ALIAS foo FOR DATABASE` statement as it was caught in the logic for not suggesting old names when creating new ones. I'm uncertain if there's a more sustainable way to catch this without updating the grammar


ncordon (changes requested):
I'm afraid this would not work in case we have more extra spaces because of that tokenIndex:

```
CREATE ALIAS fo      for DATABASE 
```

so it's better to push these changes down the grammar.

packages/language-support/src/antlr-grammar/CypherParser.g4
```diff
@@ -1553,16 +1602,22 @@ symbolicNameOrStringParameter
    | parameter["STRING"]
    ;
 
+aliasName:
+    symbolicAliasNameOrParameter;
+
+databaseName:
+    symbolicAliasNameOrParameter;
+
```
ncordon, line 1610: These are the real changes to the grammar, the rest are just bringing it up to date

packages/language-support/src/antlr-grammar/CypherParser.g4
```diff
@@ -1553,16 +1602,22 @@ symbolicNameOrStringParameter
    | parameter["STRING"]
    ;
 
+aliasName:
+    symbolicAliasNameOrParameter;
+
+databaseName:
+    symbolicAliasNameOrParameter;
```
OskarDamkjaer, line 1609: The change I was expecting was to separate into:

```
createAliasOrDatabaseName: 
    symbolicAliasNameOrParameter;

ReferenceAliasorDatabaseName:
    symbolicAliasNameOrParameter;
```

Because then we could use `ReferenceAliasorDatabaseName` instead of `RULE_symbolicAliasName` in the `preferredRules` and wouldn't have to check the tree if it's a rule that creates or references an alias/database. It could still be worth splitting database & aliases as well and this change also solves the problem, so I'm fine doing it this way if you prefer

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -621,16 +621,24 @@ function completeAliasName({
...
-    )
+    rulesCreatingNewDb.some((rule) => candidateRule.ruleList.includes(rule))
+  ) {
+    return baseSuggestions;
+  }
+
+  // For `CREATE ALIAS $1 FOR DATABASE $2`
+  // Should not suggest $1 but should suggest for $2
+  // Ideally the grammar would separate declaring and referencing database names
+  // relying on the startTokenIndex is brittle as it will change as the grammar changes
```
OskarDamkjaer, line 638: This comment is a little out of date, how about something like this instead
```suggestion
  // For `CREATE ALIAS $1 FOR DATABASE $2`
  // Should not suggest $1 but should suggest for $2
  // so we return base suggestions if we're at the `aliasName` rule 
```

packages/language-support/src/antlr-grammar/CypherParser.g4
ncordon: Yeah I agree that would be nicer for us but it could make it more difficult to enforce for engineers working in the database because it obfuscates the grammar a step further.

I'm happy to keep evolving the grammar, but we should make it gradually because pushing any change is a pain in the monorepo. Out of curiosity I completely tried to split database completions from alias completions here: https://github.com/neo4j/cypher-language-support/commit/245b807a2848c6cee1da1dab2468cef3df223c3d and it was super painful, so I decided to revert it 😬

packages/language-support/src/tests/autocompletion/databasenameCompletion.test.ts
```diff
@@ -111,6 +111,52 @@ describe('Can complete database names', () => {
...
+        // do not suggest non-string parameters
+        { label: '$param2', kind: CompletionItemKind.Variable },
+        { label: '$param3', kind: CompletionItemKind.Variable },
+      ],
+    });
+  });
+
+  test('Correctly completes database name even in a create alias statement including extra of spaces', () => {
+    testCompletions({
+      query: 'CREATE ALIAS foo          FOR     DATABASE ',
```
ncordon, line 139: I've added this extra test, which was the problematic bit in the previous solution

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -617,22 +617,29 @@ function completeAliasName({
   }
 
   // parameters are valid values in all cases of symbolicAliasName
-  const baseSuggestions = parameterCompletions(
+  const parameterSuggestions = parameterCompletions(
```
ncordon, line 620: I think renaming this makes the code read clearer because you know at any point what `baseSuggestions` contain

OskarDamkjaer: makes sense :+1:

packages/language-support/src/antlr-grammar/CypherParser.g4
OskarDamkjaer: alright, good to hear it was considered!

OskarDamkjaer:
Looking nice! Do we want a changeset for this?

ncordon (approved):

PR #219, opened 2024-05-23
Places the signature help tooltip below instead of above

ncordon:
Result:

## Why

Because the header was blocking it for some methods in upx:


OskarDamkjaer:
WDYT about surfacing this in the react props and doing below for the main editor and above for the frame editor where there's more space?

OskarDamkjaer (approved):

PR #221, opened 2024-06-07
Adding a database connection pane and connection editor to the vscode extension

daveajrussell:
This PR introduces a connection pane and connection editor to the vscode extension.

It essentially deprecates the extension settings (barring `server.trace`) in favour of a connection editor that stores connections and credentials in state. 

Connections can be managed from a new connection pane. Here, connections can be modified, deleted and connected to/disconnected from. There is also a tree view for labels and relationship types for each connection.

At the moment, we are artificially limiting the number of connections you may have to just 1. However, the storage layer has been built in such a way that supporting multiple connections will be trivial.

I have tried to establish a pattern such that there is;
- a single presentation layer which interprets commands and displays UI updates to the user.. notifications.. modals etc.
- a single data access layer which updates state and secrets
- a single service layer which handles complex command actions

Since this PR is quite large, I will annotate key changes.

TODO - E2E tests!


packages/language-server/src/server.ts
```diff
@@ -98,32 +98,51 @@ connection.onCompletion(doAutoCompletion(documents, neo4jSchemaPoller));
 
 connection.onDidChangeConfiguration(
```
daveajrussell, line 99: I need to remove this - at the moment, it remains so that the existing e2e tests work 

I need to update the e2e tests so that they create a `Connection` and connect to that, rather than relying on the settings

packages/language-server/src/server.ts
```diff
@@ -98,32 +98,51 @@ connection.onCompletion(doAutoCompletion(documents, neo4jSchemaPoller));
...
-          password: neo4jConfig.password,
-        },
-        { appName: 'cypher-language-server' },
-      );
-    }
+    changeConnection(params.settings.neo4j);
+  },
+);
+
+connection.onNotification(
```
daveajrussell, line 105: this receives notifications from the extension, presently we have a single method that it's listening for which signals that the connection has changed and we need to reconnect

packages/language-server/src/server.ts
```diff
@@ -98,32 +98,51 @@ connection.onCompletion(doAutoCompletion(documents, neo4jSchemaPoller));
...
+  'connectionUpdated',
+  (connectionSettings: Neo4jSettings) => {
+    changeConnection(connectionSettings);
   },
 );
 
+connection.onNotification('connectionDeleted', () => {
+  disconnect();
```
daveajrussell, line 113: disconnect when we delete a connection

packages/schema-poller/src/queries/index.ts
```diff
@@ -0,0 +1 @@
+export { getDataSummary } from './dataSummary';
```
daveajrussell, line 1: I wanted to expose this function so that we can use it in the vscode extension package to list the labels and relationship types in the tree view

I don't know that it really makes sense for the vscode extension to consume a package called 'schema-poller' when all it's doing is running a query.. perhaps there is a better place for these queries so that we can share them between packages.. but I wonder if this is a premature optimisation at this point

packages/vscode-extension/package.json
```diff
@@ -102,6 +200,8 @@
   },
   "dependencies": {
     "@neo4j-cypher/language-server": "2.0.0-next.7",
+    "@neo4j-cypher/schema-poller": "2.0.0-next.6",
+    "neo4j-driver": "^5.12.0",
```
daveajrussell, line 204: I wonder if it might be better to create a "persistent connection" object that wraps the driver so we don't need to expose the vscode extension to this level of abstraction

packages/vscode-extension/resources/scripts/connectionPanel.js
```diff
@@ -0,0 +1,38 @@
...
+      command: 'onAddConnection',
+      connection: this.getConnection(),
+      password: this.getPassword(),
+    });
+  });
+
+  getConnection = () => {
+    const name = document.getElementById('name').value;
+    let key = document.getElementById('key').value;
+    key = !key ? name.toLowerCase().replace(/\s/g, '-') : key;
```
daveajrussell, line 21: keys must remain unique through name changes.. this is flawed.. it needs some unique slug to ensure that a connection with the same name doesn't get the same key

packages/vscode-extension/src/extension.ts
```diff
@@ -46,13 +74,93 @@ export function activate(context: ExtensionContext) {
...
+    context.secrets,
+  );
+
+  await ConnectionRepository.instance.resetConnections();
+
+  // Register our tree view provider to render connections in the sidebar
+  const connectionTreeDataProvider = new ConnectionTreeDataProvider();
+
+  // Register commands
+  context.subscriptions.push(
```
daveajrussell, line 91: I think I may create a commands object that handles these and tidies up this activation function

packages/vscode-extension/src/managers/databaseDriverManager.ts
```diff
@@ -0,0 +1,82 @@
+import { initializeDriver } from '@neo4j-cypher/schema-poller';
```
daveajrussell, line 1: the driver manager wraps a driver instance and ensures that we maintain a single driver instance for a given connection, since initializing drivers is quite expensive

the driver will be reinitialized if either the url or username changes.. my approach to this feels a little clunky.. what if a password changes? perhaps I can store a secure hash containing the url, username _and_ password using the crypto package

packages/schema-poller/src/persistentConnection.ts
daveajrussell, line 1: this class wraps a database driver as a persistent connection to ensure that we do not excessively create and recreate drivers in the vscode extension

it uses a "key" which is sha256 hash of the url:username:password to ensure that the driver is only reinitialized if the user changes or modifies certain parts of a connection

packages/vscode-extension/src/managers/persistentConnectionManager.ts
daveajrussell, line 1: this class is a singleton wrapping the persistent connection object from the schema-poller package

it exposes functionality to check that a connection is successful, updating and closing the connection

it also exposes a HOF to execute a query using the current connection - this is used to retrieve the data summary of a database connection

packages/vscode-extension/src/extension.ts
```diff
@@ -46,13 +50,31 @@ export function activate(context: ExtensionContext) {
...
 
+  // Initialize singletons
+  LangugageClientManager.instance = new LangugageClientManager(client);
+  PersistentConnectionManager.instance = new PersistentConnectionManager();
+  ConnectionRepository.instance = new ConnectionRepository(
+    context.globalState,
+    context.secrets,
+  );
+
+  // Ensure any connections are reset - set the connected flag to false
```
daveajrussell, line 61: actually, this makes no sense - we should honour a connection that's left as connected so our extension reconnects when it activates.. this mirrors existing functionality

packages/schema-poller/src/persistentConnection.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private driverRequiresReinitialization(
+    url: string,
+    credentials: { username: string; password: string },
+  ): boolean {
+    return this._driverKey !== this.getDriverKey(url, credentials);
+  }
+
+  private getDriverKey(
```
daveajrussell, line 70: is this crazy? I want to take a secure hash of the url, username and password and compare hashes to determine if we need to re-initialize the driver for the cases:

- the url, username AND password has changed (the user has switched to a different connection)
- the username only has changed (the user has changed to a different user with the same password)
- the password has changed (the user may have changed their password on the server and is updating the connection)

packages/vscode-extension/e2e_tests/main.ts
```diff
@@ -9,16 +10,14 @@ function setSetting(file: string, variable: RegExp, value: string) {
...
-  const settingsPath = path.join(
-    __dirname,
-    '../../e2e_tests/fixtures/.vscode/',
-  );
-  const settingsTemplate = path.join(settingsPath, 'settings-template.json');
-  const settingsFile = path.join(settingsPath, 'settings.json');
-  fs.copyFileSync(settingsTemplate, settingsFile);
-  setSetting(settingsFile, /\{PORT\}/g, port.toString());
-  setSetting(settingsFile, /\{PASSWORD\}/g, password);
+function updateDotenvFile(port: number, password: string) {
```
daveajrussell, line 13: changing from settings.json to dotenv so that we can load connection details into process.env and consume them when setting up the test suites

packages/schema-poller/src/persistentConnection.ts
OskarDamkjaer: I don't understand the value of this wrapper? Surely we know when the url / password / username has changed and only re-initialize then? If not, it's trivial to pull out the data out of the driver instance and check if it's changed before re-initializing

packages/schema-poller/src/persistentConnection.ts
OskarDamkjaer: what's the value add of the hashing as opposed to comparing the string before hashing (or even the three fields individually)?

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -0,0 +1,245 @@
...
+    );
+
+    const resetCssUri = webview.asWebviewUri(resetCssPath);
+    const vscodeCssUri = webview.asWebviewUri(vscodeCssPath);
+    const connectionPanelCssUri = webview.asWebviewUri(connectionPanelCssPath);
+    const connectionPanelJsUri = webview.asWebviewUri(connectionPanelJsPath);
+
+    const nonce = getNonce();
+    return `
+      <!DOCTYPE html>
```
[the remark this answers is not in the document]
OskarDamkjaer: what do you think @ncordon ?

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -0,0 +1,245 @@
...
+        </head>
+        <body>
+          <div class="container">
+            <div>
+              <h1>${
+                this._connection ? 'Edit Connection' : 'Add New Connection'
+              }</h1>
+              <form class="form">
+                <input type="hidden" id="key" value="${
+                  this._connection?.key ?? getNonce(16)
```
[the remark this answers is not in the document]
OskarDamkjaer: is there some docs on this perhaps?

packages/schema-poller/src/queries/index.ts
OskarDamkjaer: Otherwise using vscode APIs I think would make sense?

packages/vscode-extension/src/extension.ts
```diff
@@ -46,13 +51,48 @@ export function activate(context: ExtensionContext) {
     clientOptions,
   );
 
+  // Initialize singletons
+  LangugageClientManager.instance = new LangugageClientManager(client);
+  PersistentConnectionManager.instance = new PersistentConnectionManager();
+  ConnectionRepository.instance = new ConnectionRepository(
+    context.globalState,
+    context.secrets,
+  );
```
OskarDamkjaer, line 60: is keeping a bunch of singletons around a common pattern you've seen in other vscode extensions? I worry that singletons generally make code harder to test since they add global state, have we considered other ways to achieve the same functionality?

packages/schema-poller/src/persistentConnection.ts
daveajrussell: I wanted to avoid leaving a password in plaintext in application memory

packages/schema-poller/src/queries/index.ts
daveajrussell: yes, I think this is a good change.. the way I've approached this is that the connections side panel and the language server may operate somewhat independently of one another - the connection panel allowing you to browse the labels and relationships of different connections.. but I think isolating it so that you may only browse the properties of the connection you're connected to makes sense

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -0,0 +1,245 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
```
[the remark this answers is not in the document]
daveajrussell: yeah, I'll look into why eslint has this rule in the first place and do something a bit better here

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -0,0 +1,245 @@
...
+} from 'vscode';
+import { ConnectionRepository } from '../repositories/connectionRepository';
+import { Connection } from '../types/connection';
+import {
+  CREATE_CONNECTION_COMMAND,
+  TEST_CONNECTION_COMMAND,
+} from '../util/constants';
+import { getNonce } from '../util/getNonce';
+
+type Data = {
```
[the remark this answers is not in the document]
daveajrussell: sure thing

packages/vscode-extension/src/webviews/connectionPanel.ts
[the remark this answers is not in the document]
daveajrussell: the nonces are here to ensure that only scripts with a specified nonce can run - we're setting up the CSP above to achieve this - it was something that I saw used in various webview examples.. e.g. https://github.com/microsoft/vscode-extension-samples/blob/main/webview-sample/src/extension.ts

packages/schema-poller/src/initializers/driverInitializer.ts
```diff
@@ -0,0 +1,23 @@
+import neo4j, { Config, Driver } from 'neo4j-driver';
+
```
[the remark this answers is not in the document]
daveajrussell: sounds good :)

packages/language-server/src/server.ts
```diff
@@ -96,34 +96,47 @@ connection.onSignatureHelp(doSignatureHelp(documents, neo4jSchemaPoller));
...
-          username: neo4jConfig.user,
-          password: neo4jConfig.password,
-        },
-        { appName: 'cypher-language-server' },
-      );
-    }
+connection.onNotification(
+  'connectionUpdated',
+  (connectionSettings: Neo4jSettings) => {
+    changeConnection(connectionSettings);
```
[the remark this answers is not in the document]
daveajrussell: I have updated this so that when you add a new connection, you will connect to it

packages/schema-poller/src/schemaPoller.ts
```diff
@@ -1,32 +1,23 @@
...
+import { Config } from 'neo4j-driver';
+import { initializeDriver } from './driverInitializer';
 import { MetadataPoller } from './metadataPoller';
 import { Neo4jConnection } from './neo4jConnection';
 import { listDatabases } from './queries/databases.js';
 
 export class Neo4jSchemaPoller {
   public connection?: Neo4jConnection;
   public metadata?: MetadataPoller;
+  public connectionEvents = new EventEmitter();
```
daveajrussell, line 11: create an EventEmitter that we will use for communication between the schema poller and the language server

packages/vscode-extension/src/connection.ts
```diff
@@ -0,0 +1,161 @@
...
+  [key: string]: Connection | null;
+};
+
+const CONNECTIONS_KEY: string = 'connections';
+
+export async function testConnection(
+  connection: Connection,
+  password: string,
+): Promise<boolean> {
+  return await testConnectionTransiently(
```
daveajrussell, line 34: I wonder if there's a better way to do this.. this function creates a driver, checks that the credentials are valid and there's connectivity, then disposes the driver.. this seems wasteful, but I can't think of a better way..

packages/vscode-extension/src/registerDisposables.ts
```diff
@@ -0,0 +1,146 @@
...
+            MethodName.ConnectionDeleted,
+            currentConnection,
+          );
+        }
+      },
+    ),
+    client.onNotification(
+      PollingEvent.POLLING_REFETCHED_DONE,
+      (data: Neo4jDataSummary) => {
+        // onNotification does not expect a promise to be returned
```
daveajrussell, line 136: this is horrible..

packages/schema-poller/src/queries/index.ts
daveajrussell: okay - I have made this change

I have established an EventEmitter (on the poller.. I did try to establish it in the connection, but communication wasn't picking up properly.. I'll try again in the morning..) that I am using to communicate events from the poller to the language server, the language server then uses a communication mechanism to communicate with the extension i.e. poller -> language server -> vscode extension

I opted for the additional hop so we didn't need to expose the vscode extension to the poller class

--

the reason I did this in the first place, and it's not right in hindsight, is that I was thinking we could have the connection panel independent of the language server connection.. i.e. if you had multiple connections you could browse their labels and nodes regardless of being connected to it

anyway, this is better.. and it makes more sense

packages/schema-poller/src/persistentConnection.ts
daveajrussell: I've removed this now - I've gone a little bit into why I had this in my other comment where you've asked about the driver connections

packages/vscode-extension/src/managers/persistentConnectionManager.ts
daveajrussell: this has gone

packages/schema-poller/src/persistentConnection.ts
daveajrussell: this has now been removed

packages/vscode-extension/src/webviews/connectionPanel.ts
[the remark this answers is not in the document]
daveajrussell: okay - I am now toStringing the URIs in the template expressions

packages/schema-poller/src/initializers/driverInitializer.ts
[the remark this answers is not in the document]
daveajrussell: I have removed the folder

packages/vscode-extension/src/webviews/connectionPanel.ts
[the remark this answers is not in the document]
daveajrussell: this has now been more aptly named

packages/vscode-extension/src/extension.ts
daveajrussell: these have now largely been removed

I have a small class that returns an instance of globalState and secretManager.. these 2 objects are only available on the ExtensionContext object, which (as far as I can see) you can't access elsewhere during runtime outside of the activate function.. this getAppContext function I wrote is looking at me a bit funny.. but this does seem to be the proscribed method for data storage https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage

daveajrussell:
> As you noted there's quite a lot going on here, could we find some way to break it out into smaller chunks as it usually ends up taking a long time to get a PR like this in in one go?

yes.. I've given it some thought

we could simplify it down to

- iteration 1: release the updated connection editor
  - you can simply add/edit/delete/connect to/disconnect from a single connection.. no option to list labels/rels
- iteration 2: implement tree view
  - make listing labels/rels for a single connection possible
- iteration 3: multiple connections

iteration 1 will still be a little bit chunky.. but the complexity of iteration 2 will be clearer for sure

packages/vscode-extension/src/registerDisposables.ts
```diff
@@ -0,0 +1,146 @@
...
+      },
+    ),
+    commands.registerCommand(
+      constants.SAVE_CONNECTION_COMMAND,
+      async (
+        connection: Connection,
+        password: string,
+        isNew: boolean,
+      ): Promise<boolean> => {
+        if (await testConnection(connection, password)) {
```
daveajrussell, line 61: something to consider.. do we want to avoid testing the connection each time we add it 

maybe we want to keep track of whether a connection has been tested before saving though this might make the form more complex.. it will reduce overhead on the driver

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -0,0 +1,245 @@
...
+  private readonly _extensionUri: Uri;
+  private readonly _connection: Connection | undefined;
+
+  private _disposables: Disposable[] = [];
+
+  private constructor(panel: WebviewPanel, extensionUri: Uri) {
+    this._panel = panel;
+    this._extensionUri = extensionUri;
+
+    // Aritificially limit the number of connections to 1 for now
```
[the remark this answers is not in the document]
daveajrussell: thanks, sorted

packages/vscode-extension/src/webviews/connectionPanel.ts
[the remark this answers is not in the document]
daveajrussell: yeah definitely something to consider - the change is already quite large, but it would be pragmatic

packages/vscode-extension/resources/scripts/connectionPanel.js
ncordon: Why not generate a unique uuid instead?

packages/schema-poller/src/testConnection.ts
```diff
@@ -0,0 +1,18 @@
+import { Config, Driver } from 'neo4j-driver';
+import { initializeDriver } from './driverInitializer';
+
+export async function testConnectionTransiently(
```
ncordon, line 4: Why do we need this if the plugin is already telling you it fails to connect to neo4j when trying to save a connection which credentials don't work?

ncordon:
I can't see anything under labels or relationships when I connect to a database, not sure why.

packages/language-server/src/types.ts
```diff
@@ -1,10 +1,12 @@
 // These settings are defined in the package.json
 export type Neo4jSettings = {
+  connectionKey: string;
```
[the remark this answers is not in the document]
daveajrussell: I'm going to remove this property from the Neo4jSettings type since it's not needed

but the connection key is a randomly generated 16 character slug when the connection is first created, it's not generated by the user and we don't allow it being modified

it will be used as the key in the connection key/value pair storage, so in global storage we'll have

```
    ' connections': {
        'example-connection-key': {
            ...
        },
        'another-key': {
            ...
        }
    }
```

packages/vscode-extension/e2e_tests/tests/commands.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+      CONNECTION_FAILED_MESSAGE,
+    );
+  });
+
+  test('Creating a valid connection should show a success message', async () => {
+    await vscode.commands.executeCommand(
+      SAVE_CONNECTION_COMMAND,
+      {
+        name: 'test-2',
+        key: 'test-2',
```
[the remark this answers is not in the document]
daveajrussell: I believe that we would overwrite the existing connection

packages/vscode-extension/package.json
```diff
@@ -64,25 +170,29 @@
...
+          "description": "Whether to connect to Neo4j for database aware completions.",
+          "deprecationMessage": "You can now use the connection panel to connect/disconnect from a database."
         },
         "neo4j.connectURL": {
           "order": 2,
           "type": "string",
           "default": "neo4j://localhost:7687",
-          "description": "URL of the Neo4j instance to poll data from, `bolt` & `neo4j` protocols supported."
+          "description": "URL of the Neo4j instance to poll data from, `bolt` & `neo4j` protocols supported.",
+          "deprecationMessage": "You can now use the connection panel to manage connections."
```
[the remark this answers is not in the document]
daveajrussell: sure thing

packages/vscode-extension/src/connectionPanel.ts
```diff
@@ -0,0 +1,250 @@
...
+      'scripts',
+      'connectionPanel.js',
+    );
+
+    const resetCssUri = webview.asWebviewUri(resetCssPath);
+    const vscodeCssUri = webview.asWebviewUri(vscodeCssPath);
+    const connectionPanelCssUri = webview.asWebviewUri(connectionPanelCssPath);
+    const connectionPanelJsUri = webview.asWebviewUri(connectionPanelJsPath);
+
+    const nonce = getNonce();
```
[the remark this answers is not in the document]
daveajrussell: the nonce is being used in the CSP header so that only scripts with a given nonce are allowed to execute within our extension

packages/vscode-extension/resources/scripts/connectionPanel.js
```diff
@@ -0,0 +1,71 @@
+(() => {
```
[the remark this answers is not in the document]
daveajrussell: just a plain js file to use in our webview.. I can look at moving this into a ts file and using the compiled output in our webview? or I can investigate some alternatives

packages/vscode-extension/src/connectionPanel.ts
```diff
@@ -0,0 +1,250 @@
...
+                  }" />
+                </div>
+                <div class="form--input-wrapper">
+                  <label for="scheme">Scheme</label>
+                  <select id="scheme">
+                    <option value="bolt://" ${
+                      this._connection?.scheme === 'bolt://' ? 'selected' : ''
+                    }>bolt://</option>
+                    <option value="neo4j://" ${
+                      this._connection?.scheme === 'neo4j://' ||
```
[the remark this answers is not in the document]
daveajrussell: yes, indeed - I took the mockup maybe a bit too literally

packages/schema-poller/src/testConnection.ts
daveajrussell: yes, I can switch to doing that

daveajrussell: I've been looking into this - I can see that we're logging a console error in the schema poller, the persistent connection method will then retry;

`Unable to connect to Neo4j: Neo4jError: The client is unauthorized due to authentication failure.. Retrying in 30 seconds.`

are you happy for the schema poller to abort the retry in the event of an authorization failure, return this back to the language server and, in turn, the language server to notify the extension that authorization has failed?

unless I'm missing something 😅

PR #222, opened 2024-06-11
Enables Cypher highlighting to be embedded in some programming languages

ncordon:
Adds highlighting within embedded strings in all the languages we have an official driver for:

* Java
* Python
* Typescript / Javascript
* Go
* .NET: C# and F#

It works with multiline strings which include a `//cypher` or `/*cypher*/` comment and with single line strings with a `//cypher` comment:

```java
String a = """// cypher

MATCH (n)
"""


String b = """/* cypher */

MATCH (n)
"""

String c = "//cypher MATCH (n)"
```

Also with templated strings in js/ts:

```js
/*cypher*/`
MATCH (n)
```

This is similar approach to https://marketplace.visualstudio.com/items?itemName=qufiwefefwoyn.inline-sql-syntax


OskarDamkjaer (approved):
Really cool to see the embedded highlighting working :100:

One observation, do we want to highlight the comment and quotes as well? 
CleanShot 2024-06-18 at 14 00 54

To me it'd be natural for the `"` to keep their old color and for the `/* cypher */` to render as a comment

OskarDamkjaer:
I think this change would warrant a changeset as well!

ncordon:
The changeset I'd prefer to introduce it in another pr because I'll have to rewrite the README, fix a detail in the release job, etc.

OskarDamkjaer:
Looking slick!! 💯

PR #224, opened 2024-06-11
[DRAFT] Add connection side panel and editor to vscode extension

daveajrussell:
Paired down approach, taken from https://github.com/neo4j/cypher-language-support/pull/221


packages/vscode-extension/src/webviews/controllers/connectionPanelController.ts
daveajrussell, line 1: This file is bundled using esbuild into an IIFE that's loaded into the webview. 

It binds a submit event listener to the form and performs some basic validation, before using the vscode API to post a message back to the connection panel webview where a listener is bound.

For now, I have hardcoded the connection name to 'Default connection' ('Default Connection' ?)

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -0,0 +1,251 @@
...
+                    <select id="scheme">
+                        <option value="bolt" ${
+                          this._connection?.scheme === 'bolt' ? 'selected' : ''
+                        }>bolt://</option>
+                        <option value="bolt+s" ${
+                          this._connection?.scheme === 'bolt+s'
+                            ? 'selected'
+                            : ''
+                        }>bolt+s://</option>
+                        <option value="neo4j" ${
```
daveajrussell, line 200: this sets the default selected option to neo4j when creating a new connection

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -0,0 +1,251 @@
...
+                  </div>
+                  <div class="form--input-wrapper">
+                    <label for="host">Host *</label>
+                    <input type="text" id="host" required placeholder="localhost" value="${
+                      this._connection?.host ?? ''
+                    }"/>
+                  </div>
+                  <div class="form--input-wrapper">
+                    <label for="port">Port</label>
+                    <input type="text" id="port" placeholder="7687" value="${
```
daveajrussell, line 221: port is not required for aura connections

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -0,0 +1,251 @@
...
+          </head>
+          <body>
+            <div class="container">
+              <div>
+                <h1>${
+                  this._connection ? 'Edit Connection' : 'Add New Connection'
+                }</h1>
+                <form class="form" action="" novalidate method="post">
+                  <input type="hidden" id="key" value="${
+                    this._connection?.key ?? getNonce(16)
```
daveajrussell, line 186: reuse our nonce function to generate a unique key

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -0,0 +1,251 @@
...
+        <html lang="en">
+          <head>
+            <meta charset="UTF-8">
+            <!--
+            Use a content security policy to only allow loading images from https or from our extension directory,
+            and only allow scripts that have a specific nonce.
+            -->
+            <meta http-equiv="Content-Security-Policy" content="img-src https: data:; style-src 'unsafe-inline' ${
+              webview.cspSource
+            }; script-src 'nonce-${nonce}';">
```
daveajrussell, line 169: we use a nonce to ensure that only scripts with this nonce are allowed to execute inside of our webview

packages/vscode-extension/src/languageClientService.ts
daveajrussell, line 1: wraps the language client and dispatches notifications for when the connection has been changed requiring the language server to take action

packages/vscode-extension/src/getNonce.ts
daveajrussell, line 1: a basic non-cryptographically secure function to generate a random alphanumeric string

packages/vscode-extension/src/contextService.ts
daveajrussell, line 1: wraps the the context and language client singletons so that we have a single place we can retrieve them

this also makes them more easily mockable for our tests

packages/vscode-extension/src/connectionTreeDataProvider.ts
daveajrussell, line 1: for now the treeview just displays our single connection with a label denoting whether or not it's connected

later on we'll be able to drill into our connections to see info about the connection

packages/vscode-extension/src/connectionService.ts
```diff
@@ -0,0 +1,148 @@
...
+  return {
+    ...Object.fromEntries(
+      Object.entries(connections).map(([key, connection]) => [
+        key,
+        { ...connection, connect: false },
+      ]),
+    ),
+    [connection.key]: connection,
+  };
+}
```
daveajrussell, line 121: this function ensures that we aren't mutating state and we created new instances of our connections when we add or modify them.

this function also resets the connected property to false when connections are modified to ensure that we only have one connected connection at a time

packages/vscode-extension/src/commandHandlers.ts
daveajrussell, line 1: I wanted to encapsulate any functionality that interacts with the presentation layer in a single place.. these handlers intercept commands and events, delegate any required business logic to other services, before showing notifications and modals to the user

packages/vscode-extension/e2e_tests/tests/connectionService.spec.ts
daveajrussell, line 1: this isn't really an e2e test, more a suite of unit tests.. it might be good to restructure the layout of the tests slightly, but perhaps for another changeset

packages/vscode-extension/src/contextService.ts
```diff
@@ -0,0 +1,31 @@
+import { Neo4jSettings } from '@neo4j-cypher/language-server/src/types';
+import { ExtensionContext } from 'vscode';
+
+type Neo4jLanguageClient = {
+  sendNotification: (method: string, settings: Neo4jSettings) => Promise<void>;
+};
```
daveajrussell, line 6: there might be a better way to achieve this.. but I defined this type so I could mock/stub the functionality of the LanguageClient more easily 

creating a mock object that extended the LanguageClient was proving troublesome.. there was some internal initialization that was tricky to provide dummy values for.. and since we currently only use one function of it in the extension.. sendNotification.. creating this type and using it in place of the LanguageClient for parameter types makes it a whole lot easier to stub

packages/vscode-extension/e2e_tests/main.ts
```diff
@@ -9,25 +10,24 @@ function setSetting(file: string, variable: RegExp, value: string) {
...
+  setSetting(dotenvFile, /\{PORT\}/g, port.toString());
+  setSetting(dotenvFile, /\{PASSWORD\}/g, password);
+  dotenv.config({ path: dotenvFile });
 }
 
 async function main() {
   const password = 'password';
 
-  const container = await new Neo4jContainer('neo4j:5')
+  const container = await new Neo4jContainer('neo4j:5-enterprise')
```
daveajrussell, line 26: enterprise is required for multiple databases

packages/schema-poller/src/schemaPoller.ts
```diff
@@ -76,11 +79,15 @@ export class Neo4jSchemaPoller {
     }
 
     this.reconnectionTimeout = setTimeout(() => {
-      void this.persistentConnect(url, credentials, config);
+      void this.persistentConnect(url, credentials, config, database);
     }, 30000);
   }
 
   disconnect() {
+    if (this.connection) {
```
daveajrussell, line 87: not strictly a required change, but it is handy to see in the console when we disconnect

packages/vscode-extension/resources/images/neo4j.png
daveajrussell, line 1: I've pinched this from upx - perhaps there's a brand library somewhere that has a preferred png

originally I had an svg from the NDL library - but! 

https://code.visualstudio.com/api/working-with-extensions/publishing-extension

> Note: Due to security concerns, vsce will not publish extensions that contain user-provided SVG images.
> The publishing tool checks the following constraints:
> - The badges provided in the package.json may not be SVGs unless they are from trusted badge providers

I assume this is to prevent svg bombing attacks

daveajrussell: I've since updated this with a proper SVG

ncordon (commented):
The image files don't seem to be added with lfs?

```
Encountered 3 files that should have been pointers, but weren't:
        packages/vscode-extension/resources/images/demo.gif
        packages/vscode-extension/resources/images/logo.png
        packages/vscode-extension/resources/images/neo4j.png
```

packages/schema-poller/src/neo4jConnection.ts
```diff
@@ -8,10 +8,21 @@ const METADATA_BASE = {
...
+    if (resolvedDatabase) {
+      return resolvedDatabase.name;
+    }
+  }
+
   const home = databases.find((d) => d.home);
   const def = databases.find((d) => d.default);
   const system = databases.find((d) => d.name === 'system');
+
   return home?.name ?? def?.name ?? system?.name ?? databases[0]?.name;
```
ncordon, line 26: Should we have this logic anymore? If we set an explicit connection to a specific database that does not exist, would this try to connect to the home one as a fallback? Maybe it's not what a user expects

ncordon: I think the current db should be set to undefined in that case and the healthcheck return false?

daveajrussell:
> The image files don't seem to be added with lfs?
> 
> ```
> Encountered 3 files that should have been pointers, but weren't:
>         packages/vscode-extension/resources/images/demo.gif
>         packages/vscode-extension/resources/images/logo.png
>         packages/vscode-extension/resources/images/neo4j.png
> ```

this should now be updated - I can see all 3 images in the repo are marked as stored with LFS

packages/schema-poller/src/neo4jConnection.ts
daveajrussell: yeah - that makes a lot of sense, have updated

packages/vscode-extension/src/webviews/connectionPanel.ts
ncordon: nice

ncordon (approved):
LGTM! Very good work on this

PR #225, opened 2024-06-17
Prototyping connection state

daveajrussell:
This PR prototypes the concept of connection state.

I am introducing an ``EventEmitter` to schema poller class that consumers may subscribe to to be notified of connection events. Currently there are 2.. Connected and Reconnecting.. Connected is fired when a driver is initialized and passes auth/connectivity checks. Reconnecting is fired when a driver fails to pass auth and connectivity checks. I've gone with Reconnecting since the schema poller will retry every 30 seconds until it succeeds or the driver is closed. 

In the extension UI, when you connect to a connection you are first presented with a "Connecting" state.. the label next to the connection will briefly read `connecting...` until the Connected event fires and we can update state, at which point we switch the label to `connected` and show a success info message. For connections that are unhealthy we will show a `reconnecting...` label and an accompanying error message. 

This state will automatically update itself when the schema next polls.. i.e. if a connection becomes unhealthy, the UI will receive an update.

This approach introduces a fair amount of complexity, the flow state of the application becomes harder to follow and it becomes harder to test e2e.


packages/vscode-extension/e2e_tests/tests/executeCommands.spec.ts
```diff
@@ -2,22 +2,32 @@ import { afterEach, beforeEach } from 'mocha';
 import * as sinon from 'sinon';
 import { commands, MessageOptions, window } from 'vscode';
 import { constants } from '../../src/constants';
+import { eventually } from '../helpers';
+import { testDatabaseKey } from '../suiteSetup';
 
-suite('Execute commands', () => {
+suite.skip('Execute commands', () => {
```
daveajrussell, line 8: these e2e tests unfortunately race with eachother now that the connected event we're interested in is asynchronous.. they can also interrupt with the main cypher e2e tests if we do not ensure that we reconnect back to the main connection.. if one of these tests executes in parallel with the e2e test that switches the database, I'd expect failures.. there's gotta be a better way to perform these e2e tests safely

daveajrussell:
[[VSCode Extension] Synchronicity between DBMS connection and extension](https://trello.com/c/wblTKW54/791-vscode-extension-synchronicity-between-dbms-connection-and-extension)

PR #227, opened 2024-06-19
Bumps VSCode version

ncordon:


.github/workflows/publish-vscode-extension.yaml
```diff
@@ -1,6 +1,7 @@
 name: Publish VSCode extension to Marketplace and OpenVSX
 
 on:
+  workflow_dispatch:
```
ncordon, line 4: Allows the workflow to be kicked off manually from the main branch until we can use changesets to at least generate vscode tags for us (for the trigger below to work automatically)

OskarDamkjaer (approved):

PR #232, opened 2024-06-26
Improve debug-ability of vscode extension

daveajrussell:
This PR improves the debug-ability of the vscode extension, both the e2e/unit tests and the extension itself.

I have extracted these changes out of the side panel PR.

- I am introducing an `esbuild.js` config file that only minifies if we pass a `--production` flag to it, and will otherwise produce a sourcemap for the extension
- I am adding some `dev` namespaced tasks in the extension package that will produce non-minified js and sourcemaps for the extension, which, both the vs code launch tasks will use
- I am adding a new launch task for debugging the extension tests
- I am modifying a few tests that use the `process.env` variables so that they don't fall over when debugging tests, though these are still dependent on you having an instance running locally

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -25,6 +25,24 @@
       "autoAttachChildProcesses": true,
       "sourceMaps": true,
       "preLaunchTask": "npm: build-vscode"
+    },
+    {
+      "type": "extensionHost",
+      "request": "launch",
+      "name": "Debug VSCode Tests",
+      "runtimeExecutable": "${execPath}",
+      "args": [
+        "--extensionDevelopmentPath=${workspaceRoot}/packages/vscode-extension",
+        "--extensionTestsPath=${workspaceRoot}/packages/vscode-extension/dist/e2e_tests/testRunner"
+      ],
+      "outFiles": [
+        "${workspaceRoot}/packages/vscode-extension/**/*.js",
+        "${workspaceRoot}/packages/language-support/**/*.js",
+        "${workspaceRoot}/packages/schema-poller/**/*.js"
+      ],
+      "autoAttachChildProcesses": true,
+      "sourceMaps": true,
+      "preLaunchTask": "npm: build-vscode"
     }
   ],
   "type": "pwa-node",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -20,7 +20,8 @@
       "outFiles": [
         "${workspaceRoot}/packages/vscode-extension/**/*.js",
         "${workspaceRoot}/packages/language-support/**/*.js",
-        "${workspaceRoot}/packages/schema-poller/**/*.js"
+        "${workspaceRoot}/packages/schema-poller/**/*.js",
+        "${workspaceRoot}/packages/language-server/**/*.js"
       ],
       "autoAttachChildProcesses": true,
       "sourceMaps": true,
@@ -38,7 +39,8 @@
       "outFiles": [
         "${workspaceRoot}/packages/vscode-extension/**/*.js",
         "${workspaceRoot}/packages/language-support/**/*.js",
-        "${workspaceRoot}/packages/schema-poller/**/*.js"
+        "${workspaceRoot}/packages/schema-poller/**/*.js",
+        "${workspaceRoot}/packages/language-server/**/*.js"
       ],
       "autoAttachChildProcesses": true,
       "sourceMaps": true,
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -33,7 +33,7 @@
       "runtimeExecutable": "${execPath}",
       "args": [
         "--extensionDevelopmentPath=${workspaceRoot}/packages/vscode-extension",
-        "--extensionTestsPath=${workspaceRoot}/packages/vscode-extension/dist/e2e_tests/testRunner"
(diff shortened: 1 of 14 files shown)


.vscode/launch.json
```diff
@@ -25,6 +25,24 @@
...
       "preLaunchTask": "npm: build-vscode"
+    },
+    {
+      "type": "extensionHost",
+      "request": "launch",
+      "name": "Debug VSCode Tests",
+      "runtimeExecutable": "${execPath}",
+      "args": [
+        "--extensionDevelopmentPath=${workspaceRoot}/packages/vscode-extension",
+        "--extensionTestsPath=${workspaceRoot}/packages/vscode-extension/dist/e2e_tests/testRunner"
```
daveajrussell, line 36: points to our tests entry point, ensures that tests are launched when the extension launches

package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "prepare": "husky install",
     "build": "turbo run build",
-    "build-vscode": "turbo run build --filter=neo4j-for-vscode",
+    "build-vscode": "turbo run build neo4j-for-vscode#build:dev --filter !neo4j-for-vscode",
```
daveajrussell, line 18: perhaps there's a better way to do this.. 

because we're dependent on a build to run the extension, and turbo runs tasks in parallel, I found that there was a race condition when both the build and build:dev tasks were running and one would overwrite the other

so here I am filtering out the vscode extension package from the build task, and running the new vscode build:dev task to ensure that only build:dev is ran on the vscode extension

ncordon: I'm not understanding how both `build` and `build#dev` tasks end up being run in parallel?

ncordon (commented):
Just some open questions

.vscode/launch.json
```diff
@@ -25,6 +25,24 @@
       "autoAttachChildProcesses": true,
       "sourceMaps": true,
       "preLaunchTask": "npm: build-vscode"
+    },
+    {
+      "type": "extensionHost",
+      "request": "launch",
+      "name": "Debug VSCode Tests",
```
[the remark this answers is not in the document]
daveajrussell: yes that is true, admittedly I didn't consider it and was just using my local instance, which may not be ideal in the long run

I'm playing with a few ideas to spin up a test container when we launch.. I've had _some_ success with creating a new task that spins one up.. the problem is tearing it down afterwards

 did you have any thoughts on how we could do this?

package.json
daveajrussell: alright, I've taken another look at this - I've reverted this back to simply `turbo run neo4j-for-vscode#build:dev` 

when I first had this configuration in, both the build and the build:dev tasks were running

now only the build:dev task is running as I'd expect.. I don't _really_ understand what changed, but this looks much better

.vscode/launch.json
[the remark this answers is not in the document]
daveajrussell: okay, what do you reckon to this change?


https://github.com/neo4j/cypher-language-support/pull/232/commits/85beda8d5e584378b90fa94ab7fcfc31c30faa49

I was finding that with the preLaunchTask, creating a testcontainer and loading config with dotenv was working fine, but I think where this fell down was the dotenv variables were not loaded into the process that launches vscode.. I was finding that the container was also ending prematurely.. I wondered if this had something to do with the container shutting down when process/thread used to run the preLaunchTask was killed/returned to the worker pool.

in this commit, I'm creating a debug version of the testRunner entry point that creates and starts a testcontainer for the tests to use when debugging.. I don't know if this is the cleanest way of achieving this.. but it seems to work well enough and keeps the concerns of running tests in a debugging context and CI/CLI context separate

.vscode/launch.json
```diff
@@ -25,6 +25,24 @@
...
+      "runtimeExecutable": "${execPath}",
+      "args": [
+        "--extensionDevelopmentPath=${workspaceRoot}/packages/vscode-extension",
+        "--extensionTestsPath=${workspaceRoot}/packages/vscode-extension/dist/e2e_tests/testRunnerDebug"
+      ],
+      "outFiles": [
+        "${workspaceRoot}/packages/vscode-extension/**/*.js",
+        "${workspaceRoot}/packages/language-support/**/*.js",
+        "${workspaceRoot}/packages/schema-poller/**/*.js"
```
ncordon, line 41: Add `"${workspaceRoot}/packages/language-server/**/*.js"` here so we can debug the events in the language server as well

.vscode/launch.json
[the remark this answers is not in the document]
ncordon: Nice!

I think the only thing left to do is close the container? I don't know if that's possible or we'd need to clean up already running containers before we spin up a new one?

.vscode/launch.json
daveajrussell: done :)

.vscode/launch.json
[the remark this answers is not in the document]
daveajrussell: yeah I was thinking about this

I'm observing that the container does seem to clean itself up once the tests finish and the instance of vscode closes.. like you said, I expect it might be sending a kill signal to child processes that's stopping the containers

it would be nice to explicitly close the container in code to avoid them sticking around

there must be a lifecycle hook we can use to do this

ncordon (approved):

.vscode/launch.json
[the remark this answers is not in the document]
ncordon: Happy with the process killing all children as it seems to be happening. Good work!

PR #233, opened 2024-06-27
execute single line query on enter by default

OskarDamkjaer:
This matches the behavior of the cypher editor in Browser by default


packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -157,9 +164,32 @@ export interface CypherEditorProps {
...
+          ? []
+          : [
+              {
+                key: 'Enter',
+                preventDefault: true,
+                run: (view: EditorView) => {
+                  const doc = view.state.doc.toString();
+                  if (doc.includes('\n')) {
+                    return false;
+                  }
```
ncordon, line 183: What's this doing?

packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -157,9 +164,32 @@ export interface CypherEditorProps {
...
+                  if (doc.trim() !== '') {
+                    onExecute(doc);
+                  }
+
+                  return true;
+                },
+              },
+            ]),
         {
           key: 'Ctrl-Enter',
```
ncordon, line 194: Do we want to leave this way of executing as well?

packages/react-codemirror/src/e2e_tests/historyNavigation.spec.tsx
```diff
@@ -113,6 +117,51 @@ RETURN n;`;
...
+  await editorPage.getEditor().fill('Misty');
+  await editorPage.getEditor().press('Enter');
+  expect(history.length).toBe(2);
+  expect(history).toEqual(['Misty', 'Brock']);
+
+  // Ensure cmd+enter is required in multiline
+  await editorPage.getEditor().fill(`multiline
+
+  `);
+  await editorPage.getEditor().press('Enter');
```
ncordon, line 154: Maybe add another one introducing the new line by pressing shift + Enter?

packages/react-codemirror/src/CypherEditor.tsx
OskarDamkjaer: I'll add a comment as well

packages/react-codemirror/src/e2e_tests/historyNavigation.spec.tsx
OskarDamkjaer: sure!

ncordon (commented):
This is missing the bump in `codemirror/view` right?

Also could you please rebase with master? I see the changes to the package-lock have an old version of the neo4j vscode plugin

packages/react-codemirror/src/e2e_tests/historyNavigation.spec.tsx
```diff
@@ -31,7 +31,37 @@ test('respects preloaded history', async ({ page, mount }) => {
...
+    useInnerText: true,
+  });
+
+  // Enter adds new lines
+  await editorPage.getEditor().fill('Brock');
+  await editorPage.getEditor().press('Enter');
+  await editorPage.getEditor().press('Enter');
+  await expect(editorComponent).toHaveText('1\n2\n3\nBrock', {
+    useInnerText: true,
+  });
```
ncordon, line 51: I'm not understanding this test. So we are in the case `newLineOnEnter = false` here, which means a single line + enter should execute the query, which is not what this is doing.

danieladugyan: It is true that `newLineOnEnter = false`, but `onExecute` is also not set in this test so we should _never_ execute the query. So single line + enter should _not_ execute the query.

As for testing the case where `onExecute` is not set and `newLineOnEnter = true`, I considered that as well, but this case is equivalent to when `newLineOnEnter = false`. It doesn't matter whether it is true or false, the editor behaves the same when `onExecute` is false. I almost think it's better  to change the component prop type so it's not allowed to set `newLineOnEnter = true` if `onExecute = false`.

danieladugyan:
> This is missing the bump in `codemirror/view` right?
> 
> Also could you please rebase with master? I see the changes to the package-lock have an old version of the neo4j vscode plugin

Yes I will make sure to rebase with master! As for the version bump, I can bump if you want but this PR solves the issue either way.

packages/react-codemirror/src/e2e_tests/historyNavigation.spec.tsx
ncordon: Well my point is we are not testing that code. Your reasoning is correct, but because you know how the code is going to behave.

ncordon:
> Yes I will make sure to rebase with master! As for the version bump, I can bump if you want but this PR solves the issue either way.

Yes please, bump the version.

packages/react-codemirror/src/e2e_tests/historyNavigation.spec.tsx
danieladugyan: I see what you are saying and I've added a test.

I still think it would be nice if we could prevent the user from setting `newLineOnEnter = true` if `onExecute = false` since this is arguably invalid, but I couldn't find a non-breaking way to enforce that with the TypeScript prop types.

ncordon (approved):
LGTM

PR #234, opened 2024-06-27
Add warnings on functions that don't exist in the database

angrykoala:
Similar to non-existing labels, a function that does not exist in the database will show as a warning

### DIFF
--- a/packages/language-support/src/parserWrapper.ts
+++ b/packages/language-support/src/parserWrapper.ts
@@ -8,6 +8,7 @@ import { _internalFeatureFlags } from './featureFlags';
 import {
   ClauseContext,
   default as CypherParser,
+  FunctionNameContext,
   LabelNameContext,
   LabelNameIsContext,
   LabelOrRelTypeContext,
@@ -40,6 +41,7 @@ export interface ParsedStatement {
   stopNode: ParserRuleContext;
   collectedLabelOrRelTypes: LabelOrRelType[];
   collectedVariables: string[];
+  collectedFunctions: ParsedFunction[];
 }
 
 export interface ParsingResult {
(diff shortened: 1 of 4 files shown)


ncordon (changes requested):
I think we are missing some edge cases:

* A test where there are spaces in between the dots `.`:

```
RETURN apoc    .     text     .    decapitalize    ("Marvin")
```


* A test to check the positions are calculated correctly when the function name is a escaped sequence:

```
RETURN     `some
multiline
thing`("Marvin")
```

* A test to check that this is still valid (because you can single escape individual parts of the functions):


```
RETURN `apoc`.coll.`avg`([1,2,3])
```

But this would get flagged as the function doesn't exist: ``RETURN `apoc.coll.avg`([1,2,3])``

* A test to check **only builtin** functions are case insensitive:

```
RETURN ABS(6.4)
```

So many edge cases we haven't tested or made the code work for yet

packages/language-support/src/parserWrapper.ts
```diff
@@ -267,6 +287,79 @@ class VariableCollector implements ParseTreeListener {
...
+        column: ctx.start.column,
+        offsets: {
+          start: ctx.start.start,
+          end: ctx.stop.stop + 1,
+        },
+      });
+    }
+  }
+
+  private getFunctionName(ctx: FunctionNameContext): string {
```
ncordon, line 337: Suggestion for making the code a bit more understandable: rename this to `getNormalizedFunctionName`

packages/language-support/src/parserWrapper.ts
```diff
@@ -92,6 +95,17 @@ export type LabelOrRelType = {
   };
 };
 
+export type ParsedFunction = {
+  name: string;
+  rawText: string;
```
ncordon, line 100: Maybe rename to `normalizedName` and `parsedName`?

packages/language-support/src/parserWrapper.ts
ncordon, line 337: Rename to `getNormalizedFunctionName` please

ncordon (approved):

PR #240, opened 2024-07-16
Listing databases for an active Connection

daveajrussell:
This change updates the connection tree view to include a dropdown for active connections that will list the database available for that connection.

The home database is annotated with a 'Home' emoji

The active database is highlighted green, and annotated with an 'active' label

To facilitate this, I am exposing the databases array that's established in the neo4j connection - my thinking is we needn't over complicate this by doing anything too clever, like automatically refreshing the list of databases when the schema poller refetches data since databases are likely to be fairly static


daveajrussell:
[[VSCode Extension] Listing databases for a Connection in the connection pane](https://trello.com/c/e5koPqDN/163-vscode-extension-listing-databases-for-a-connection-in-the-connection-pane)

PR #241, opened 2024-07-16
Adding a 'switch to database' command to databases in the tree view

daveajrussell:
This change updates the connection tree view to include a context menu option to databases in the treeview and accompanying command handler to switch databases. It also removes the optional `database` field in the connection form.

### DIFF
--- a/packages/vscode-extension/e2e_tests/tests/commandHandlers.spec.ts
+++ b/packages/vscode-extension/e2e_tests/tests/commandHandlers.spec.ts
@@ -1,16 +1,21 @@
 import { afterEach, beforeEach } from 'mocha';
 import * as sinon from 'sinon';
 import { ConfigurationChangeEvent } from 'vscode';
-import { handleNeo4jConfigurationChangedEvent } from '../../src/commandHandlers';
+import {
+  handleNeo4jConfigurationChangedEvent,
+  switchToDatabase,
+} from '../../src/commandHandlers';
 import * as connection from '../../src/connectionService';
 import { getMockConnection } from '../helpers';
 import { MockLanguageClient } from '../mocks/mockLanguageClient';
+import { MockSchemaPoller } from '../mocks/mockSchemaPoller';
 import { setupMockContextStubs } from '../mocks/setupMockContextStubs';
 
 suite('Command handlers spec', () => {
   let sandbox: sinon.SinonSandbox;
 
   let mockLanguageClient: MockLanguageClient;
+  let mockSchemaPoller: MockSchemaPoller;
 
   let getActiveConnectionStub: sinon.SinonStub;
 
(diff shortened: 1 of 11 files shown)


packages/vscode-extension/e2e_tests/tests/executeCommands.spec.ts
```diff
@@ -303,4 +303,45 @@ suite('Execute commands spec', () => {
...
+        { type: 'database', key: 'bad-database' },
+      );
+
+      sandbox.assert.calledWith(
+        showErrorMessageStub,
+        `${CONSTANTS.MESSAGES.ERROR_SWITCHING_DATABASE_MESSAGE} 'bad-database'. Unable to get a routing table for database 'bad-database' because this database does not exist. Double check that the database exists.`,
+      );
+    });
+
+    test('Switching to a non-database should do nothing', async () => {
```
ncordon, line 337: What do we mean here by non-database?

packages/vscode-extension/e2e_tests/tests/commandHandlers.spec.ts
```diff
@@ -88,4 +94,78 @@ suite('Command handlers spec', () => {
...
+      'connectionUpdated',
+      {
+        trace: { server: 'off' },
+        connect: true,
+        connectURL: 'neo4j://localhost:7687',
+        database: 'schemas',
+        user: 'neo4j',
+        password: null,
+      },
+    );
```
ncordon, line 126: Aren't we missing testing that database is an `activeDatabase` now?

ncordon (approved):

packages/vscode-extension/package.json
```diff
@@ -60,6 +60,10 @@
       {
         "command": "neo4j.saveConnection",
         "title": "Create Connection..."
+      },
+      {
+        "command": "neo4j.switchDatabase",
+        "title": "Switch to database..."
```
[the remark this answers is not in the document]
daveajrussell: yeah, I think this is unnecessary here

it's meant to signify that an additional action must be performed to complete this action https://uxpickle.com/significance-of-the-three-dots-or-ellipses-in-ui-design/#:~:text=On%20a%20button%2C%20they%20may,the%20text%20of%20the%20sentence.

so where I've added 3 dots after 'Delete Connection...', it's because when you click that you get a modal prompt to confirm deletion

but for switching, there is no additional action.. similarly, I don't think we show the 'Create Connection...' command anywhere in the UI so it can also be removed

packages/vscode-extension/e2e_tests/tests/commandHandlers.spec.ts
daveajrussell: we've got a test in the `connectionTreeDataProvider` spec that tests for this (`An active Connection should show the specified database as active`) if this is what you were thinking?

PR #242, opened 2024-07-16
Adding a database information pane to the connection side panel

daveajrussell:
This change updates the side panel to add a database information tree view. This new view lists the Labels and Relationships  of the active database.

I've opted to hook into our eventing flow.. since the db schema data is not populated deterministically when a connection is changed.. I've tried to keep this straightforward as possible.. by passing our `events` object from the schema poller to the metadata poller, and emitting a single `schemaFetched` event. The vscode extension listens to this event, and refreshes the database information pane. At which point the db schema data will be available on the schema poller reference.


daveajrussell:
[Add "Database Information" pane to connection panel](https://trello.com/c/WuJisOTa/165-add-database-information-pane-to-connection-panel)

packages/vscode-extension/src/connectionService.ts
```diff
@@ -462,8 +487,11 @@ function attachSchemaPollerConnectionFailedEventListeners(): void {
  * Attaches an event listener to handle connection errors.
  * This event is only handled once.
  */
-function attachSchemaPollerConnectionErrorEventListener(): void {
+function attachSchemaPollerConnectionEventListeners(): void {
   const schemaPoller = getSchemaPoller();
+  schemaPoller.events.on('schemaFetched', () => {
+    databaseInformationTreeDataProvider.refresh();
+  });
```
ncordon, line 494: Does this really guarantee the data inside the language server and the one shown in VSCode is in sync? Am I right understanding we have two schema pollers and one could not be in sync with the other?

daveajrussell: no, this only really guarantees that the vscode UI is in sync with the data that's been polled in the instance of the schema poller that it itself uses

you're right that we have 2 schema pollers, and the vscode extension acting as a kind of message broker for when the language server should make changes to its own schema poller instance (i.e. we've disconnected, or switched databases, etc)

I had this idea that we could build in a "runtime strategy" to the language server.. if it's being consumed by the vscode extension, it could share a schema poller instance, and hook in to the same events, as the extension. If it's running in a standalone way, it could create its own schema poller instance and operate in the same way that it currently does

packages/vscode-extension/src/connectionService.ts
```diff
@@ -286,15 +288,59 @@ export async function establishPersistentConnectionToSchemaPoller(
...
+    if (!database) {
+      return {
+        name: name,
+        default: false,
+        home: false,
+      };
+    } else {
+      return database;
+    }
+  }, []);
```
daveajrussell, line 323: This ensures that we maintain an up to date list of databases in the connection tree view.

Because the schema poller only returns a list of database names, we need to map these to the actual `Database` objects in our connection. 

If a database is not found by name, we can can construct a new `Database` object that our tree view can consume

packages/vscode-extension/src/connectionService.ts
```diff
@@ -286,15 +288,62 @@ export async function establishPersistentConnectionToSchemaPoller(
...
+ */
+export function getDbSchemaInformation(): {
+  labels: string[];
+  relationships: string[];
+} | null {
   const schemaPoller = getSchemaPoller();
-  return schemaPoller.connection?.databases ?? [];
+
+  if (!schemaPoller.metadata || !schemaPoller.metadata.dbSchema) {
+    return null;
```
ncordon, line 340: Suggestion to make this conformant to how most of the code of the repo is written: normally we return undefined instead of null

packages/vscode-extension/src/treeviews/databaseInformationTreeDataProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+    ];
+  }
+
+  private getDatabaseInformationElements(
+    element: DatabaseInformationItem,
+  ): DatabaseInformationItem[] {
+    const dbSchemaInformation = getDbSchemaInformation();
+
+    return element.type === 'label'
+      ? this.mapDbSchemaProperties('label', dbSchemaInformation.labels)
```
ncordon, line 67: Couldn't `dbSchemaInformation` be null here?

daveajrussell: yep, you're right - I'll add an undefined check here

ncordon (approved):
Tried it out manually and it works well for me

packages/vscode-extension/src/connectionService.ts
```diff
@@ -286,15 +288,62 @@ export async function establishPersistentConnectionToSchemaPoller(
...
+  }
+
+  return {
+    labels: schemaPoller.metadata?.dbSchema.labels,
+    relationships: schemaPoller.metadata?.dbSchema.relationshipTypes,
```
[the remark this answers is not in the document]
daveajrussell: looking at this function again, I've already got a guard clause above this return, so it doesn't seem necessary to null check `metadata` 

```
  if (!schemaPoller.metadata || !schemaPoller.metadata.dbSchema) {
    return undefined;
  }

  return {
    labels: schemaPoller.metadata.dbSchema.labels,
    relationships: schemaPoller.metadata.dbSchema.relationshipTypes,
  };
```

packages/vscode-extension/src/connectionService.ts
daveajrussell: good shout :)

PR #243, opened 2024-07-17
Update logic for handling externally set values

danieladugyan:
Fix a bug causing debounced value updates to get cancelled erroneously


danieladugyan:
[Failing E2E tests after bumping up Codemirror version](https://trello.com/c/V0vcsnuO/820-failing-e2e-tests-after-bumping-up-codemirror-version)

PR #245, opened 2024-07-18
Adds errors for non-declared procedures

angrykoala:
Extends #234 with the same ~warnings~ errors for procedures


packages/language-support/src/parserWrapper.ts
```diff
@@ -39,11 +40,12 @@ export interface ParsedStatement {
   // A statement needs to be parsed with the .statements() rule because
   // it's the one that tries to parse until the EOF
   ctx: StatementsOrCommandsContext;
-  diagnostics: SyntaxDiagnostic[];
+  syntaxErrors: SyntaxDiagnostic[];
```
OskarDamkjaer, line 43: why rename this? I think even if "syntaxErrors" could be more desriptive, it's nice to match the type name imo

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -312,3 +356,25 @@ function warnOnUndeclaredFunctions(
...
+    const proceduresInQuery = parsingResult.collectedProcedures;
+
+    proceduresInQuery.forEach((parsedProcedure) => {
+      const error = detectNonDeclaredProcedure(
+        parsedProcedure,
+        dbSchema.procedures,
+      );
+
+      if (error) errors.push(error);
+    });
```
OskarDamkjaer, line 376: The extra layer of `detectNonDeclaredProcedure` isn't very useful imo, I think it's easier to understand and better to inline it. That way we can also remove one of the if statements:

```suggestion
    proceduresInQuery.forEach((parsedProcedure) => {
      const procedureExists = Boolean(proceduresSchema[parsedProcedure.name]);
      if (!procedureExists) {
         errors.push(generateProcedureNotFoundError(parsedProcedure))
      }
}
```

OskarDamkjaer (approved):
Only got some code style/structure remarks, impl is good to go from my point over view :+1:

packages/language-support/src/parserWrapper.ts
ncordon: It makes the code more readable in my opinion. Especially when we are using web workers and we need that logic of: no syntax errors, then execute the web worker. It reads better than diagnostics.

Another example is what has changed in `syntaxValidation.ts`:

```ts
    const anySyntacticError =
      statements.filter((statement) => statement.syntaxErrors.length !== 0)
        .length > 0;
```

We know what's happening inside better than if it read diagnostics. 

What we could do is rename this type:
```ts
export type SyntaxDiagnostic = Diagnostic & {
  offsets: { start: number; end: number };
};
```

where `Diagnostic` is coming from the `vscode-types`. Any suggestions?

OskarDamkjaer: I don't feel very strongly about it & agree syntaxError reads better. Part of me was assuming there could be more than errors from a "Diagnostics" though

OskarDamkjaer: It's beside the point but for the code example you could use `[some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)` instead, it's more efficient as it stops iterating as soon as it finds a match 

```ts
const anySyntacticError =
      statements.some((statement) => statement.syntaxErrors.length !== 0)
```

packages/language-support/src/syntaxValidation/syntaxValidation.ts
ncordon: done

packages/vscode-extension/syntaxes/cypher.json
```diff
@@ -38,7 +38,7 @@
...
...|FULLTEXT|FUNCTION|FUNCTIONS|GRANT|GRAPH|GRAPHS|GROUP|GROUPS|HEADERS|HOME|ID|IF|IMMUTABLE|IMPERSONATE|IN|INDEX|INDEXES|INF|INFINITY|INSERT|INT|INTEGER|IS|JOIN|KEY|LABEL|LABELS|LEADING|LIMIT|LIST|LOAD|LOCAL|LOOKUP|MANAGEMENT|MAP|MATCH|MERGE|NAME|NAMES|NAN|NEW|NFC|NFD|NFKC|NFKD|NODE|NODETACH|NODES|NONE|NORMALIZE|NORMALIZED|NOT|NOTHING|NOWAIT|NULL|OF|ON|ONLY|OPTION|OPTIONAL|OPTIONS|OR|ORDER|OUTPUT|PASSWORD|PASSWORDS|PATH|PATHS|PERIODIC|PLAINTEXT|POINT|POPULATED|PRIMARY|PRIMARIES|PRIVILEGE|PRIVILEGES|PROCEDURE|PROCEDURES|PROPERTIES|PROPERTY|PROVIDER|PROVIDERS|RANGE|READ|REALLOCATE|REDUCE|REL|RELATIONSHIP|RELATIONSHIPS|REMOVE|RENAME|REPEATABLE|REPLACE|REPORT|REQUIRE|REQUIRED|RETURN|REVOKE|ROLE|ROLES|ROW|ROWS|SCAN|SECONDARY|SECONDARIES|SEC|SECOND|SECONDS|SEEK|SERVER|SERVERS|SET|SETTING|SETTINGS|SHORTEST|shortestPath|SHOW|SIGNED|SINGLE|SKIP|START|STARTS|STATUS|STOP|VARCHAR|STRING|SUPPORTED|SUSPENDED|TARGET|TERMINATE|TEXT|THEN|TIME|TIMESTAMP|TIMEZONE|TO|TOPOLOGY|TRAILING|TRANSACTION|TRANSACTIONS|TRAVERSE|TRIM|TRUE|TYPE|TYPED|TYPES|UNION|UNIQUE|UNIQUENESS|UNWIND|URL|USE|USER|USERS|USING|VALUE|VECTOR|VERBOSE|VERTEX|WAIT|WHEN|WHERE|WITH|WITHOUT|WRITE|XOR|YIELD|ZONE|ZONED|EXPLAIN|PROFILE)\\b",
```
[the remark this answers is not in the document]
ncordon: It's just the `ZONE` keyword, but we haven't touched anything in this pr. I think it's because we don't have a CI job making sure that grammar has been regenerated when we make changes.

PR #247, opened 2024-07-25
Prototyping a runtime strategy to allow different runtime behaviours in the language server

daveajrussell:
Currently with our vscode extension, we have 2 instances on the schema poller.. 1 in the vscode extension, that is used for connecting to and listing databases and their properties. The other is in the language server itself, and is used by the language server for database aware features. 

The vscode extension brokers messages via the language client for language server to update its own schema poller instance when the connection in the extension is modified.

There are a few problems with this approach;
- we are opening more connections than are necessary to a neo4j instance 
- the drivers may individually become unhealthy, leading to the schema pollers falling out of sync

In this prototype I am;

**Creating the concept of a 'Runtime' in the language server**
This does make the language server somewhat less client agnostic, as this approach makes it aware of the client that is consuming it.

However it does allow us to configure the language server to manage its own instance of the schema poller, or allow schema updates to be deferred to the consumer.

In the standalone runtime, we construct a schema poller instance and establish a persistent connection to it.

In the extension runtime, we establish a single listener to a `schemaFetched` message, that updates an internal dbSchema instance which the database aware features will use. This message will be dispatched from the vscode extension when it receives an event from its schema poller instance.

`schema poller:emit('schemaFetched') -> vscode extension:languageServer.sendNotification('schemaFetched', {}) -> language server:onNotification('schemaFetched', (dbSchema) => {})`

It would be nice to share the actual schema poller instance that the vscode extension creates between the vscode extension and the language server to avoid the need for having to rebroadcast a message from the extension, but this does not seem possible.. parameters may be passed via the `initializationOptions`, however these are serialized and deserialized, and the event emitter doesn't pick up messages inside of the language server.


OskarDamkjaer:
I like this approach/solution to the problem, I feel it makes it simpler to understand how the connection to neo4j works :+1:

PR #250, opened 2024-07-26
Migrate from jest to vitest

danieladugyan:
We've seen the need for good tests for the debounce behaviour in the CypherEditor. These tests are better suited for unit testing that E2E Playwright tests. However, using Jest proved problematic for React component testing. It can't resolve Vite's worker imports natively and has no support for ES modules. Therefore, this PR replaces Jest with Vitest in preparation for an upcoming PR that adds unit tests for the CypherEditor.

Note especially the changes in `completionAssertionHelpers.ts`. I think the test was incorrectly written before and happened to work due to Jest's implementation of array equality checking. However, it caused the tests to fail in Vitest.


danieladugyan:
[Migrate from Jest to Vitest](https://trello.com/c/t9A8YMpe/174-migrate-from-jest-to-vitest)

daveajrussell (approved):
Looks good to me.. running the entire suite locally is working nicely for me

ncordon (changes requested):
A couple of comments:


* `vscode-extension` is showing on the VSCode testing panel, it would be good if we could filter it out from there because it has no unit tests:

danieladugyan:
> A couple of comments:
> 
> * `vscode-extension` is showing on the VSCode testing panel, it would be good if we could filter it out from there because it has no unit tests:
> 
> * I'm getting a warning:
> 
> `The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.`
> 
> when I run `npm test`. What implications does this have? Because as I understand it we want to keep supporting CJS in the future?

@ncordon I have fixed the first issue by hard-coding the paths to the packages that have unit-tests (in `vitest.workspace.ts`). I fixed the second issue by making sure Vite is running in ESM mode everywhere. AFAIK this will have no impact on whether we can keep on supporting CJS, it's just about whether Vite is running in CJS or ESM mode.

packages/language-support/src/tests/autocompletion/patternCompletion.test.ts
```diff
@@ -74,7 +74,7 @@ describe('MATCH auto-completion', () => {
     });
   });
 
-  test.only('Correctly completes unstarted label when caret is passed and there is a space', () => {
```
ncordon, line 77: good catch 👍

ncordon (approved):
Nice work!

PR #254, opened 2024-07-31
Replace ts-ignore with proper Vite types

danieladugyan:
I noticed a minor stylistic issue in how a worker was imported. Fixed by adding the [Vite client types](https://vitejs.dev/guide/features#client-types).


OskarDamkjaer (approved):
🙏

PR #255, opened 2024-08-01
Adds a webview template engine for easily creating webviews

daveajrussell:
Introducing a webview template engine that defines a common html document and stylesheets.

Webviews can be configured to include additional script files and stylesheets for their own purposes.


ncordon (commented):
LGTM, waiting on #253

PR #257, opened 2024-08-09
Archives VSCode artifacts

ncordon:
Archives VSCode artifacts when merging to `main`. The result can be seen in [0b79209](https://github.com/neo4j/cypher-language-support/pull/257/commits/0b79209cd60f4d5b1ab87734a5b779f9bf9a31fa)

### DIFF
--- a/.github/workflows/archive-vscode-artifacts.yaml
+++ b/.github/workflows/archive-vscode-artifacts.yaml
@@ -1,0 +1,38 @@
+on:
+  push:
+    branches:
+      - main
+      - archive-vscode-artifacts
+
+env:
+  NODE_OPTIONS: '--max_old_space_size=8192'
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/archive-vscode-artifacts.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Build project
+        env:
+          NODE_OPTIONS: '--max_old_space_size=8192'
+        run: npm run build
+
+      - name: Generate VSCode plugin package
+        run: |
+          cd packages/vscode-extension
+          npx vsce package
```
daveajrussell, line 30: could we use the --pre-release flag here?

daveajrussell (approved):
Tested by installing the artefact - looks good!

PR #258, opened 2024-08-20
Add row limit to query execution

OskarDamkjaer:
Adds a new cypher running function for user queries instead of the one created for Neo4j SDK functions. It's using the [AsyncIterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncIterator) api in the driver to stop consuming records after a set limit is hit. I also update the view to inform the user if the limit was hit.

I got an error message from Turborepo that told me to update the version 🤷


ncordon (approved):
LGTM

PR #259, opened 2024-08-21
Adds standalone completion API needed by cypher-shell

ncordon:


ncordon:
Closing this as we are not going to go the graaljs route 😢

PR #260, opened 2024-08-26
Make debounce tests pass on mac

OskarDamkjaer:
Use `enter` instead of `Control+Enter` since the latter keybinding only works on linux.


packages/react-codemirror/src/e2e_tests/sanityChecks.spec.tsx
```diff
@@ -32,18 +32,14 @@ test('the editor can report changes to the text ', async ({ mount, page }) => {
...
   const textField = page.getByRole('textbox');
 
+  await textField.fill('');
   await textField.fill('RETURN 12');
+  await expect(textField).toHaveText('RETURN 12');
 
   // editor update is debounced, retry wait for debounced
   await expect(() => {
     expect(editorValueCopy).toBe('RETURN 12');
   }).toPass({ intervals: [300, 300, 1000] });
```
ncordon, line 42: I don't completely understand, where's the Enter used instead of Control + Enter?

packages/react-codemirror/src/e2e_tests/sanityChecks.spec.tsx
```diff
@@ -32,18 +32,14 @@ test('the editor can report changes to the text ', async ({ mount, page }) => {
...
-
-  await page.keyboard.type('34');
-
-  await expect(() => {
-    expect(editorValueCopy).toBe('RETURN 12');
-  }).toPass({ intervals: [300, 300, 1000] });
 });
```
ncordon, line 43: Why have we removed this?

packages/react-codemirror/src/e2e_tests/sanityChecks.spec.tsx
OskarDamkjaer: this test doesn't use enter, it's a leftover from fiddling with reducing flakyness, I can remove it and perhaps include it in more when looking into flakyness again

ncordon (approved):

PR #268, opened 2024-09-23
Enables dynamic labels and properties

ncordon:


packages/language-support/src/antlr-grammar/Cypher5Lexer.g4
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-lexer grammar CypherLexer;
+lexer grammar Cypher5Lexer;
```
OskarDamkjaer, line 17: do we need/want to do this rename before we've done the spikes into cypher 5/6 mgmt and decided how we're going to handle that?

OskarDamkjaer (approved):

packages/language-support/src/antlr-grammar/Cypher5Lexer.g4
ncordon: This is the format they come as from the monorepo, so every time I was copying over I had to rename.

PR #269, opened 2024-10-02
Enable semantic token highlighting

jamthief:


ncordon:
I've tried this and it works well 👍 

intellij

PR #274, opened 2024-10-23
Expose tab focus mode prop so we can override tab-trapping behaviour

daveajrussell:
It would be useful to expose this property so that we are able to override the tab behaviour of the editor.

By default in the neo4j setup, we set up handlers for Tab and Shift-Tab - however this traps the user in the editor. Whilst it's then possible to use Esc to remove focus from the editor, in some circumstances (e.g. the editor is in a modal where Esc closes the modal), it'd be nice to be able to disable tab trapping - even though this does mean disabling Tab for accepting autocompletions

https://codemirror.net/docs/ref/#view.EditorView.setTabFocusMode

### DIFF
--- a/packages/react-codemirror/src/CypherEditor.tsx
+++ b/packages/react-codemirror/src/CypherEditor.tsx
@@ -166,14 +166,14 @@ export interface CypherEditorProps {
   ariaLabel?: string;
 
   /**
-   * Whether to enable or disable tab focus mode.
+   * Whether keybindings for inserting indents with the Tab key should be disabled.
    *
-   * true will allow the brower's default tabbing behavior.
-   * false will enable tab-trapping within the editor.
+   * true will not create keybindings for inserting indents.
+   * false will create keybindings for inserting indents.
    *
    * @default false
    */
-  setTabFocusMode?: boolean;
+  moveFocusOnTab?: boolean;
 }
 
 const executeKeybinding = (
@@ -320,7 +320,7 @@ export class CypherEditor extends Component<
     theme: 'light',
     lineNumbers: true,
     newLineOnEnter: false,
-    setTabFocusMode: false,
+    moveFocusOnTab: false,
   };
 
   private debouncedOnChange = this.props.onChange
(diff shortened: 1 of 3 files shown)


daveajrussell:
[Can not navigate to save button using keyaboard when editing saved cypher](https://trello.com/c/xiO7ZnFn/1039-can-not-navigate-to-save-button-using-keyaboard-when-editing-saved-cypher)

packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -161,9 +161,19 @@ export interface CypherEditorProps {
...
+
+  /**
+   * Whether to enable or disable tab focus mode.
+   *
+   * true will allow the brower's default tabbing behavior.
+   * false will enable tab-trapping within the editor.
+   *
+   * @default false
+   */
+  setTabFocusMode?: boolean;
```
OskarDamkjaer, line 176: The default behavior of the codemirror editor is for tab to move focus, I think it'd make sense to conditionally remove the override:

```
    {
      key: 'Tab',
      preventDefault: true,
      run: insertTab,
    },
    {
      key: 'Shift-Tab',
      preventDefault: true,
      run: indentLess,
    },
 ```
 
 Instead of overriding the override. 
 
 Imo it's fine if it's not a reactive setting (i don't see when you'd dynamically toggle this on and off), so the boolean prop can just be passed to the `basicNeo4jSetup`. call in the `componentDidMount`. Not sure if the current solution also does it, but this way tab would still accept completions for example

OskarDamkjaer: If we want don't want to increase the prop surface, we could also call `this.editorView.current.setTabFocusMode` from outside as it's reachable outside the component through a ref i think

OskarDamkjaer: I think a `moveFocusOnTab` (or simliarly named) defaulting to false prop could make senes

daveajrussell: makes total sense

based on our discussion I've renamed the property we're exposing here, removed the setTabFocusMode setting and instead have used the new property to conditionally set the tab keybindings in the setup module

from testing, I found that the tab to accept completions still trapped the focus inside the editor - whilst it was possible to shift-tab _out_ of the editor, tabbing took you straight back in.. so I've included this in the conditional

OskarDamkjaer (approved):

PR #276, opened 2024-10-28
Updates grammar and semantic analysis to neo4j 5.25

ncordon:


modify-semantic-analysis.js
```diff
@@ -0,0 +1,59 @@
+var path = require('path');
```
OskarDamkjaer, line 1: this is already part of the other pr right?

OskarDamkjaer (approved):

modify-semantic-analysis.js
ncordon: Good catch 😬

PR #280, opened 2024-11-03
Fixes signature help

ncordon:
Some work has gone into updating descriptions of procedures and functions so it has broken the signature help functionality for newer versions. Our e2e tests were already failing because of that.

## After
old-vscode-behaviour

### With older neo4j versions

In VSCode:

In codemirror:

### With newer neo4j versions

In VSCode:

In codemirror:

### DIFF
--- a/packages/react-codemirror/src/e2e_tests/signatureHelp.spec.tsx
+++ b/packages/react-codemirror/src/e2e_tests/signatureHelp.spec.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/unbound-method */
 import { testData } from '@neo4j-cypher/language-support';
 import { expect, test } from '@playwright/experimental-ct-react';
 import { Locator } from 'playwright/test';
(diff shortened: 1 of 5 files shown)


packages/react-codemirror/src/e2e_tests/signatureHelp.spec.tsx
```diff
@@ -103,8 +107,8 @@ test('Signature help shows the description for the first argument when the curso
 
   await testTooltip(tooltip, {
     includes: [
-      'nodes :: LIST<MAP>',
-      'Imports `NODE` and `RELATIONSHIP` values with the given labels and types from the provided CSV file',
+      importCsvProc.argumentDescription[0].description,
+      importCsvProc.description,
```
ncordon, line 111: Just makes these tests more mantainable by not having to update strings all the time

packages/react-codemirror/src/e2e_tests/signatureHelp.spec.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/unbound-method */
```
ncordon, line 1: When building from VSCode, the linter was complaining about the ` async ({ page, mount }) => {` parts in the tests

OskarDamkjaer (approved):

PR #281, opened 2024-11-04
Adds feature to run selection in VSCode

ncordon:
## Why

If we select some specific text in VSCode, we want to run only that part of the query.

## Considerations

One thing I don't quite like is that the order in which the statements would be executed depends on what order you highlight them. Should we maintain that?

### DIFF
--- a/packages/vscode-extension/src/commandHandlers.ts
+++ b/packages/vscode-extension/src/commandHandlers.ts
@@ -199,17 +199,13 @@ export async function runCypher(): Promise<void> {
     const documentUri = editor.document.uri;
 
     if (selections.length === 0) {
-      const documentText = editor.document.getText();
-      await cypherRunner.run(activeConnection, documentUri, documentText);
+      const text = editor.document.getText();
+      await cypherRunner.run(activeConnection, documentUri, Array.of(text));
     } else {
-      const statements = selections.map((selection) => {
+      const text = selections.map((selection) => {
         return editor.document.getText(selection);
       });
-      await cypherRunner.runStatements(
-        activeConnection,
-        documentUri,
-        statements,
-      );
+      await cypherRunner.run(activeConnection, documentUri, text);
     }
   }
 }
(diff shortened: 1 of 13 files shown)


OskarDamkjaer:
I agree it'd be better to run from top to bottom, rather than in selection order

packages/vscode-extension/src/commandHandlers.ts
```diff
@@ -192,8 +192,20 @@ export async function runCypher(): Promise<void> {
...
+    const selections = editor.selections.filter(
+      (selection) => !selection.isEmpty && editor.document.getText(selection),
+    );
+
     const documentUri = editor.document.uri;
-    await cypherRunner.run(activeConnection, documentUri, documentText);
+
+    if (selections.length === 0) {
+      const text = editor.document.getText();
+      await cypherRunner.run(activeConnection, documentUri, Array.of(text));
```
OskarDamkjaer, line 203: It works but I've never seen anyone do `Array.of(stringValue)` instead of `[stringValue]`

packages/vscode-extension/tests/webviewUtils.ts
```diff
@@ -16,19 +16,36 @@ export async function waitUntilNotification(
...
+      // eslint-disable-next-line
+      const editor = await vscode.window.showTextDocument(document);
+
+      if (options.selectLines && options.selectLines > 0) {
+        const lastLine = options.selectLines - 1;
+        // eslint-disable-next-line
+        editor.selection = new vscode.Selection(
+          // eslint-disable-next-line
+          new vscode.Position(0, 0), // Start at the beginning of the first line
+          // eslint-disable-next-line
```
OskarDamkjaer, line 42: I think it's nice to know what specific rule gets disabled, then it's easier to remove the eslint comment later, but I'm fine to keep it as well if you think it's too unruly

OskarDamkjaer (approved):

PR #284, opened 2024-11-07
Fixes signature help bug with default arguments in VSCode

ncordon:

### DIFF
--- a/packages/language-support/src/signatureHelp.ts
+++ b/packages/language-support/src/signatureHelp.ts
@@ -36,15 +36,35 @@ interface ParsedMethod {
 export function toSignatureInformation(
   curr: Neo4jFunction | Neo4jProcedure,
 ): SignatureInformation {
-  const { signature, argumentDescription, description } = curr;
+  const { name, argumentDescription, description } = curr;
+  const argDescriptions = argumentDescription.map((arg) => {
+    let label = '';
+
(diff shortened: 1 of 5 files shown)


packages/language-support/src/signatureHelp.ts
```diff
@@ -36,15 +36,35 @@ interface ParsedMethod {
...
+    }
+
+    return {
+      label: label,
+      documentation: arg.description,
+    };
+  });
+
+  const argsString = argDescriptions.map((arg) => arg.label).join(', ');
+  const signature = `${name}(${argsString})`;
```
ncordon, line 62: This ensures the signature is computed with whatever is returned in the argument descriptions

packages/language-support/src/signatureHelp.ts
```diff
@@ -36,15 +36,35 @@ interface ParsedMethod {
...
 
   return SignatureInformation.create(
     signature,
     description,
-    ...argumentDescription.map((arg) => ({
-      label: `${arg.name} :: ${arg.type}`,
```
ncordon, line 45: The problem is that the highlighting on the signature gets done matching against this. If the signature of a specific argument includes a default value (e.g `= ms`) but here we didn't have it, then it wasn't been highlighted in blue.

OskarDamkjaer (approved):

PR #293, opened 2024-11-18
Add support for sysinfo, welcome, connect, disconnect and server commands

daveajrussell:
This PR adds language support for a number of new commands.

`:welcome` - creates a welcome frame in the stream
`:connect` - creates a connection frame in the stream, where users can trigger a connection flow
`:disconnect` - disconnects the current connect
`:server connect` - alias for `:connect`
`:server disconnect` - alias for `:disconnect`
`:sysinfo` - not yet implemented, but will be used to display a database system info frame

I have added some highlighting overrides for `:server` so that we differentiate between the console command `:server` and the keyword `SERVER`


daveajrussell:
[Add `:connect`, `:welcome`, `:disconnect`, `:server disconnect`, `:server connect` to the language support grammar](https://trello.com/c/u1HA98FC/1073-add-connect-welcome-disconnect-server-disconnect-server-connect-to-the-language-support-grammar)

packages/language-support/src/tests/consoleCommands.test.ts
```diff
@@ -93,11 +97,82 @@ describe('sanity checks', () => {
...
+          line: 0,
+          startCharacter: 1,
+          startOffset: 1,
+        },
+        token: 'disconnect',
+        tokenType: 'consoleCommand',
+      },
+    ]);
+    expect(applySyntaxColouring(':sysinfo')).toEqual([
+      {
```
ncordon, line 145: Given `server` is the only especial command (at colouring level), could you please add a test for that?

daveajrussell: I've got a  `:server` suite below, which includes some syntax colouring tests, are these covering what you're asking?

PR #295, opened 2024-11-20
Fixes bug with preparser options making semantic errors not show up correctly

ncordon:
## What

It only extracts the statement in the command parsing, omitting the preparsing options.

## Why

We were not showing semantic errors correctly for queries that included a preparsing option, because the semantic analysis does not have a preparser inside, it expects already a pre-processed query.

```
EXPLAIN MATCH (n)
// It should be showing an error in the MATCH (n) here because we are missing a RETURN
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22570,10 +22570,10 @@
     },
     "packages/language-server": {
       "name": "@neo4j-cypher/language-server",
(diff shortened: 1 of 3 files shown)


anderson4j (approved):
Looks good to me!

PR #297, opened 2024-11-22
Removes the vendored antlr4 dependency

ncordon:
Removes the vendored antlr4 dependency.


anderson4j (approved):
Guessing we dont need the dependency since we have our own version of antlr4 locally? Didn't see anything that popped out as odd to me (although there's so many removals it's not really feasible to check what they all do individually) so I trust that you know what to remove and what not to.

PR #298, opened 2024-11-27
(Very minor) Using Set over List for already given return names in YIELD

anderson4j:
Updates existing change adding YIELD autocompletions so that we check suggestions against existing elements stored in a Set (Hash-set) so that it's reliably quick in case of large amounts of return names.

### DIFF
--- a/.changeset/old-shoes-provide.md
+++ b/.changeset/old-shoes-provide.md
@@ -1,0 +1,5 @@
+---
+'@neo4j-cypher/language-support': patch
+---
+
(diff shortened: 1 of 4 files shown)


ncordon:
Solve the conflicts please, the PR is showing more changes than the ones from commit  3a769970b75

anderson4j:
Rebasing was unnecessarily messy + changeset shouldnt be repeated. Just moved relevant commit to it's own branch and remade the PR here https://github.com/neo4j/cypher-language-support/pull/302 (so will close this one)

PR #299, opened 2024-11-27
Alter so labels with numbers and underscores dont get needless backticks

anderson4j:
Before if for example had relationship rel_123, we autocompleted `MATCH (n)-[:rel^]` to
` MATCH (n)-[:`\``rel_123`\``] `
Now instead we dont backtick if we have underscore and numbers in label names/rel names/properties and just give
`MATCH (n)-[:rel_123]`

### DIFF
--- a/packages/language-support/src/autocompletion/completionCoreCompletions.ts
+++ b/packages/language-support/src/autocompletion/completionCoreCompletions.ts
@@ -32,7 +32,7 @@ import { CompletionItem, Neo4jFunction, Neo4jProcedure } from '../types';
 const uniq = <T>(arr: T[]) => Array.from(new Set(arr));
 
 function backtickIfNeeded(e: string): string | undefined {
-  if (/\s/.test(e) || /[^a-zA-Z]/.test(e)) {
+  if (/\s/.test(e) || /[^a-zA-Z0-9_]/.test(e)) {
     return `\`${e}\``;
   } else {
     return undefined;
--- a/packages/language-support/src/autocompletion/completionCoreCompletions.ts
+++ b/packages/language-support/src/autocompletion/completionCoreCompletions.ts
@@ -32,7 +32,7 @@ import { CompletionItem, Neo4jFunction, Neo4jProcedure } from '../types';
 const uniq = <T>(arr: T[]) => Array.from(new Set(arr));
 
 function backtickIfNeeded(e: string): string | undefined {
-  if (/\s/.test(e) || /[^a-zA-Z0-9_]/.test(e)) {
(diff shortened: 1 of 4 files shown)


packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -32,7 +32,7 @@ import { CompletionItem, Neo4jFunction, Neo4jProcedure } from '../types';
 const uniq = <T>(arr: T[]) => Array.from(new Set(arr));
 
 function backtickIfNeeded(e: string): string | undefined {
-  if (/\s/.test(e) || /[^a-zA-Z]/.test(e)) {
+  if (/\s/.test(e) || /[^a-zA-Z0-9_]/.test(e)) {
```
ncordon, line 35: There's an edge case we are not really handling well here: a label, rel type or property key that starts with a number still needs backticks

anderson4j: Also checked with chatGPT and apparently non-english letters is okay anywhere in the names, will fix both at the same time

ncordon (changes requested):
Looks good to me but I'd like to see a case where the label / rel type starts by underscore added to the test cases

packages/language-support/src/autocompletion/completionCoreCompletions.ts
ncordon: I've just checked with an instance of the database to be sure. Please check always with the database itself because you've seen ChatGPT already give hallucinations regarding neo4j documentation for procedures.

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -32,7 +32,7 @@ import { CompletionItem, Neo4jFunction, Neo4jProcedure } from '../types';
 const uniq = <T>(arr: T[]) => Array.from(new Set(arr));
 
 function backtickIfNeeded(e: string): string | undefined {
-  if (/\s/.test(e) || /[^a-zA-Z]/.test(e)) {
+  if (/[^\p{L}\p{N}_]/u.test(e) || /[^\p{L}_]/u.test(e[0])) {
```
ncordon, line 35: We should check first the string is not empty:

```typescript
if (/[^\p{L}\p{N}_]/u.test(e) || (e.length > 0 && /[^\p{L}_]/u.test(e[0]))
```

packages/language-support/src/autocompletion/completionCoreCompletions.ts
anderson4j: I do indeed know about hallucinations and have also checked, no need to worry

ncordon (approved):

PR #304, opened 2024-12-03
Updates grammar and semantic analysis to the LTS version

ncordon:


modify-semantic-analysis.js
```diff
@@ -20,38 +20,10 @@ fs.readFile(file, 'utf8', function (err, data) {
...
-    cnsa_Main_updateSignatureResolver(null);
-    cnsa_Main_analyzeQuery(var$1[0]);
-    cnsa_Main_updateSignatureResolver(null);
-    cnsa_Main_analyzeQuery(var$1[0]);`,
-    `    cnsa_Main_$callClinit();`,
-  );
-
-  result = result.replace(
-    `$rt_exports.main = $rt_export_main;`,
+    `export { $rt_export_main as main };`,
```
OskarDamkjaer, line 23: I'd expected the transpiled code not to need modification and the `export { $rt_export_main as main };`. I think it's outlined in the [docs](https://teavm.org/docs/runtime/js-modules.html) under `Generating a module` how to not need/use a main with the phony calls we have now:

>By default, you use Java convention with public static main method of the main class specified in configuration. A module will be generated that exports one method called main. If you need to convert a library, don't follow this convention. Instead, create static methods in your main class and annotate them with @JSExport annotation

OskarDamkjaer: If that'd let use get rid of this extra step, it'd help us get closer to automatic publishing + depending on semantic analysis

OskarDamkjaer: Perhaps then we could try

```
<!-- Whether TeaVM should produce minified JavaScript. Can reduce JavaScript file size more than
                                 two times -->
<minifying>false</minifying>
```

minifying again? 🤔 I don't remember if why we don't have it on

ncordon: Nice! https://github.com/neo-technology/neo4j/pull/28588

OskarDamkjaer (approved):
😌 😌 😌

PR #306, opened 2024-12-04
Extended pre parser options

anderson4j:
I just added one commit but git suggests a new PR. The old PR doesnt get the push it seems?


anderson4j:
branch names were somehow different (correct camelcase on github or vice versa), found a workaround

PR #307, opened 2024-12-04
Adds simple warnings on deprecated procedures and functions

anderson4j:
Writing queries like:
`MATCH (n) RETURN id(n)`
`CALL db.create.setVectorProperty...`
Will now get 'deprecated' tags on the deprecated functions/procedures used, with messages like
"Function id is deprecated."/"Procedure db.create.setVectoryProperty is deprecated.". 

In vscode having this tag also means getting strike-through on the function name in-query.


packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -138,6 +142,60 @@ function generateProcedureNotFoundError(
...
+): SyntaxDiagnostic {
+  const rawText = parsedProcedure.rawText;
+  const nameChunks = rawText.split('\n');
+  const linesOffset = nameChunks.length - 1;
+  const lineIndex = parsedProcedure.line - 1;
+  const startColumn = parsedProcedure.column;
+  const endColumn =
+    linesOffset == 0
+      ? startColumn + rawText.length
+      : nameChunks.at(-1)?.length ?? 0;
```
ncordon, line 156: This piece of code is repeated several times in the codebase already, so I'd suggest maybe splitting it to a helper method?

ncordon (changes requested):
We need to add the strikethrough to the codemirror editor as well.

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -62,6 +66,35 @@ function detectNonDeclaredLabel(
...
+): SyntaxDiagnostic {
+  const rawText = parsedProcOrFunc.rawText;
+  const nameChunks = rawText.split('\n');
+  const linesOffset = nameChunks.length - 1;
+  const lineIndex = parsedProcOrFunc.line - 1;
+  const startColumn = parsedProcOrFunc.column;
+  const endColumn =
+    linesOffset == 0
+      ? startColumn + rawText.length
+      : nameChunks.at(-1)?.length ?? 0;
```
ncordon, line 83: This part is still used in `detectNonDeclaredLabel`, so maybe worth collapsing that here as well

PR #309, opened 2024-12-06
Changes param in cmd-parser to take escaped strings too

anderson4j:
Cypher5/25Parser.g4 define
parameter[String paramType]
   : DOLLAR parameterName[paramType]
   ;

parameterName[String paramType]
   : (symbolicNameString | UNSIGNED_DECIMAL_INTEGER)
   ;
   
 In CypherCmdParser.g4 instead we had:
 paramsCmd: PARAM paramsArgs?;

paramsArgs: (CLEAR | listCompletionRule | map | lambda);

lambda: unescapedSymbolicNameString EQ GT expression;

So params created with lambda could never be escaped, even though regular parser allows it. this aligns cmd-parser with regular.

### DIFF
--- a/packages/language-support/src/tests/consoleCommands.test.ts
+++ b/packages/language-support/src/tests/consoleCommands.test.ts
@@ -317,6 +317,25 @@ describe('parameters', () => {
     ]);
   });
 
+  test('allows setting parameters with escaped name', () => {
+    expectParsedCommands(':param `foo foo` => bar', [
+      {
+        type: 'set-parameters',
+        parameters: [{ name: '`foo foo`', expression: 'bar' }],
+      },
+    ]);
+
+    expectParsedCommands(':param {`a a a`: 2, `b-b-b`: rand()}', [
+      {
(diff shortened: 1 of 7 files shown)


packages/language-support/src/tests/consoleCommands.test.ts
```diff
@@ -456,7 +475,7 @@ describe('parameters', () => {
         length: 1,
         position: { line: 0, startCharacter: 7, startOffset: 7 },
         token: 'x',
-        tokenType: 'variable',
```
ncordon, line 459: This is not correct, it should still be a variable

anderson4j: After shifting the definition of the lambda expression in the parser to use parameterName, this shifted to be symbolicName. I figured that was just an effect of shifting to parsing as a different token, but maybe not? Having a bit of trouble finding where it's assigned..

ncordon (approved):

PR #310, opened 2024-12-09
Adds backticks to dbNames/aliases if needed

anderson4j:
Before autocompletions for database names/aliases would be like:
Ex. database name = "matrix-db"
`SHOW DATABASE ^` -> `SHOW DATABASE matrix-db` which can not be parsed

now we get
                                  -> `SHOW DATABASE ``matrix-db```
                                  
Note: we might allow parameters to have escaped names too (`my-param`) in which case we should update their usages for db-name-suggestions too

Further possibility for stuff that might need backticks - constraints/indexes once we suggest for these.


packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -41,6 +41,16 @@ function backtickIfNeeded(e: string): string | undefined {
   }
 }
 
+function backtickDbNameIfNeeded(e: string): string | undefined {
```
anderson4j, line 44: We could skip this separate method and expand "backtickIfNeeded" with an option-parameter that decides which symbols we should allow. Let me know if you have a preference

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -767,7 +779,10 @@ function completeSymbolicName({
     // so target should be non-existent
     (ruleList.includes(CypherParser.RULE_renameUser) && afterToToken)
   ) {
-    return parameterSuggestions;
+    return parameterSuggestions.map((completionItem) => ({
+      label: completionItem.label,
+      kind: completionItem.kind,
+    }));
```
ncordon, line 785: Why do we need these mappings to label and kind? This is not doing anything I think?

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -786,7 +801,10 @@ function completeSymbolicName({
       })),
     ];
 
-    return result;
+    return result.map((completionItem) => ({
+      label: completionItem.label,
+      kind: completionItem.kind,
+    }));
```
ncordon, line 807: Same here

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -802,7 +820,10 @@ function completeSymbolicName({
         label: roleName,
         kind: CompletionItemKind.Value,
       })),
-    ];
+    ].map((completionItem) => ({
+      label: completionItem.label,
+      kind: completionItem.kind,
+    }));
```
ncordon, line 826: Same here

packages/language-support/src/tests/autocompletion/databasenameCompletion.test.ts
```diff
@@ -15,6 +27,32 @@ describe('Can complete database names', () => {
     },
   };
 
+  test('Correctly completes database names/aliases with special symbols using backticks', () => {
+    const query = 'SHOW DATABASE ';
```
ncordon, line 31: This test is unnecessary, you could just add the completions to the existing test?

packages/language-support/src/tests/autocompletion/databasenameCompletion.test.ts
```diff
@@ -15,6 +27,32 @@ describe('Can complete database names', () => {
...
+        {
+          label: 'office-db',
+          kind: CompletionItemKind.Value,
+          insertText: '`office-db`',
+        },
+      ],
+    });
+  });
+
   test('Correctly completes database names and aliases in SHOW DATABASE', () => {
```
ncordon, line 56: There's also a `SHOW ALIAS ` test in this file which you should add the corresponding checks to

packages/language-support/src/autocompletion/completionCoreCompletions.ts
anderson4j: Used to backtick if needed parameters too (so had insertText: backtickIfNeeded(...label) too), but they could not be added with backticks yet. Looks like I forgot to remove the whole thing. 

Either way, since we're changing the parser-file for creating params, maybe we should just add backticks if needed here already, so once it's valid, we're good to go?

packages/language-support/src/tests/autocompletion/databasenameCompletion.test.ts
anderson4j: I split it on purpose so you would see that only this test failed if something purely related to backticks broke. You don't think there's any value to this?

packages/language-support/src/autocompletion/completionCoreCompletions.ts
anderson4j: With a parameter to check what regex to use then, the difference between the methods is that this method allows for "." in the text without adding backticks

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -702,7 +716,11 @@ function completeAliasName({
     candidateRule.ruleList.includes(CypherParser.RULE_createAlias) &&
     candidateRule.ruleList.includes(CypherParser.RULE_aliasName)
   ) {
-    return parameterSuggestions;
+    return parameterSuggestions.map((completionItem) => ({
+      label: completionItem.label,
+      kind: completionItem.kind,
+      insertText: backtickDbNameIfNeeded(completionItem.label),
```
ncordon, line 722: Can a parameter include the `.`, otherwise this should be `backtickIfNeeded`, no?

anderson4j: Ah yeah, this is not valid

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -51,6 +51,22 @@ function backtickDbNameIfNeeded(e: string): string | undefined {
   }
 }
 
+function backtickParamIfNeeded(e: string): string | undefined {
+  if (e == null || e == '') {
+    return undefined;
+  }
+  const namePart = e.slice(1);
+  if (e[0] == '$') {
```
anderson4j, line 59: Should I even have this test, or just trust the backtickParam... -method to only be called with params?

ncordon: I think it's okay to have double guards in case something goes wrong in the other part of the code.

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -41,6 +41,32 @@ function backtickIfNeeded(e: string): string | undefined {
...
+    if (/[^\p{L}\p{N}_]/u.test(namePart) || /[^\p{L}_]/u.test(namePart[0])) {
+      return '$' + `\`${namePart}\``;
+    } else {
+      return undefined;
+    }
+  } else {
+    return backtickIfNeeded(e);
+  }
+}
```
ncordon, line 68: You could refactor all this method to depend on the `backtickIfNeeded` as:

```ts
function backtickParamIfNeeded(e: string): string | undefined {
  if (e == null || e == '') {
    return undefined;
  }
  const namePart = e.slice(1);
  const backticked = backtickIfNeeded(namePart);
  if (backticked) {
    return '$' + backticked;
  } else {
    return undefined;
  }
}

File: packages/language-support/src/autocompletion/completionCoreCompletions.ts [hunk 10]
```diff
@@ -702,7 +732,11 @@ function completeAliasName({
     candidateRule.ruleList.includes(CypherParser.RULE_createAlias) &&
     candidateRule.ruleList.includes(CypherParser.RULE_aliasName)
   ) {
-    return parameterSuggestions;
+    return parameterSuggestions.map((completionItem) => ({
+      label: completionItem.label,
+      kind: completionItem.kind,
+      insertText: backtickParamIfNeeded(completionItem.label),
+    }));
```
ncordon, line 739: We should move the insertText thing to where the parameterSuggestions are defined instead of repeating this in the code a few times

packages/language-support/src/autocompletion/completionCoreCompletions.ts
anderson4j: Looks like fixing comment 2 will actually mean I can get ahead of adding the $ and just use backtickIfNeeded from scratch

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -312,10 +296,16 @@ const parameterCompletions = (
...
-    .map(([paramName]) => ({
-      label: `$${paramName}`,
-      kind: CompletionItemKind.Variable,
-    }));
+    .map(([paramName]) => {
+      const backtickedName = backtickIfNeeded(paramName);
+      return {
+        label: `$${paramName}`,
+        kind: CompletionItemKind.Variable,
+        ...(backtickedName != undefined
```
ncordon, line 304: This comparison needs to be `!==` I think.

You can just do `backtickedName ? { insertText: `$${backtickIfNeeded(paramName)}` } : {}`

ncordon (approved):
Address my last comment please and ready to go 😄

PR #313, opened 2024-12-18
Disables flaky debouncing test

ncordon:
I've carded this as one of the paper cuts we should address in Q1.

## Why

After a serious of PRs trying to address debouncing problems, the functionality seems to still be flaky somehow

https://github.com/neo4j/cypher-language-support/pulls?q=is%3Apr+author%3Adanieladugyan+is%3Aclosed

### DIFF
--- a/packages/react-codemirror/src/e2e_tests/debounce.spec.tsx
+++ b/packages/react-codemirror/src/e2e_tests/debounce.spec.tsx
@@ -28,43 +28,47 @@ test.fail(
 );
 
 // TODO Fix this test
-test.skip('onExecute updates should override debounce updates', async ({
-  mount,
-  page,
-}) => {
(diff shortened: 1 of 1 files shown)


packages/react-codemirror/src/e2e_tests/debounce.spec.tsx
```diff
@@ -27,7 +27,8 @@ test.fail(
   },
 );
 
-test('onExecute updates should override debounce updates', async ({
+// TODO Fix this test
+test.skip('onExecute updates should override debounce updates', async ({
```
daveajrussell, line 31: I think we can use `test.fixme` which I reckon signals the intent to fix better than `test.skip`

daveajrussell (approved):

PR #315, opened 2024-12-19
Fixes base when serving vite files

ncordon:
Taken from [here](https://stackoverflow.com/questions/59925804/blocked-because-of-a-disallowed-mime-type-text-html-angular-8-deployed-on)

## Why
Because I was getting a nasty error opening the playground from Firefox, which was not happening before 🤷


anderson4j (approved):
good stuff!

PR #317, opened 2024-12-23
Moves syntax errors to the semantic analysis worker

ncordon:
## What
Up until now we were computing all of the syntax errors as a listener on parsing, but we want to move to a pattern of supporting both syntax and semantic errors in a separate thread (web worker).

## Why
Because we want to be able to support both Cypher 5 and Cypher 25, plus minor versions for Cypher 25 and this gives us flexibility to just reload the web worker for the appropriate version, while auto-completing with the latest Cypher 25 grammar.

The semantic analysis packages both Cypher 5 and Cypher 25 knowledge. This would mean some auto-completions could potentially be too modern, but at least we would mark the errors appropriately to the user (right now for latest Cypher 5 or LTS and latest Cypher 25, in the future even for minor Cypher 25 versions if we need to).


packages/language-support/src/syntaxValidation/completionCoreErrors.ts
```diff
@@ -127,14 +127,6 @@ export function completionCoreErrormessage(
...
 
   if (options.length === 0) {
-    // options length is 0 should only happen when RULE_consoleCommand is hit and there are no other options
-    if (
-      ruleCandidates.find(
-        (ruleNumber) => ruleNumber === CypherParser.RULE_consoleCommand,
-      )
-    ) {
-      return 'Console commands are unsupported in this environment.';
-    }
```
ncordon, line 137: We don't need this anymore

packages/react-codemirror/src/lang-cypher/lintWorker.ts
```diff
@@ -1,14 +1,28 @@
...
-workerpool.worker({ validateSemantics });
+function lintCypherQuery(
+  query: string,
+  dbSchema: DbSchema,
+  featureFlags: { consoleCommands?: boolean } = {},
+) {
+  // We allow to override the consoleCommands feature flag
+  if (featureFlags.consoleCommands !== undefined) {
+    _internalFeatureFlags.consoleCommands = featureFlags.consoleCommands;
+  }
```
ncordon, line 16: Before the internal feature flags (which are a global variable) only affected things that were run in the same thread than the editor (namely the syntax errors). Now we are moving the syntax errors down to the semantic analysis worker, so this change requires us to set that global variable there as well.

There's been talks for reshaping the way we package the language support into a class, but until then, this is the way forward.

packages/language-support/src/tests/syntaxValidation/syntacticValidation.test.ts
```diff
@@ -675,6 +683,41 @@ describe('Syntactic validation spec', () => {
     ]);
   });
 
+  test.fails(
+    'Syntax validation errors on multiline unfinished property keys',
+    () => {
+      const query = `RETURN {\`something
+    foo
+    
+    bar: "hello"}`;
```
ncordon, line 692: The error for this case has worsened and it would require some work in the mono-repo to fix it

packages/language-support/src/tests/syntaxValidation/syntacticValidation.test.ts
```diff
@@ -732,28 +776,12 @@ describe('Syntactic validation spec', () => {
...
-        },
-        severity: 1,
-      },
     ]);
   });
 
-  test('Syntax validation errors on an expected procedure name', () => {
+  // TODO FIX ME
+  // Problem here is we were getting a better error before
+  test.fails('Syntax validation errors on an expected procedure name', () => {
```
ncordon, line 784: The error for this case has also worsened, so this would have to be addressed from the mono repo

packages/language-support/src/tests/syntaxValidation/syntacticValidation.test.ts
```diff
@@ -820,14 +848,14 @@ describe('Syntactic validation spec', () => {
...
+        message: `Invalid input '"foo"': expected a graph pattern, a parameter, ')', ':', 'IS', 'WHERE' or '{'`,
         offsets: {
-          end: 14,
+          end: 15,
           start: 9,
         },
         range: {
           end: {
-            character: 14,
+            character: 15,
```
ncordon, line 858: Because of the way we are computing the positions finish in the mono repo, some errors are spanning new characters now.

That is, when we receive a syntax error, we just know where the token starts, but not where it finishes. And the only approach we have in that case is to consider the token finishes at the next non space position.

packages/react-codemirror/src/e2e_tests/e2eUtils.ts
```diff
@@ -60,7 +60,7 @@ export class CypherEditorPage {
   }
 
   async checkNoNotificationMessage(type: 'error' | 'warning') {
-    await this.page.waitForTimeout(1000);
+    await this.page.waitForTimeout(3000);
```
ncordon, line 63: This was giving flakiness

packages/react-codemirror/src/e2e_tests/performanceTest.spec.tsx
```diff
@@ -100,7 +100,7 @@ test('benchmarking & performance test session', async ({
 
   await editorPage.checkErrorMessage(
     'RETRN',
-    'Unexpected token. Did you mean RETURN?',
+    `Invalid input 'RETRN': expected a graph pattern, 'FOREACH', ',', 'ORDER BY', 'CALL', 'CREATE', 'LOAD CSV', 'DELETE', 'DETACH', 'FINISH', 'INSERT', 'LIMIT', 'MATCH', 'MERGE', 'NODETACH', 'OFFSET', 'OPTIONAL', 'REMOVE', 'RETURN', 'SET', 'SKIP', 'UNION', 'UNWIND', 'USE', 'USING', 'WHERE', 'WITH' or <EOF>`,
```
ncordon, line 103: Some errors like this one have significantly worsened, I kind of liked the old one better but it's not easy to fix this

packages/react-codemirror/src/e2e_tests/syntaxValidation.spec.tsx
```diff
@@ -182,33 +182,34 @@ test('Validation errors are correctly overlapped', async ({ page, mount }) => {
...
-  );
-  await editorPage.checkWarningMessage(
-    'id',
-    "Function id is deprecated.",
-  );
-
+  await expect(
+    editorPage.page.locator('.cm-deprecated-element').last(),
+  ).toBeVisible({ timeout: 3000 });
+  await editorPage.checkWarningMessage('id', 'Function id is deprecated.');
```
ncordon, line 196: Formatting changes 🤷

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -185,80 +185,6 @@ function warnOnUndeclaredLabels(
...
-function fixSemanticAnalysisPositions({
-  semanticElements,
-  parseResult,
-}: FixSemanticPositionsArgs): SyntaxDiagnostic[] {
-  const cmd = parseResult.command;
-  return semanticElements.map((e) => {
-    let token: Token | undefined = undefined;
-
-    const start = Position.create(
-      e.position.line - 1 + cmd.start.line - 1,
```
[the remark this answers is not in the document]
anderson4j: Tried switching to ...line - 2 here too, and got line = -1 in some tests, so I guess intended. Still a bit confused why we needed -1x2 before and -1 now

packages/language-support/src/tests/syntaxValidation/syntacticValidation.test.ts
```diff
@@ -111,15 +114,15 @@ describe('Syntactic validation spec', () => {
 
     expect(getDiagnosticsForQuery({ query })).toEqual([
       {
+        message:
+          "Invalid input 'n': expected an expression, 'FOREACH', 'ORDER BY', 'CALL', 'CREATE', 'LOAD CSV', 'DELETE', 'DETACH', 'FINISH', 'INSERT', 'LIMIT', 'MATCH', 'MERGE', 'NODETACH', 'OFFSET', 'OPTIONAL', 'REMOVE', 'RETURN', 'SET', 'SKIP', 'UNION', 'UNWIND', 'USE', 'WITH' or '}'",
         offsets: {
-          end: 114,
```
anderson4j, line 115: Think we mentioned these changes in offsets yesterday, but I can't quite remember - why do they happen again?

anderson4j (approved):
Had some spots where I got confused but didnt find any errors. Here might be a good place to confirm I got things correctly. We now get syntax errors from the transpiled semanticAnalysis.js right? (So change is - remove our own syntax error creation + hook in syntax errors from semanticAnalysis)

packages/language-support/src/syntaxValidation/syntaxValidation.ts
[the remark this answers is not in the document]
ncordon: Because lines and columns start at 1 for antlr (and I think for the semantic errors in the database), so we were returning those. I adjusted the ones in the transpiled semantic anallysis so we don't have to worry about those from our side.

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -185,80 +185,6 @@ function warnOnUndeclaredLabels(
...
-      e.position.column - 1 + (e.position.line === 1 ? cmd.start.column : 0),
-    );
-
-    const startOffset = e.position.offset + cmd.start.start;
-    const toExplore: ParseTree[] = [parseResult.ctx];
-
-    while (toExplore.length > 0) {
```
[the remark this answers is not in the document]
ncordon: Correct, I've moved that down to the semantic analysis, because we have a proper parsing there (cypher 5 and cypher 25), whereas in the Javascript side we'll only maintain the Cypher 25 parser

packages/language-support/src/tests/syntaxValidation/syntacticValidation.test.ts
ncordon: I've managed to fix a few of these. They were happening because syntax errors in the database just carry the start position.

Adding the end position for all of those is a really big task, so I've managed to rescue the tokens we lexed and finding the token that corresponds to this position. It works better for most of the cases I'd say.

packages/language-support/src/tests/syntaxValidation/syntacticValidation.test.ts
```diff
@@ -132,31 +135,6 @@ describe('Syntactic validation spec', () => {
     ]);
   });
 
-  test('Misspelt keyword in the middle of the statement', () => {
-    const query = "MATCH (n:Person) WERE n.name = 'foo'";
```
ncordon, line 136: Ah yeah, this test was repeated

PR #318, opened 2024-12-23
Relints documents when connecting / disconnecting to a database

ncordon:
The failing behaviour can be seen in the test added by 706f9ec.

## Why

Because if we ask to lint a document when the schema is yet not fetched, we will never see schema dependent errors (non existent labels, deprecation warnings, non existent procedures, etc). This can happen when we open the  the editor open with a cypher file when we haven't yet connected to a database.

Conversely, when we disconnect from a database, this should also relint the document.

This will become more relevant when we have several connections and we can be connected to different databases with different procedures / functions, different labels, ...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19600,10 +19600,10 @@
     },
     "packages/language-server": {
       "name": "@neo4j-cypher/language-server",
-      "version": "2.0.0-next.12",
+      "version": "2.0.0-next.13",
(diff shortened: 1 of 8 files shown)


daveajrussell (approved):
This makes sense to me 👍 

I can see the relint function running (only once) when the schema fetched event fires

PR #320, opened 2024-12-31
Fixes bug reporting missing label / rel type when inside opposite pattern

ncordon:
## What

We are reporting a missing label for a relationship type when it's a node predicate. This pr fixes that bug.

## Why

Because we were traversing the tree to check whether we were inside a node pattern first, without accounting for the fact we could be in a relationship type inside a node predicate.


anderson4j (approved):
LGTM

PR #322, opened 2025-01-08
Added hint when using function as procedure and vice versa

anderson4j:
Will yield errors like:
`CALL abs(-50) YIELD *` -> Procedure abs is not present in the database. Did you mean to call the function abs? Only procedures can be called inside a CALL clause.
and
`RETURN apoc.create.uuids(50)` -> Function apoc.create.uuids is not present in the database. Did you mean to call the procedure apoc.create.uuids? Procedures must be called inside a CALL clause.

Instead of the regular "Procedure/Function is not present in the database"-messages.

### DIFF
--- a/packages/language-support/src/syntaxValidation/syntaxValidation.ts
+++ b/packages/language-support/src/syntaxValidation/syntaxValidation.ts
@@ -89,10 +89,8 @@ function detectNonDeclaredFunction(
 ): SyntaxDiagnostic | undefined {
   const exists = functionExists(parsedFunction, functionsSchema);
   if (!exists) {
-    let existsAsProcedure = false;
-    if (procedureSchema) {
-      existsAsProcedure = Boolean(procedureSchema[parsedFunction.name]);
-    }
+    const existsAsProcedure =
+      procedureSchema && Boolean(procedureSchema[parsedFunction.name]);
     if (existsAsProcedure) {
       return generateProcedureUsedAsFunctionError(parsedFunction);
     }
@@ -101,7 +99,7 @@ function detectNonDeclaredFunction(
 }
 
 function functionExists(
-  functionCandidate: ParsedFunction  ,
+  functionCandidate: ParsedFunction,
   functionsSchema: Record<string, Neo4jFunction>,
 ): boolean {
   if (!functionCandidate || !functionsSchema) {
@@ -115,15 +113,11 @@ function functionExists(
     functionsSchema[lowercaseFunctionName];
 
   // Built-in functions are case-insensitive in the database
-  if (
+  return (
     functionExistsWithExactName ||
     (caseInsensitiveFunctionInDatabase &&
       caseInsensitiveFunctionInDatabase.isBuiltIn)
-  ) {
-    return true;
-  } else {
-    return false;
-  }
+  );
 }
 
 function generateFunctionUsedAsProcedureError(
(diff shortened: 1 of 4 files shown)


packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -85,25 +85,56 @@ function generateSyntaxDiagnostic(
...
-  const functionExistsWithExactName = Boolean(
-    functionsSchema[parsedFunction.name],
+function generateFunctionUsedAsProcedureError(
+  parsedProcedure: ParsedProcedure,
+): SyntaxDiagnostic {
+  return generateSyntaxDiagnostic(
+    parsedProcedure.rawText,
+    parsedProcedure,
+    DiagnosticSeverity.Error,
+    `Procedure ${parsedProcedure.name} is not present in the database. Did you mean to call the function ${parsedProcedure.name}? Only procedures can be called inside a CALL clause.`,
```
ncordon, line 136: It would be better if the error was actionable IMHO:

```
${parsedProcedure.name} is a function, not a procedure. Did you mean to use the function ${parsedProcedure.name} with a RETURN clause instead of a CALL clause?
```

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -85,25 +85,56 @@ function generateSyntaxDiagnostic(
...
   functionsSchema: Record<string, Neo4jFunction>,
+  procedureSchema: Record<string, Neo4jProcedure>,
 ): SyntaxDiagnostic | undefined {
-  const lowercaseFunctionName = parsedFunction.name.toLowerCase();
+  const exists = functionExists(parsedFunction, functionsSchema);
+  if (!exists) {
+    let existsAsProcedure = false;
+    if (procedureSchema) {
+      existsAsProcedure = Boolean(procedureSchema[parsedFunction.name]);
+    }
```
ncordon, line 95: This can be rewritten as:
```
const existsAsProcedure = procedureSchema && Boolean(procedureSchema[parsedFunction.name]);
```

packages/language-support/src/tests/syntaxValidation/functionsValidation.test.ts
```diff
@@ -74,6 +74,42 @@ describe('Functions semantic validation spec', () => {
     ]);
   });
 
+  test('Syntax validation error on function used as procedure returns helpful message', () => {
+    const query = `CALL abs(-50) YIELD *`;
```
ncordon, line 78: It'd be good to add a test to check we are matching case insensitive functions:

```
CALL ABS(-1)
```

packages/language-support/src/tests/syntaxValidation/proceduresValidation.test.ts
```diff
@@ -59,6 +59,42 @@ meaning that it expects at least 3 arguments of types NODE, STRING, ANY
     ]);
   });
 
+  test('Syntax validation error on procedure used as function returns helpful message', () => {
+    const query = `RETURN apoc.create.uuids(50)`;
```
ncordon, line 63: Conversely, we should add a test to check

```
RETURN APOC.create.uuids(50)
```

would not give any error (because procedures are case sensitive).

anderson4j: Implemented this now, but then got to thinking.. probably any time someone writes strings that match procedure names except for the wrong case, they actually mean the procedure name, so maybe the better solution is to suggest the procedure name with correct case? (so case-insensitive check + case-correct suggest)

daveajrussell (approved):
Tested out in the codemirror playground - working for me nicely 👍

PR #323, opened 2025-01-08
1.7.0 release

anderson4j:


packages/vscode-extension/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # neo4j-for-vscode
 
+## 1.7.0
+
+- Adds "deprecated by" to deprecation warnings when available
+- Adds support for parameters to have escaped names
+- Adds autocompletions for Cypher version
+- Firefox path bugfix
```
ncordon, line 8: The Firefox path bugfix is not relevant to the VSCode realese

ncordon (changes requested):
We are missing more updates on the list:

- Adds support for CYPHER preparser options (runtime, planner, etc)
- Relints documents when connecting / disconnecting to a database
- Completes database and alias names with backticks if needed

anderson4j:
The previous update mentions the disconnect stuff and also "improving" backticking, and since the release was after the dbName/alias backtick-commit I figure that's what it's referring to

ncordon:
Yep you are right, I was looking at 1.6.0 not 1.6.1

ncordon (approved):

PR #325, opened 2025-01-09
Updates semantic analysis and grammar for version 2025.01.0

ncordon:
There are no changes to the grammar files


anderson4j (approved):
not much to say, good stuff

PR #327, opened 2025-01-10
Added automatic trigger of completions on "YIELD " 

anderson4j:
Previously you would need to trigger autocompletions with "control + space". This triggers the completion automatically after you type  "YIELD ".

Does not trigger automatically after one return value has been written (ex. "YIELD name, " needs manual triggering)


anderson4j:
Maybe not fixed in windows afterall... re-did a build and it failed again :/

packages/language-support/src/autocompletion/autocompletionHelpers.ts
```diff
@@ -0,0 +1,11 @@
+export function shouldAutoCompleteYield(query: string, offset: number) {
+  const yieldTriggerPhrase = 'yield ';
+  const text = query.slice(0, offset);
+  const yieldStart = offset - yieldTriggerPhrase.length;
+  if (yieldStart >= 0) {
+    const precedingText = text.slice(yieldStart, offset);
+    return precedingText.toLowerCase() === yieldTriggerPhrase;
+  } else {
```
daveajrussell, line 8: nitpick, redundant else.. could just return false, since we return inside the if

daveajrussell (approved):
Changes look good, and I can see YIELD autocompleting

packages/language-support/src/tests/autocompletion/autocompletionHelper.test.ts
```diff
@@ -0,0 +1,20 @@
+import { shouldAutoCompleteYield } from '../../autocompletion/autocompletionHelpers';
+
+test('shouldAutoCompleteYield does not produce false negatives', () => {
+  const query = 'MATCH dbms.components() YIELD ';
```
ncordon, line 4: Not a biggie but this should be a `CALL`

packages/language-support/src/tests/autocompletion/autocompletionHelper.test.ts
```diff
@@ -0,0 +1,20 @@
...
+  expect(shouldAutoCompleteYield(query, 30)).toBe(true);
+});
+
+test('shouldAutoCompleteYield is not case sensitive', () => {
+  const query = 'MATCH dbms.components() yIeLd ';
```
ncordon, line 9: Same

PR #328, opened 2025-01-13
Adds support for cypher versions in semantic analysis-linting

anderson4j:
Now if defaultLanguage-field is defined on a database, this language version will be used by default in the analysis, otherwise cypher 5 is used.
If the version is defined in-query, like `CYPHER 25 ...` this version will overwrite the default version, if it is given in an array of supported versions we store ourselves (currently 5 and 25).

### DIFF
--- a/packages/language-support/src/parserWrapper.ts
+++ b/packages/language-support/src/parserWrapper.ts
@@ -32,6 +32,7 @@ import {
 } from './helpers';
 import { SyntaxDiagnostic } from './syntaxValidation/syntaxValidation';
 import { SyntaxErrorsListener } from './syntaxValidation/syntaxValidationHelpers';
+import { CypherVersion } from './types';
 
 export interface ParsedStatement {
   command: ParsedCommand;
@@ -46,7 +47,7 @@ export interface ParsedStatement {
   collectedVariables: string[];
   collectedFunctions: ParsedFunction[];
   collectedProcedures: ParsedProcedure[];
-  cypherVersion?: string;
+  cypherVersion?: CypherVersion;
 }
 
 export interface ParsingResult {
@@ -176,13 +177,13 @@ export function createParsingResult(query: string): ParsingResult {
       const labelsCollector = new LabelAndRelTypesCollector();
       const variableFinder = new VariableCollector();
       const methodsFinder = new MethodsCollector(tokens);
-      const preparserFinder = new PreparserCollector();
+      const cypherVersionCollector = new CypherVersionCollector();
       const errorListener = new SyntaxErrorsListener(tokens);
       parser._parseListeners = [
         labelsCollector,
         variableFinder,
         methodsFinder,
-        preparserFinder,
+        cypherVersionCollector,
       ];
       parser.addErrorListener(errorListener);
       const ctx = parser.statementsOrCommands();
@@ -212,7 +213,7 @@ export function createParsingResult(query: string): ParsingResult {
         collectedVariables: variableFinder.variables,
         collectedFunctions: methodsFinder.functions,
         collectedProcedures: methodsFinder.procedures,
-        cypherVersion: preparserFinder.cypherVersion,
+        cypherVersion: cypherVersionCollector.cypherVersion,
       };
     });
 
@@ -412,8 +413,8 @@ class MethodsCollector extends ParseTreeListener {
   }
 }
 
-class PreparserCollector extends ParseTreeListener {
-  public cypherVersion: string;
+class CypherVersionCollector extends ParseTreeListener {
+  public cypherVersion: CypherVersion;
 
   constructor() {
     super();
@@ -431,7 +432,7 @@ class PreparserCollector extends ParseTreeListener {
 
   exitEveryRule(ctx: unknown) {
     if (ctx instanceof CypherVersionContext) {
-      this.cypherVersion = ctx.getText();
+      this.cypherVersion = `CYPHER ${ctx.getText()}` as CypherVersion;
     }
   }
 }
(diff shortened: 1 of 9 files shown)


packages/language-support/src/tests/syntaxValidation/semanticValidation.test.ts
```diff
@@ -3,6 +3,53 @@ import { testData } from '../testData';
...
+
+  test('Semantic analysis defaults to cypher 5 when no default version is given, and no version is given in query', () => {
+    const query1 = 'CYPHER  5 MATCH (n)-[r]->(m) SET r += m';
+    const diagnostics1 = getDiagnosticsForQuery({ query: query1 });
+    const query2 = 'MATCH (n)-[r]->(m) SET r += m';
+    const diagnostics2 = getDiagnosticsForQuery({ query: query2 });
+    expect(diagnostics1[0].message).toEqual(diagnostics2[0].message);
+  });
+
+  //TODO: Maybe this should actually yield a warning
```
anderson4j, line 22: In this card https://trello.com/c/0sFWd1iZ, it's mentioned that syntax errors arent working in general now that we've switched to semantic-worker errors, so maybe we want to handle this error there too?

Should be quite simple to check and throw as is though, so maybe I should just do this one now and handle the rest later?

packages/language-support/src/parserWrapper.ts
```diff
@@ -45,6 +46,7 @@ export interface ParsedStatement {
   collectedVariables: string[];
   collectedFunctions: ParsedFunction[];
   collectedProcedures: ParsedProcedure[];
+  cypherVersion?: string;
```
ncordon, line 49: Same here

packages/schema-poller/src/queries/databases.ts
```diff
@@ -30,6 +30,7 @@ export type Database = {
   writer?: boolean;
   access?: string;
   constituents?: string[];
+  defaultLanguage?: string; // to be introduced
```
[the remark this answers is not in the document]
ncordon: Again, best to probably use `'CYPHER 5' | 'CYPHER 25'` as the type here?

packages/language-support/src/syntaxValidation/semanticAnalysisWrapper.ts
```diff
@@ -51,6 +51,7 @@ function copySettingSeverity(
 export function wrappedSemanticAnalysis(
   query: string,
   dbSchema: DbSchema,
+  parsedVersion: string,
```
ncordon, line 54: This needs to be optional: `parsedVersion?: string`

ncordon (commented):
Good work, just left a few nits that can be improved

packages/language-support/src/parserWrapper.ts
```diff
@@ -174,8 +176,14 @@ export function createParsingResult(query: string): ParsingResult {
       const labelsCollector = new LabelAndRelTypesCollector();
       const variableFinder = new VariableCollector();
       const methodsFinder = new MethodsCollector(tokens);
+      const preparserFinder = new PreparserCollector();
```
ncordon, line 179: Every collector is in charge of extracting one type of structure here. I'd say say `cypherVersionCollector` here is more consistent with how we've named the rest

packages/language-support/src/tests/syntaxValidation/semanticValidation.test.ts
```diff
@@ -3,6 +3,53 @@ import { testData } from '../testData';
 import { getDiagnosticsForQuery } from './helpers';
 
 describe('Semantic validation spec', () => {
+  test('Semantic analysis is dependant on cypher version', () => {
+    const query1 = 'CYPHER  5 MATCH (n)-[r]->(m) SET r += m';
+    const diagnostics1 = getDiagnosticsForQuery({ query: query1 });
+    const query2 = 'CYPHER 25 MATCH (n)-[r]->(m) SET r += m';
+    const diagnostics2 = getDiagnosticsForQuery({ query: query2 });
+    expect(diagnostics1[0].message).not.toEqual(diagnostics2[0].message);
```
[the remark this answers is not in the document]
anderson4j: I liked this way since the focus of the test is not on what type of error is received, but rather that changing to "CYPHER 25" changes what errors we get in this example. I agree that they might break if the order of errors is sometimes changed, but I guess a `sort()` could fix that

Either way, no strong feelings so I can change it

packages/schema-poller/src/queries/databases.ts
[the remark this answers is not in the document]
anderson4j: Looks to work without :)

packages/language-support/src/tests/syntaxValidation/semanticValidation.test.ts
anderson4j: in follow-up

packages/language-support/src/syntaxValidation/semanticAnalysisWrapper.ts
```diff
@@ -63,7 +65,22 @@ export function wrappedSemanticAnalysis(
       });
     }
 
-    const semanticErrorsResult = analyzeQuery(query, 'cypher 5');
+    const validCypherVersions = ['CYPHER 5', 'CYPHER 25'];
+    let cypherVersion = 'CYPHER 5';
+    const defaultVersion = dbSchema.defaultLanguage?.toUpperCase();
+
+    if (parsedVersion && validCypherVersions.includes(parsedVersion)) {
```
ncordon, line 72: You don't need `validCypherVersions` anymore because `CypherVersion` is a type that can only be two possible values

ncordon (approved):

PR #329, opened 2025-01-14
Updates parser to Cypher 25

anderson4j:
Swaps out the current Cypher 5 lexer/parser for the Cypher 25 equivalents. If I remember correctly the motivation for just using the cypher 25 parser is that the changes should mostly be additive, and that we dont want to maintain 2 versions of things if we can get away with it.

A few tests broke, but not too much


packages/language-support/src/antlr-grammar/Cypher25Parser.g4
```diff
@@ -14,20 +14,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-parser grammar Cypher5Parser;
```
loveleif, line 17: FYI, the grammar files are uploaded as a zip file somewhere when we build the mono repo if that is helpful.

loveleif (approved):
Nothing to add from my point of view :)

anderson4j:
Looks like some e2e-tests might need to be changed, but since I dont really know how these textmate-tests work, I figure it's safer to wait for you @ncordon

ncordon (approved):
I updated the textmate files. I can explain when you come back 😄

PR #331, opened 2025-01-22
Adapts procedure registry to depend on the cypher version

ncordon:
Builds on top of #328.

* `packages/language-support/src/tests/testData.ts` just adds the new shape for the registry in the `mockSchema`
* `packages/language-support/src/syntaxValidation/semanticAnalysis.js` brings in the new changes to be able to update either the registry for Cypher 5 or Cypher 25.

## What
Polls for the procedure registry for both Cypher 5 and Cypher 25 and wires that into the code.

This means for older versions we do `SHOW PROCEDURES` and `SHOW FUNCTIONS` and store that under 'cypher 5' in the schema.

For newer versions we poll all of the following queries:

```cypher
CYPHER 5  SHOW PROCEDURES
CYPHER 25 SHOW PROCEDURES

CYPHER 5  SHOW FUNCTIONS
CYPHER 25 SHOW FUNCTIONS
```

## Why

Starting from neo4j `2025.01` we will support two concurrent cypher versions, which will mean we could have procedures that were deprecated in Cypher 5 and removed in Cypher 25, or even arguments that were deprecated in Cypher 5 and removed in Cypher 25.


packages/language-server/src/lintWorker.ts
```diff
@@ -1,6 +1,22 @@
...
 
+function lintCypherQuery(
+  query: string,
+  dbSchema: DbSchema,
+  featureFlags: { cypher25?: boolean } = {},
+) {
+  // We allow to override the consoleCommands feature flag
+  if (featureFlags.cypher25 !== undefined) {
+    _internalFeatureFlags.cypher25 = featureFlags.cypher25;
+  }
```
ncordon, line 16: We need to pass the feature flags to the lint worker because the `_internalFeatureFlags` variable from the outside lives in another thread

packages/language-server/src/server.ts
```diff
@@ -12,14 +12,21 @@ import {
...
 import { Neo4jSchemaPoller } from '@neo4j-cypher/schema-poller';
 import { doAutoCompletion } from './autocompletion';
 import { cleanupWorkers, lintDocument } from './linting';
 import { doSignatureHelp } from './signatureHelp';
 import { applySyntaxColouringForDocument } from './syntaxColouring';
 import { Neo4jSettings } from './types';
 
+if (process.env.CYPHER_25 === 'true') {
+  _internalFeatureFlags.cypher25 = true;
+}
```
ncordon, line 28: This is so we can set the feature flag in the language server from the vscode extension

packages/language-support/package.json
```diff
@@ -48,7 +48,7 @@
     "vscode-languageserver-types": "^3.17.3"
   },
   "scripts": {
-    "gen-parser": "antlr4 -Dlanguage=TypeScript -visitor src/antlr-grammar/CypherCmdLexer.g4 src/antlr-grammar/CypherCmdParser.g4 -o src/generated-parser/ -Xexact-output-dir",
+    "gen-parser": "export ANTLR4_TOOLS_ANTLR_VERSION=4.13.2 && antlr4 -Dlanguage=TypeScript -visitor src/antlr-grammar/CypherCmdLexer.g4 src/antlr-grammar/CypherCmdParser.g4 -o src/generated-parser/ -Xexact-output-dir",
```
ncordon, line 51: This is because https://github.com/antlr/antlr4-tools will look for the most up to date version of antlr otherwise. But it was failing contacting the relevant url, so this hardcodes the version we want to use (which we should anyway because we are using 4.13.2 and we wouldn't want to use a more recent one for generating the code but an older one as runtime).

packages/language-support/src/dbSchema.ts
```diff
@@ -9,6 +9,7 @@ export interface DbSchema {
   roleNames?: string[];
   parameters?: Record<string, unknown>;
   propertyKeys?: string[];
-  procedures?: Record<string, Neo4jProcedure>;
-  functions?: Record<string, Neo4jFunction>;
+  procedures?: Partial<Record<CypherVersion, Record<string, Neo4jProcedure>>>;
+  functions?: Partial<Record<CypherVersion, Record<string, Neo4jFunction>>>;
```
ncordon, line 13: We need a different procedure registry for each version because the same function / procedure could have a different signature in different versions.

At the moment there are not many examples of methods that have different signatures (just `apoc.cypher.runTimeboxed`), but there will be in the future.

packages/schema-poller/src/metadataPoller.ts
```diff
@@ -142,47 +161,65 @@ export class MetadataPoller {
...
-          }, {});
-          this.dbSchema.procedures = procedures;
-        }
-      },
+    const versions: (CypherVersion | undefined)[] = isNewerNeo4j
+      ? cypherVersions
+      : [undefined];
+
+    versions.forEach((cypherVersion) => {
+      const effectiveCypherVersion: CypherVersion = cypherVersion ?? 'cypher 5';
```
ncordon, line 169: If we are not connected to a new neo4j, just poll for `cypher 5`

packages/vscode-extension/src/extension.ts
```diff
@@ -30,6 +30,7 @@ export async function activate(context: ExtensionContext) {
     debug: {
       module: debugServer,
       transport: TransportKind.ipc,
+      options: { env: { CYPHER_25: 'true' } },
```
ncordon, line 33: This only affects the server version we use in the tests, not the production one. So we'll have `CYPHER_25` support enabled there by default.

packages/vscode-extension/tests/specs/api/syntaxValidation.spec.ts
```diff
@@ -15,29 +15,19 @@ import {
...
-        if (textFile) {
-          const docUri = getDocumentUri(textFile);
-          diagnostics = vscode.languages.getDiagnostics(docUri);
-        } else {
-          // We need to filter diagnostics for untitled files
-          // because there could be other opened files
-          diagnostics = vscode.languages
-            .getDiagnostics()
-            .filter(([uri]) => uri.scheme === 'untitled')
-            .flatMap(([, diagnostics]) => diagnostics);
```
ncordon, line 39: This was giving problems when having two documents opened that were temporary and not saved to file.

packages/schema-poller/src/queries/databases.ts
```diff
@@ -45,7 +45,11 @@ export function listDatabases(): ExecuteQueryArgs<{
 
   const resultTransformer = resultTransformers.mappedResultTransformer({
     map(record) {
-      return record.toObject() as Database;
+      const obj = record.toObject();
+      if (obj.defaultLanguage) {
+        obj.defaultLanguage = (obj.defaultLanguage as string).toLowerCase();
```
anderson4j, line 50: This part wasnt necessary in my semantic analysis PR (i set to uppercase there), and tests pass for me if I comment this out on this branch too.

I guess we could remove the toUpper in semanticAnalysisWrapper.wrappedSemanticAnalysis() and just do it here (upper instead of lower since I changed the casing of CypherVersion-values), but shouldnt have both I guess

anderson4j (changes requested):
Good stuff! Had some minor suggestions for improvement, and a question or two.

Also think some of the changes though good might have been a bit "off topic" (reformatting tests to use URI/better messages are what I think of at the top of my head). If im correct in thinking these could be their own PR, reviewing would be a little easier if we split these into their own thing

Also I made some comments about "cypher 5" vs "CYPHER 5" merging with my PR before you had done it - they are gone for me since you fixed it, but still in comment count, so ignore if they still pop up on corrected spots

packages/language-support/src/parserWrapper.ts
```diff
@@ -402,6 +412,35 @@ class MethodsCollector extends ParseTreeListener {
   }
 }
 
+class PreparserCollector extends ParseTreeListener {
```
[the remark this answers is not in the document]
ncordon: This does not apply anymore

packages/language-support/src/signatureHelp.ts
```diff
@@ -200,10 +200,20 @@ export function signatureHelp(
...
 
       if (method !== undefined) {
+        const cypherVersion = resolveCypherVersion(
+          statement.cypherVersion,
+          dbSchema,
+        );
         if (method.methodType === MethodType.function) {
-          result = toSignatureHelp(dbSchema.functions, method);
+          result = toSignatureHelp(
+            dbSchema?.functions?.[cypherVersion] ?? {},
```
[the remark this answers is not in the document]
ncordon: I think it should never be undefined no

packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -99,7 +102,7 @@ export interface CypherEditorProps {
    */
   featureFlags?: {
     consoleCommands?: boolean;
-    signatureInfoOnAutoCompletions?: boolean;
```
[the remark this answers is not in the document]
ncordon: This feature flag has been true for a long time, so it's not needed anymore

packages/schema-poller/src/metadataPoller.ts
```diff
@@ -149,47 +160,65 @@ export class MetadataPoller {
...
-            const { name } = curr;
-            acc[name] = curr;
-            return acc;
-          }, {});
-          this.dbSchema.procedures = procedures;
-        }
-      },
+    const versions: (CypherVersion | undefined)[] = isNewerNeo4j
+      ? cypherVersions
+      : [undefined];
```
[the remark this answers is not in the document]
ncordon: Because if I'm in an older version and I pass `CYPHER 5` down here we are going to use `CYPHER 5 SHOW FUNCTIONS / PROCEDURES`, when that version does not support the cypher annotation.

So there are three possible flavours to what we pass into the functions polling (similar for procedures):

`CYPHER 5`, `CYPHER 25` or `undefined` (in which case don't preprend, just assume the results go into `cypher 5`)

packages/language-support/src/tests/autocompletion/functionsCompletion.test.ts
```diff
@@ -3,10 +3,19 @@ import {
...
 import { testData } from '../testData';
 import { testCompletions } from './completionAssertionHelpers';
 
 describe('function invocations', () => {
+  beforeAll(() => {
+    _internalFeatureFlags.cypher25 = true;
+  });
+
+  afterAll(() => {
+    _internalFeatureFlags.cypher25 = false;
```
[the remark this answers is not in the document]
ncordon: I've changed it for this and for the tests using `_internalFeatureFlags.consoleCommands`

packages/language-support/package.json
```diff
@@ -48,7 +48,7 @@
     "vscode-languageserver-types": "^3.17.3"
   },
   "scripts": {
-    "gen-parser": "antlr4 -Dlanguage=TypeScript -visitor src/antlr-grammar/CypherCmdLexer.g4 src/antlr-grammar/CypherCmdParser.g4 -o src/generated-parser/ -Xexact-output-dir",
+    "gen-parser": "ANTLR4_TOOLS_ANTLR_VERSION=4.13.2 antlr4 -Dlanguage=TypeScript -visitor src/antlr-grammar/CypherCmdLexer.g4 src/antlr-grammar/CypherCmdParser.g4 -o src/generated-parser/ -Xexact-output-dir",
```
anderson4j, line 51: Im still having issues with this. shifting to "cross-env ANTLR4...." seems to work

packages/schema-poller/src/metadataPoller.ts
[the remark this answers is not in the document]
anderson4j: aha, we pass the cypherVersion value "undefined" to listProcedures for the prepending, while effectiveCypher... is for having a key in this.procedures, makes sense

anderson4j (approved):
Works now!

anderson4j (approved):
Good stuff! Performance improvement and the small test-fixes look good, but had some small questions

Also, build still works on Windows :)

packages/react-codemirror/src/e2e_tests/syntaxValidation.spec.tsx
```diff
@@ -227,14 +227,14 @@ test('Syntax validation depends on the Cypher version', async ({
 
   const editorPage = new CypherEditorPage(page);
   const textField = page.getByRole('textbox');
-  await textField.fill('CYPHER 5 CALL apoc.create.uuids(5)');
+  await textField.fill('CYPHER 5 CALL apoc.create.uuids(5)', { force: true });
```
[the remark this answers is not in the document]
ncordon: I think this is not needed

packages/react-codemirror/src/e2e_tests/snippets.spec.tsx
```diff
@@ -65,11 +65,11 @@ test('can accept completion inside pattern snippet', async ({
     page.locator('.cm-tooltip-autocomplete').getByText('City'),
   ).toBeVisible();
 
-  await textField.press('Tab');
+  await textField.press('Tab', { delay: 300 });
```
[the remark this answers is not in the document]
ncordon: Yeah exactly, it seemed flaky, so adding a delay to act more like a normal user seemed to fix it

PR #332, opened 2025-01-24
Fix bug using console commands as variables labels

ncordon:
The rule we override to allow the console commands in the parser to be used as variable names, labels, etc was not the correct one, so we couldn't use `clear`, `param`, etc in some places in the query:

```
antlr4-parse ./{CypherCmdLexer.g4,CypherCmdParser.g4} statementsOrCommands


CREATE (n:clear {use: 0})
WITH $param AS map
RETURN map.propertyKey

line 1:10 mismatched input 'clear' expecting {ESCAPED_SYMBOLIC_NAME, ACCESS, ACTIVE, ADMIN, ADMINISTRATOR, ALIAS, ALIASES, ALL_SHORTEST_PATHS, ALL, ALTER, AND, ANY, ARRAY, AS, ASC, ASCENDING, ASSIGN, AT, AUTH, BINDINGS, BOOL, BOOLEAN, BOOSTED, BOTH, BREAK, BUILT, BY, CALL, CASCADE, CASE, CHANGE, CIDR, COLLECT, COMMAND, COMMANDS, COMPOSITE, CONCURRENT, CONSTRAINT, CONSTRAINTS, CONTAINS, COPY, CONTINUE, COUNT, CREATE, CSV, CURRENT, DATA, DATABASE, DATABASES, DATE, DATETIME, DBMS, DEALLOCATE, DEFAULT, DEFINED, DELETE, DENY, DESC, DESCENDING, DESTROY, DETACH, DIFFERENT, '$', DISTINCT, DRIVER, DROP, DRYRUN, DUMP, DURATION, EACH, EDGE, ENABLE, ELEMENT, ELEMENTS, ELSE, ENCRYPTED, END, ENDS, EXECUTABLE, EXECUTE, EXIST, EXISTENCE, EXISTS, ERROR, FAIL, FALSE, FIELDTERMINATOR, FINISH, FLOAT, FOR, FOREACH, FROM, FULLTEXT, FUNCTION, FUNCTIONS, GRANT, GRAPH, GRAPHS, GROUP, GROUPS, HEADERS, HOME, ID, IF, IMPERSONATE, IMMUTABLE, IN, INDEX, INDEXES, INF, INFINITY, INSERT, INT, INTEGER, IS, JOIN, KEY, LABEL, LABELS, '!', LEADING, LIMITROWS, LIST, LOAD, LOCAL, LOOKUP, '(', MANAGEMENT, MAP, MATCH, MERGE, '%', NAME, NAMES, NAN, NFC, NFD, NFKC, NFKD, NEW, NODE, NODETACH, NODES, NONE, NORMALIZE, NORMALIZED, NOT, NOTHING, NOWAIT, NULL, OF, OFFSET, ON, ONLY, OPTIONAL, OPTIONS, OPTION, OR, ORDER, PASSWORD, PASSWORDS, PATH, PATHS, PLAINTEXT, POINT, POPULATED, PRIMARY, PRIMARIES, PRIVILEGE, PRIVILEGES, PROCEDURE, PROCEDURES, PROPERTIES, PROPERTY, PROVIDER, PROVIDERS, RANGE, READ, REALLOCATE, REDUCE, RENAME, REL, RELATIONSHIP, RELATIONSHIPS, REMOVE, REPEATABLE, REPLACE, REPORT, REQUIRE, REQUIRED, RESTRICT, RETURN, REVOKE, ROLE, ROLES, ROW, ROWS, SCAN, SEC, SECOND, SECONDARY, SECONDARIES, SECONDS, SEEK, SERVER, SERVERS, SET, SETTING, SETTINGS, SHORTEST_PATH, SHORTEST, SHOW, SIGNED, SINGLE, SKIPROWS, START, STARTS, STATUS, STOP, STRING, SUPPORTED, SUSPENDED, TARGET, TERMINATE, TEXT, THEN, TIME, TIMESTAMP, TIMEZONE, TO, TOPOLOGY, TRAILING, TRANSACTION, TRANSACTIONS, TRAVERSE, TRIM, TRUE, TYPE, TYPED, TYPES, UNION, UNIQUE, UNIQUENESS, UNWIND, URL, USE, USER, USERS, USING, VALUE, VARCHAR, VECTOR, VERTEX, WAIT, WHEN, WHERE, WITH, WITHOUT, WRITE, XOR, YIELD, ZONE, ZONED, IDENTIFIER}
```

### DIFF
--- a/.changeset/serious-ducks-sleep.md
+++ b/.changeset/serious-ducks-sleep.md
@@ -1,0 +1,5 @@
+---
+'@neo4j-cypher/language-support': patch
+---
+
+Fixes bug using console commands as variables or labels
--- a/packages/language-support/src/antlr-grammar/CypherCmdParser.g4
+++ b/packages/language-support/src/antlr-grammar/CypherCmdParser.g4
@@ -52,7 +52,7 @@ useCompletionRule: USE;
 serverCompletionRule: SERVER;
 
 // This rule overrides the identifiers adding EXPLAIN, PROFILE, etc
-unescapedLabelSymbolicNameString: 
+unescapedSymbolicNameString: 
     preparserKeyword 
     | HISTORY
     | CLEAR
--- a/packages/language-support/src/tests/syntaxColouring/preparser.test.ts
+++ b/packages/language-support/src/tests/syntaxColouring/preparser.test.ts
@@ -535,4 +535,244 @@ describe('Preparser syntax colouring', () => {
       },
     ]);
   });
+
+  test('Correctly colours console commands used as variables or labels', () => {
+    const query = `CREATE (n:clear {use: 0})
+WITH $param AS map
+RETURN map.propertyKey`;
+    expect(applySyntaxColouring(query)).toEqual([
+      {
+        bracketInfo: undefined,
+        length: 6,
+        position: {
+          line: 0,
+          startCharacter: 0,
+          startOffset: 0,
+        },
+        token: 'CREATE',
+        tokenType: 'keyword',
+      },
+      {
+        bracketInfo: {
+          bracketLevel: 0,
+          bracketType: 'parenthesis',
+        },
+        length: 1,
+        position: {
+          line: 0,
+          startCharacter: 7,
+          startOffset: 7,
+        },
+        token: '(',
+        tokenType: 'bracket',
+      },
+      {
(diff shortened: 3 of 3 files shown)


anderson4j (approved):
LGTM

PR #334, opened 2025-01-28
Adds standardization check to formatter tests

tomasnyberg:
### Description
When formatting, we want to ensure that we do not accidentally modify the parse tree in doing so. As a way of implementing this, I propose a "standardizer" which simply walks the tree and prints it as is, in order to make sure we don't e.g. drop some terminal nodes. This PR adds such a standardizer and runs it on every test. The idea is also to use this on a larger sample of queries once we get access to these (I've asked in #product_analytics). 

Since some styleguide rules change the tree, we can't simply do a comparison between the parsetree.toStringTree() strings. One styleguide rule says that ON CREATE should come before ON MATCH for instance, and so we need to take this into account in the standardizer. 

### Testing
- npm run test


ncordon (approved):
LGTM

PR #335, opened 2025-01-28
Add formatting keybind button

simonthuresson:
# Description

This pull request has the intent to connect the formatter with the Query Editor part of the repo. The Query Editor imports the formatter and then exposes the formatting functionality by a function but also allows the formatter to be ran by key-binding. The package react-codemirror-playground adds a button (see picture) which also runs the formatter. 

The functionality of the formatter is hidden behind a feature flag in react-codemirror-playground/src/App.tsx. Change the constant formatFlag to true in order to test. 

## Testing
Tested by running `npm run test` in root directory. All tests passed. 

Skärmavbild 2025-01-28 kl  16 37 12

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -46,8 +46,8 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   buffer: string[] = [];
   indentation = 0;
   indentationSpaces = 2;
-  targetToken = -1;
-  cursorPos = -1;
+  targetToken?: number;
+  cursorPos?: number;
 
   constructor(private tokenStream: CommonTokenStream) {
     super();
@@ -239,7 +239,6 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       this.cursorPos = this.buffer.join('').length;
     }
     this.buffer.push(result);
-
     this.addCommentsAfter(node);
   };
 
@@ -403,10 +402,8 @@ export function formatQuery(
   const tree = parser.statementsOrCommands();
   const visitor = new TreePrintVisitor(tokens);
 
-  if (cursorPosition === undefined) {
-    const result = visitor.format(tree);
-    return result;
-  }
+  if (cursorPosition === undefined) return visitor.format(tree);
+
   if (cursorPosition >= query.length || cursorPosition === 0) {
     const result = visitor.format(tree);
     return {
@@ -416,12 +413,11 @@ export function formatQuery(
   }
 
   const targetToken = findTargetToken(tokens.tokens, cursorPosition);
-
   const relativePosition = cursorPosition - targetToken.start;
   visitor.targetToken = targetToken.tokenIndex;
-  const result = visitor.format(tree);
+
   return {
-    formattedString: result,
+    formattedString: visitor.format(tree),
     newCursorPos: visitor.cursorPos + relativePosition,
   };
 }
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -364,7 +364,7 @@ interface FormattingResultWithCursor {
 
 export function formatQuery(query: string): string;
 export function formatQuery(
-  query: string,
+  quer: string,
   cursorPosition: number,
 ): FormattingResultWithCursor;
 export function formatQuery(
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -1,14 +1,6 @@
+import { CommonTokenStream, ParserRuleContext, TerminalNode } from 'antlr4';
+import { default as CypherCmdLexer } from '../generated-parser/CypherCmdLexer';
 import {
-  CharStreams,
-  CommonTokenStream,
-  ParserRuleContext,
-  TerminalNode,
-} from 'antlr4';
-import {
-  default as CypherCmdLexer,
-  default as CypherLexer,
-} from '../generated-parser/CypherCmdLexer';
-import CypherCmdParser, {
   ArrowLineContext,
   BooleanLiteralContext,
   ClauseContext,
@@ -31,6 +23,8 @@ import CypherCmdParser, {
 import CypherCmdParserVisitor from '../generated-parser/CypherCmdParserVisitor';
 import {
   findTargetToken,
+  getParseTreeAndTokens,
+  handleMergeClause,
   isComment,
   wantsSpaceAfter,
   wantsSpaceBefore,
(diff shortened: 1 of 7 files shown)


packages/react-codemirror-playground/src/App.tsx
```diff
@@ -112,10 +116,20 @@ export function App() {
               theme={darkMode ? 'dark' : 'light'}
               history={Object.values(demos)}
               schema={schema}
-              featureFlags={{ signatureInfoOnAutoCompletions: true }}
+              featureFlags={{
+                signatureInfoOnAutoCompletions: true,
+                format: formatFlag,
```
OskarDamkjaer, line 121: Do you feel we need a feature flag at the library level @ncordon? To me it'd be fine enough with only a feature flags at the VSCode / Query level. In query the keyboard command would still work, but I don't think people would accidentally run into that and if they did, i don't think there's any hard done. It'd be more about not advertising

packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -276,6 +276,28 @@ export class CypherEditor extends Component<
   editorView: React.MutableRefObject<EditorView> = createRef();
   private schemaRef: React.MutableRefObject<CypherConfig> = createRef();
 
+  /**
+   * Format query code
```
OskarDamkjaer, line 280:
```suggestion
   * Format Cypher query
 ```

File: packages/react-codemirror/src/CypherEditor.tsx [hunk 3]
```diff
@@ -382,6 +404,17 @@ export class CypherEditor extends Component<
           }),
         ]
       : [];
+    if (this.props.featureFlags && this.props.featureFlags.format) {
+      extraKeybindings.push({
+        key: 'Ctrl-Shift-f',
+        mac: 'Alt-Shift-f',
```
OskarDamkjaer, line 410: seems to match the vscode defaults, make sense

packages/react-codemirror-playground/src/App.tsx
```diff
@@ -32,6 +32,7 @@ RETURN count(*)`,
 type DemoName = keyof typeof demos;
 
 export function App() {
+  const formatFlag = false;
```
OskarDamkjaer, line 35: in any case i think it'd make sense to have it on in the demo application, could be an easy way for greg to try it out earlier/faster iterations

packages/language-support/src/formatting/formatting.ts
```diff
@@ -189,14 +192,19 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
         this.addSpace();
       }
     }
+
     if (node.symbol.type === CypherCmdLexer.EOF) {
       return;
     }
+    if (node.symbol.tokenIndex === this.targetToken) {
+      this.cursorPos = this.buffer.join('').length;
+    }
```
OskarDamkjaer, line 201: why are these two:
`this.cursorPos = this.buffer.join('').length;`
needed?

OskarDamkjaer (commented):
Nice that we keep the cursor location, worked well in my testing

packages/language-support/src/formatting/formatting.ts
simonthuresson: This code tracks the cursor's position after text formatting occurs. We note which token the cursor was at in the unformatted text. Then, as we process the formatting changes, once hitting the same token, we mark this as a rough spot of where the cursor will be.

packages/react-codemirror-playground/src/App.tsx
simonthuresson: Absolutely, I will remove this flag so the formatting is always visible in the demo.

packages/language-support/src/formatting/formatting.ts
```diff
@@ -346,8 +357,40 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+
+export function formatQuery(query: string): string;
+export function formatQuery(
+  query: string,
+  cursorPosition: number,
+): FormattingResultWithCursor;
+export function formatQuery(
+  query: string,
+  cursorPosition?: number,
+): string | FormattingResultWithCursor {
```
ncordon, line 373: Dummy question, but how does this work? How does the compiler know if I just pass a string I never expect a `FormattingResultWithCursor`?

packages/react-codemirror-playground/src/App.tsx
ncordon: This is fine to me, the library is still at experimental versions

packages/language-support/src/formatting/formatting.ts
ncordon: I was trying to test it out with comments (which I think might be a problem for preserving the cursor position because they are a full token and the cursor could be in the middle of any).

The cursor just disappears after I click on the `format query` button in codemirro playground. Does that functionality work for you?

Also it would be worth to add some tests?

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -79,3 +80,16 @@ export function getParseTreeAndTokens(query: string) {
   const tree = parser.statementsOrCommands();
   return { tree, tokens };
 }
+
+export function findTargetToken(tokens: Token[], cursorPosition: number) {
+  let targetToken = tokens[0];
```
ncordon, line 85: I think it's best to assume tokens could be empty?

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -79,3 +80,16 @@ export function getParseTreeAndTokens(query: string) {
...
+  for (const token of tokens) {
+    if (token.channel === 0) {
+      targetToken = token;
+    }
```
ncordon, line 89: What happens if the cursor was in a comment / space?

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -79,3 +80,16 @@ export function getParseTreeAndTokens(query: string) {
...
+    if (cursorPosition >= token.start && cursorPosition <= token.stop) {
+      break;
```
ncordon, line 91: This is personal preference so feel free to ignore, but for consistency with the code in the project we try not to use `break`

packages/language-support/src/formatting/formattingHelpers.ts
simonthuresson: As the comment/space will not appear in the AST tree, we do not want to search for it there. Therefore we ignore does when trying to find the tokenIndex of the cursor position

packages/language-support/src/formatting/formatting.ts
simonthuresson: I think it knows because if only a string is passed as an argument, only a string should be return. This is called [function overloading](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads)

packages/language-support/src/formatting/formatting.ts
simonthuresson: I think we discussed about the cursor disappearing. 

There was absolutely tests missing, they are now added.

packages/language-support/src/formatting/formattingHelpers.ts
simonthuresson: I totally understand, I have changed so we return directly instead

packages/language-support/src/formatting/formatting.ts
OskarDamkjaer: >The cursor just disappears after I click on the format query button in codemirror playground. Does that functionality work for you?

Clicking the button moves the focus out of the editor, so to keep the cursor position, you'd use the keyboard shortcut

OskarDamkjaer: Ahh, I was a little confused, it looked like this was part of the constructor of the class, i see now that it's in the visitor for terminals :+1:

OskarDamkjaer (approved):

PR #336, opened 2025-01-29
Adds subquery and parameter handling to formatter

tomasnyberg:
### Description
Adds two things missing from the styleguide recommendations:
1. (bugfix) - no space between dollar sign and parameter name
2. Subquery clauses shuold get indentation like the exists clauses, e.g. this query

```
CALL () { RETURN 'hello' AS innerReturn }
```
should be formatted into

```
CALL () {
  RETURN 'hello' AS innerReturn
}
```

I've also added a couple of examples from the styleguide into the tests, such as the "no padding space within function call parentheses" one.

### Testing
- npm run test 
- unit tests included in this PR for the new features


ncordon (approved):
LGTM

PR #338, opened 2025-01-30
Adds property checks to formatter

tomasnyberg:
### Description

I was reading through the [scalafmt thesis](https://geirsson.com/assets/olafur.geirsson-scalafmt-thesis.pdf) and came across section 4.2 which outlined some property based testing they used during development. They describe these as having "played a vital role in the development of scalafmt and gave us confidence that the [formatting algorithms] behave well against real world input". The three checks they relied on were

1. Can-format: If the scala compiler can parse the code, the formatter should be able to format it
2. AST-integrity: the AST should stay identical after formatting
3. Idempotency: Formatting twice should give the same output

Given that scalafmt is hugely successful, I figured we want to follow their advice. I already added nr 2 in a previous PR, but this PR introduces an error handler to ensure that if the input cannot be parsed, it throws, and otherwise it should succeed. For 3, I added another check in to all tests that verify idempotency.

PS
Throwing when trying to format a syntax error might not be the best approach. I'm happy to amend that part.

### Testing
- npm run test
- 1 new unit test
- tried formatting a syntactically incorrect query in the codemirror playground and the error from the formatter was caught silently

### DIFF
--- a/packages/language-support/src/formatting/formattingHelpers.ts
+++ b/packages/language-support/src/formatting/formattingHelpers.ts
@@ -1,7 +1,10 @@
 import {
   CharStreams,
+  CommonToken,
   CommonTokenStream,
+  ErrorListener as ANTLRErrorListener,
   ParseTree,
+  Recognizer,
   TerminalNode,
   Token,
 } from 'antlr4';
@@ -12,6 +15,22 @@ import CypherCmdParser, {
   UnescapedSymbolicNameString_Context,
 } from '../generated-parser/CypherCmdParser';
 import { lexerKeywords, lexerOperators } from '../lexerSymbols';
+
+class FormatterErrorsListener implements ANTLRErrorListener<CommonToken> {
+  syntaxError<T extends Token>(
+    _r: Recognizer<CommonToken>,
+    offendingSymbol: T,
+    line: number,
+    column: number,
+  ) {
+    throw new Error(
+      `Could not format due to syntax error at line ${line}:${column} near "${offendingSymbol?.text}"`,
+    );
+  }
+  public reportAmbiguity() {}
+  public reportAttemptingFullContext() {}
+  public reportContextSensitivity() {}
+}
 
 export function handleMergeClause(
   ctx: MergeClauseContext,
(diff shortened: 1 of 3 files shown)


packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -13,6 +16,22 @@ import CypherCmdParser, {
 } from '../generated-parser/CypherCmdParser';
 import { lexerKeywords, lexerOperators } from '../lexerSymbols';
 
+class FormatterErrorsListener implements ANTLRErrorListener<CommonToken> {
```
ncordon, line 19: I'm not entirely sure whether we should be doing this. Because even if the query is not 100% correct, antlr should still be able to parse a portion of it (and not discard the rest), so we can still format what we could parse?

packages/language-support/src/tests/formatting/formatting.test.ts
```diff
@@ -6,11 +6,15 @@ function verifyFormatting(query: string, expected: string): void {
...
   const formattedStandardized = standardizeQuery(formatted);
+  // AST integrity check
   if (formattedStandardized !== queryStandardized) {
     throw new Error(
       `Standardized query does not match standardized formatted query`,
     );
   }
+  // Idempotency check
+  const formattedTwice = formatQuery(formatted);
+  expect(formattedTwice).toEqual(formatted);
```
ncordon, line 17: This makes sense

packages/language-support/src/formatting/formattingHelpers.ts
tomasnyberg: Yeah I'm happy to change the behavior here, but I'm not sure what's best. My reasoning was that it might produce unexpected results when the syntactically incorrect parts interact with the syntactically incorrect ones, i.e. what if 

```
MATCH (n) syntaxerror
RETURN 5
```

gets turned into something like


```
MATCH (n)
syntaxerrorRETURN 5
```
breaking the query completely. I know some formatters flat out refuse to format anything that is not syntactically correct (my json formatter behaves like this), but some, like prettier, don't seem to have a problem with it. I can play around with it a bit and see if it's possible to get reasonable behavior on syntax errors.

OskarDamkjaer: In my experience prettier also doesn't format until the syntax is correct, could be because i use it with eslint. We can look into if it's tricky, but imo it'd make sense to keep the scope smaller to start with and only concern ourselves with valid queries (unless it's trivial to handle broken ones)

tomasnyberg: I played around with it a bit and while I think it might be possible to format everything we can despite errors, it would be quite tricky and probably come with tons of edge cases. I think it makes a lot more sense to just give up, especially considering many (most?) formatters don't tolerate syntax errors. It's also nice to keep it out of scope, as Oskar mentioned.

ncordon (approved):

PR #341, opened 2025-01-31
Fixes union, case expr and namespace bugs in formatter

tomasnyberg:
### Description
While working on line breaks I found the following cases which were not getting formatted properly:

- Union statements (should be newline separated twice and with two indents)
- case expressions (each case shuold be on a newline and indented
- namespaced functions (calls to e.g. apoc.function got spaces in the middle since they were treated as operators.)
Actual code examples can be found in the test part of the PR.

This PR fixes those bugs and adds tests for them.

### Testing
- npm run test
- 4 unit tests added in this PR


packages/language-support/src/tests/formatting/formatting.test.ts
```diff
@@ -262,6 +262,102 @@ WITH N, RAND() AS Rand, $pArAm AS MAP
...
+UNION DISTINCT
+  ALL 
+  MATCH (m:Movie)
+  RETURN m.title AS name
+}
+RETURN name, count(*) AS count ORDER BY count`;
+    const expected = `CALL () {
+  MATCH (a:Actor)
+  RETURN a.name AS name
+    UNION DISTINCT
```
ncordon, line 315: Is it correct the `ALL` has disappeared from here?

packages/language-support/src/tests/formatting/formatting.test.ts
```diff
@@ -262,6 +262,102 @@ WITH N, RAND() AS Rand, $pArAm AS MAP
...
+    const expected = `MATCH (n:Person)
+RETURN n.name,
+CASE n.age
+  WHEN = 0, = 1, = 2 THEN "Baby"
+  WHEN <= 13 THEN "Child"
+  WHEN < 20 THEN "Teenager"
+  WHEN < 30 THEN "Young Adult"
+  WHEN > 1000 THEN "Immortal"
+  ELSE "Adult"
+END AS result`;
```
ncordon, line 358: Really nice!

ncordon (approved):

packages/language-support/src/tests/formatting/formatting.test.ts
tomasnyberg: It is not correct, nice catch! Ironically, the AST integrity check from #338 would have caught this since it's a missing node.

Will fix this when I'm back.

PR #345, opened 2025-02-04
Fixes bug with syntax highlighting of preparser options

ncordon:
## What
To highlight the tokens we were:

1. Doing a first pass using the lexer
2. Fixing the highlighting with the parsing tree

The problem is we were storing the colouring in a map with key `line,column`, but we were not taking into account how many characters that highlighting affected. So for cases like:

```
CYPHER runtime = "some unfinished 
    string
```

we would not colour the string correctly.

This PR adds the length of the colouring as a third coordinate in that map key, so we can solve those edge cases.

## Why

We were getting the highlighting of multiline constructs in the preparser incorrectly

### DIFF
--- a/packages/language-support/src/syntaxColouring/syntaxColouringHelpers.ts
+++ b/packages/language-support/src/syntaxColouring/syntaxColouringHelpers.ts
@@ -13,8 +13,11 @@ export interface TokenPosition {
   startOffset: number;
 }
 
-export function tokenPositionToString(tokenPosition: TokenPosition): string {
-  return `${tokenPosition.line},${tokenPosition.startCharacter}`;
+export function tokenPositionToString(
+  tokenPosition: TokenPosition,
+  tokenLength: number,
+): string {
+  return `${tokenPosition.line},${tokenPosition.startCharacter},${tokenLength}`;
 }
 
 export interface ParsedCypherToken {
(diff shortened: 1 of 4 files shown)


packages/language-support/src/syntaxColouring/syntaxColouringHelpers.ts
```diff
@@ -13,8 +13,11 @@ export interface TokenPosition {
   startOffset: number;
 }
 
-export function tokenPositionToString(tokenPosition: TokenPosition): string {
-  return `${tokenPosition.line},${tokenPosition.startCharacter}`;
+export function computeTokenKey(
+  tokenPosition: TokenPosition,
+  tokenLength: number,
+): string {
+  return `${tokenPosition.line},${tokenPosition.startCharacter},${tokenLength}`;
```
anderson4j, line 20: From my understanding - so we get unique key (and multiple entries in tokens-map) when line1=line2 and start1=start2 but length differ

packages/language-support/src/syntaxColouring/syntaxColouringHelpers.ts
```diff
@@ -191,14 +194,20 @@ export function sortTokens(tokens: ParsedCypherToken[]) {
     const lineDiff = a.position.line - b.position.line;
     if (lineDiff !== 0) return lineDiff;
 
-    return a.position.startCharacter - b.position.startCharacter;
+    const colDiff = a.position.startCharacter - b.position.startCharacter;
+    if (colDiff !== 0) return colDiff;
+    else return b.length - a.length;
```
anderson4j, line 199: So we reach longer length first in sorted array and thus remove the overlapping smaller ones

anderson4j (approved):
1 nitpick, and 2 comments so you could double check that I got it, otherwise looks good

packages/language-support/src/syntaxColouring/syntaxColouringHelpers.ts
ncordon: Yep that's correct

PR #346, opened 2025-02-04
Fixes bug splitting the statements into commands

ncordon:
…ands

### DIFF
--- a/packages/language-support/src/parserWrapper.ts
+++ b/packages/language-support/src/parserWrapper.ts
@@ -610,8 +610,8 @@ function parseToCommand(
 
       return { type: 'parse-error', start, stop };
     }
-    const statement = inputstream.getText(start.start, stop.stop);
-    return { type: 'cypher', statement, start: start, stop: stop };
+    const statement = inputstream.getText(start.start, tokens.at(-1).stop);
+    return { type: 'cypher', statement, start: start, stop: tokens.at(-1) };
   }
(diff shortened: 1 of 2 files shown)


packages/language-support/src/parserWrapper.ts
```diff
@@ -610,8 +610,9 @@ function parseToCommand(
 
       return { type: 'parse-error', start, stop };
     }
-    const statement = inputstream.getText(start.start, stop.stop);
-    return { type: 'cypher', statement, start: start, stop: stop };
+    const stopToken = stop ?? tokens.at(-1);
```
anderson4j, line 613: Double checking my understanding - I would explain like:
"When ANTLR can't finish the parsing (but can tokenize) because a comment/string/escaped sequence is unfinished we get here with stop===null"

anderson4j (commented):
Mostly LGTM, but would we not get the same issue with unfinished strings/escaped sequences? (If so, we might as well add tests for these too)

And if my summary of the issue (see comment) is correct, would we encounter it for single line comments?

packages/language-support/src/parserWrapper.ts
ncordon: The problem if we just a comment is that's not passed to the parser, so the CST is empty, therefore because the parser expects something, it errors and cannot find a stop. Strings and backticked elements are not on the hidden channel, so they are passed to the parser

PR #347, opened 2025-02-04
Adds error on incorrect cypher version

anderson4j:
Before we didnt get warnings of "incorrect version" when using a cypher language version that is not 5 or 25 in the query. Now:
`CYPHER 50 MATCH (n) RETURN n` -> "50 is not a valid option for cypher version. Valid options are: 5, 25" 
also found that we were misaligned with the preparser from the coreDB on what is parsed as "version" so they get version errors on for ex.
`CYPHER 007 MATCH (n) RETURN n` -> "007 is not a valid option for cypher version. Valid options are: 5, 25" We did not get this error before, but parser adjustment fixes this


packages/language-support/src/antlr-grammar/CypherPreLexer.g4
```diff
@@ -9,4 +9,7 @@ PROFILE:
    P R O F I L E;
 
 CYPHER:
-   C Y P H E R;
\ No newline at end of file
+   C Y P H E R;
+
+VERSION: 
+  ([0-9] | '.')+;
```
anderson4j, line 15: This shadows DECIMAL_DOUBLE and UNSIGNED_DECIMAL_INTEGER - not nice (in main project it's no problem since they tokenize/parse preparser and rest individually)

anderson4j: Could:
A: reset to old
B: use a more accurate combo of existing tokens to capture the same logic, like (UNSIGNED_DECIMAL_INTEGER | DECIMAL_DOUBLE | DOT)+;
C: Do one of the above and card for the future to do preparsing in a separate step, to stay in line with coreDB and not need workarounds like this (we could then use their preparser straight up I suppose)

packages/language-support/src/antlr-grammar/CypherPreParser.g4
```diff
@@ -29,4 +29,4 @@ cypherOptionName:
    IDENTIFIER;
 
 cypherVersion:
-   UNSIGNED_DECIMAL_INTEGER;
\ No newline at end of file
+   (EXTENDED_IDENTIFIER | UNSIGNED_DECIMAL_INTEGER | DOT | DECIMAL_DOUBLE)+;
```
anderson4j, line 32: Since tokenization picks the "largest possible token" always, a version like 1.5 would be tokenized as DECIMAL_DOUBLE, so we would this token here too.

same thing means 007 is tokenized to EXTENDED_IDENTIFIER since it doesnt match a single DECIMAL_DOUBLE/UNSIGNED_DECIMAL_INTEGER. EXTENDED_IDENTIFER could be ANYTHING with a lot of letters, but I guess stuff following CYPHER directly wont be mistaken anyway since tests pass.

packages/language-support/src/parserWrapper.ts
```diff
@@ -431,14 +438,27 @@ class CypherVersionCollector extends ParseTreeListener {
...
   exitEveryRule(ctx: unknown) {
     if (ctx instanceof CypherVersionContext) {
+      const parsedVersion = ctx.getText();
       this.cypherVersion =
         ctx.getText() === '5'
           ? 'CYPHER 5'
           : ctx.getText() === '25'
           ? 'CYPHER 25'
           : undefined;
+      if (parsedVersion !== '5' && parsedVersion !== '25') {
```
anderson4j, line 448: We have quite a few checks like ctx.getText() ===  or !== '5'/'25', where we manually type the version strings. Maybe we should have a global list like ValidCypherVersions ['CYPHER 5', 'CYPHER 25'] or ['5', '25] so we can refer to that, and when we add a version in the future we wont have to search for all the places with '5'/'CYPHER 5' etc?

anderson4j: I guess nicest might be ValidCypherVersions : CypherVersion[] = [...] so above we could do, 
foreach ver -> if 'CYPHER ' + ctx.getText() === ver -> parsedVersion = ver

anderson4j: And for cases where we only want the options, we could have a helper method like getValidVersionNumbers(validCypherVersions)

packages/language-support/src/tests/syntaxValidation/semanticValidation.test.ts
```diff
@@ -102,13 +102,89 @@ describe('Semantic validation spec', () => {
...
+            line: 0,
+          },
+        },
+        severity: 2,
+      },
+    ]);
+    expect(diagnostics2).toEqual([
+      {
+        message:
+          '007 is not a valid option for cypher version. Valid options are: 5, 25',
```
anderson4j, line 152: This showed an unexpected case where 007 was tokenized to something I didnt expect. Maybe I should add some additional tests with unexpected versions that would be matched by the coreDB preparser, like "1.2.3", "0.1.2", ".12"

packages/language-support/src/antlr-grammar/CypherPreLexer.g4
anderson4j: After discussion, ive gone for C with main solution from B

packages/language-support/src/parserWrapper.ts
```diff
@@ -41,6 +41,7 @@ export interface ParsedStatement {
   // it's the one that tries to parse until the EOF
   ctx: StatementsOrCommandsContext;
   syntaxErrors: SyntaxDiagnostic[];
+  preparserErrors: SyntaxDiagnostic[];
```
ncordon, line 44: In this other pr: https://github.com/neo4j/cypher-language-support/pull/342 we are surfacing the syntax errors back again, so this might not be needed

packages/language-support/src/parserWrapper.ts
```diff
@@ -200,12 +201,17 @@ export function createParsingResult(query: string): ParsingResult {
       if (!_internalFeatureFlags.consoleCommands) {
         syntaxErrors.push(...errorOnNonCypherCommands(collectedCommand));
       }
+      const preparserErrors = cypherVersionCollector.invalidVersionError
+        ? [cypherVersionCollector.invalidVersionError]
+        : [];
+      cypherVersionCollector.resetError();
```
ncordon, line 207: This `resetError` is a bit hacky in my opinion. Let's avoid calculating the error inside the collector.

You could define a `ParsedCypherVersion` type (similar to `ParsedProcedure`):

```typescript
type ParsedCypherVersion = {
  cypherVersion: CypherVersion;
  range: Range;
  offsets: {
    start: number;
    end: number;
  };
};
```
and compute the error from outside the collector:


```typescript
      const preparserErrors: SyntaxDiagnostic[] = !cypherVersion
        ? [
            {
              severity: DiagnosticSeverity.Error,
              range: cypherVersion.range,
              offsets: cypherVersion.offsets,
              message: `... is not a valid option for cypher version. Valid options are: 5, 25`,
            },
          ]
        : [];
```

packages/language-support/src/parserWrapper.ts
ncordon: Yeah it sounds sensible to introduce a single point where we have the cypher versions listed so give it a try

packages/language-support/src/tests/syntaxValidation/semanticValidation.test.ts
```diff
@@ -102,13 +102,89 @@ describe('Semantic validation spec', () => {
...
+          },
+          start: {
+            character: 7,
+            line: 0,
+          },
+        },
+      },
+      {
+        message:
+          'The use of nodes or relationships for setting properties is deprecated and will be removed in a future version. Please use properties() instead.',
```
anderson4j, line 170: Forgot to mention this in my last pre-review, but when the parsed version is faulty we have the unique scenario of not knowing what version the user wants. I set it so we get other errors from semantic analysis with cypher 5 in the initial semantic worker error PR and that's still used. 

So as you can see here, we dont only get the "version error" but also any errors we would get if we didn't use parsedVersion. If you think this is misleading I suppose we could prevent other errors while parsedVersion is faulty.

ncordon: Yeah good point. I think it could make sense to maybe have a:

```
syntaxErrors: SyntaxDiagnostic[],
cypherVersionError: SyntaxDiagnostic | undefined
```

and not show any other error until the `cypherVersionError` is `undefined`?

packages/language-support/src/parserWrapper.ts
anderson4j: I don't like the proposed approach since we would be introducing a whole new object, when we have the info to create the error right away. I had a look though, and realised we recreate the cypherVersionCollector any time we parse anyway, so resetError was unnecessary. 

If you dislike looking at an attribute the listener we could pass in a list or something to the constructor of the collector and fill that when we get errors. like:
const versionErrors = new ArrayList<string>();
const cypherVersionCollector = new CypherVersionCollector(versionErrors);
...
{
...
preparserErrors: versionErrors
...
}

anderson4j: Also, for multiple statements separated by semicolon if this is counted as one parse I guess we could get multiple "invalid version"-errors, so then we would want a list anyway

packages/language-support/src/parserWrapper.ts
anderson4j: switched so we have 
  cypherVersionError: SyntaxDiagnostic | undefined;
now, and checking if that is defined, so considering this resolved

packages/language-support/src/types.ts
```diff
@@ -10,6 +10,11 @@ export type ReturnDescription = {
 export const cypherVersions = ['CYPHER 25', 'CYPHER 5'];
 export type CypherVersion = (typeof cypherVersions)[number];
 
+export const validCypherVersions: CypherVersion[] = ['CYPHER 5', 'CYPHER 25'];
```
ncordon, line 13: You don't need the `validCypherVersions`, it's literally the `cypherVersions` above.

packages/language-support/src/types.ts
```diff
@@ -10,6 +10,11 @@ export type ReturnDescription = {
...
+export const validCypherVersionNumbers: string[] = validCypherVersions.map(
+  (item) => item.replace(/\D+/g, ''),
+);
```
ncordon, line 16: I find a regex an overkill. It's fine with just an array of the versions:
```typescript
export const validCypherVersionNumbers: string[] = ['5', '25']
```

packages/language-support/src/types.ts
anderson4j: ah, I forgot it was changed from
export type CypherVersion = 'CYPHER 25' | 'CYPHER 5'

ncordon:
Good work! I'll merge this because I want to make a pr on top of it.

PR #351, opened 2025-02-06
Completes CYPHER 25 if the feature flag is enabled

ncordon:
We want to get the `CYPHER 25` completion if the corresponding feature flag is enabled


anderson4j (approved):
Some spelling stuff for a variable name and one of the test names that could use a quick fix, but logic LGTM

PR #353, opened 2025-02-10
Adds duplicated placeholder code for second iteration of Cypher formatter

tomasnyberg:
### Description
This introduces the first draft of the second iteration of the cypher formatter, intended to be able to handle line breaks in a reasonably good way. The PR does not change any of the existing files, and instead creates new files with a "v2" extension that are copies of the old files, with new changes introduced. This is done because of the following reasons:
1. It's hard to consolidate the v2 version with the existing code without changing existing functionality
2. The new version gives very different outputs in cases where line breaks are needed, and that output is not necessarily production ready yet
3. We want to be able to iteratively release updates for this new iteration instead of introducing everything at once

Given that the files are fully copied, the easiest way to review is probably to check this out locally and open a diff view in your editor of choice, e.g. 

```
nvim -d formatting.ts formattingv2.ts
```
since otherwise everything will look like a change.

### Changes (v2 compared to v1)
This PR introduces significant architectural changes to the formatter, since that is necessary for the line break logic. Much of it is inspired by [this blog post](https://journal.stuffwithstuff.com/2015/09/08/the-hardest-program-ive-ever-written/)

- Instead of visiting and adding tokens directly to our output, we now consolidate (possibly multiple) tokens into `Chunks`
  - This comes with quite a few changes in formatting.ts since spaces between tokens is now encoded differently. For instance, we now avoid spaces where needed instead of explicitly adding them when needed
 - formattingHelpers introduces ~10 new constructs to help with the best first graph search and constructing the output
 - formattingHelpers introduces ~6 new methods for turning the lists of chunks into the formatted output through the graph search
A simplified overview of the flow from source code to formatted source code would be something like this:

image

### Testing
- All the old formatting tests pass (except for one, commented out with a TODO since it will require some architectural changes to get right)
  - The tests are copied over to a v2 test file as well to ensure it's formatted by the new version
- Ten new queries that have lines that go beyond the max width of 80, and a test that verifies that linebreaks keep them within the max width

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10934,6 +10934,15 @@
         "he": "bin/he"
       }
     },
+    "node_modules/heap-js": {
+      "version": "2.6.0",
+      "resolved": "https://registry.npmjs.org/heap-js/-/heap-js-2.6.0.tgz",
+      "integrity": "sha512-trFMIq3PATiFRiQmNNeHtsrkwYRByIXUbYNbotiY9RLVfMkdwZdd2eQ38mGt7BRiCKBaj1DyBAIHmm7mmXPuuw==",
+      "license": "BSD-3-Clause",
+      "engines": {
+        "node": ">=10.0.0"
+      }
+    },
     "node_modules/highlight.js": {
       "version": "10.7.3",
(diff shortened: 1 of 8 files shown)


tomasnyberg:
Closing this PR in favor of a new one (posting soon) since it's outdated.

PR #355, opened 2025-02-11
Provides VSCode setting to disable linting

ncordon:
Because linting at the moment is not tailored to a particular neo4j version so it can be annoying to get big chunks of the query linted when we are connected to old versions.


ncordon:
Note for myself: missing a changelog here

packages/vscode-extension/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Adds support for parameters to have escaped names
 - Adds autocompletions for Cypher version
 - Adds support for CYPHER preparser options (runtime, planner, etc)
+- Adds setting neo4j.features.linting to disable linting of Cypher files
```
anderson4j, line 9: Shouldnt this be on 1.8.0 or whatever we call the next release? 1.7.0 is released

anderson4j (changes requested):
I like adding stuff to the changelog in each PR (I havent been doing it until now, but will start). 1.7.0 is the last release though so we should add 1.8.0 I guess (multiple connections feels like big enough of a feature to bump minor version) and add this text there

anderson4j (approved):
Looks like something went wrong when I tried to release 1.7.0 a long time ago, so current can be 1.7.0

PR #357, opened 2025-02-12
Supports multiple connections in the neo4j VSCode plugin

ncordon:
Adds the ability to maintain several connections in the VSCode plugin.


packages/vscode-extension/tests/webviewUtils.ts
```diff
@@ -6,13 +9,25 @@ export async function waitUntilNotification(
...
       );
 
-      return messages.includes(notification);
+      const found = notificationsAndMsgs.find(
+        (value) => value.msg === notification,
+      );
+      if (found) {
+        await found.notification.dismiss();
+        return true;
+      } else {
```
ncordon, line 26: We need this so we always close a notification when we've found it. Otherwise notifications keep piling up and if I connect to an instance and I disconnect, I would still be able to find the notification 'Connected to neo4j'

packages/vscode-extension/tests/specs/webviews/connection.spec.ts
```diff
@@ -1,63 +1,47 @@
...
-      const menuText = await menuItem.elem.getText();
-      return menuText === item;
-    });
-
-    if (connect) {
-      const connectOption = await connect.elem;
-      await connectOption.click();
-    }
-  }
-
```
ncordon, line 47: I just extracted this to common helpers

packages/vscode-extension/tests/specs/webviews/queryExecution.spec.ts
```diff
@@ -1,76 +1,111 @@
...
+      await expect(text).toContain('Variable `n` not defined');
+    });
   });
 
-  test('should error on invalid cypher', async () => {
-    await openFixtureFile(browser, 'invalid.cypher');
-    await workbench.executeCommand(CONSTANTS.COMMANDS.RUN_CYPHER);
-    const webviews = await workbench.getAllWebviews();
-    await expect(webviews.length).toBe(1);
+  test('results should depend on the instance we are connected to', async function () {
```
ncordon, line 81: This is really the only new test from this file, the rest is restructuring

packages/vscode-extension/tests/fixtures/.env
```diff
@@ -1,6 +0,0 @@
-NEO4J_SCHEME=neo4j
-NEO4J_HOST=localhost
-NEO4J_PORT={PORT}
-NEO4J_USER=neo4j
-NEO4J_DATABASE=neo4j
-NEO4J_PASSWORD={PASSWORD}
```
ncordon, line 6: I realized we don't need this file anymore because it was used to pass the port and password to the tests, but we can do that via a better mechanism

packages/vscode-extension/tests/runApiAndUnitTests.ts
```diff
@@ -26,13 +26,16 @@ async function main() {
       launchArgs: [path.join(__dirname, '../../tests/fixtures/')],
       extensionDevelopmentPath,
       extensionTestsPath,
-      extensionTestsEnv: { CYPHER_25: 'true' },
+      extensionTestsEnv: {
+        CYPHER_25: 'true',
+        NEO4J_PORT: neo4jInstance.getMappedPort(7687).toString(),
```
ncordon, line 31: This is the proper way to pass env variables to the tests

packages/vscode-extension/tests/setupTestContainer.ts
```diff
@@ -19,31 +18,8 @@ export async function createAndStartTestContainer(
...
-
-function updateDotenvFile(port: number, password: string) {
-  const dotenvPath = path.join(__dirname, '../../tests/fixtures/');
-  const dotenvTemplate = path.join(dotenvPath, '.env');
-  const dotenvFile = path.join(dotenvPath, '.env.test');
-  fs.copyFileSync(dotenvTemplate, dotenvFile);
-  setSetting(dotenvFile, /\{PORT\}/g, port.toString());
-  setSetting(dotenvFile, /\{PASSWORD\}/g, password);
-  dotenv.config({ path: dotenvFile });
-}
```
ncordon, line 49: Just cleans up these because we can pass the env variables better to the tests

anderson4j (commented):
I had one question, otherwise looks good. I tested around a bit with it in the playground though, and found 2 things that were a bit unexpected.

1 - (and I think this one is how it was before too) - if I want to close down a "create new connection" pop-up, I dont have an x in the corner, or something like pressing esc or outside the pop-up to close it which is what I first expected. I then found that the pop-up actually opens like a new tab/file in the playground. If you agree that this is less intuitive (and its not too hard to change) maybe we should change it in a follow-up PR?

2 - 
Creating new connections - to me it works like this:
if active db at given connection info (port etc) exists, creating connection with password x only works if x is the correct password. Too many tries gets your session blocked, You are also allowed to create multiple sessions with the exact same info (including password) this way

If db at connection info is not active, you get a pop-up ("Unable to connect...") and are allowed to add the connection. 

The disparity of how this works is confusing to me, I would expect to either 
A: never be able to add a db I can connect to, or 
B: to always be allowed to add connections that dont exist at the moment. 
(As a user, I would prefer option B, since I might want to create 3 dbs with same info but different passwords, and having to start/stop dbs between to create connections (instead of doing all at once) seems annoying)

Also once these connections are created, the order of creation and password length is the only way to tell them apart, so I could easily see myself forgetting which has which password. Possible solutions - nicknames for connections displayed next to the existing info. Or maybe we just dont allow multiple connections to same scheme/host/port/user, but I think I prefer the other one.

anderson4j:
Just had another connections-bug while trying to run a query:
image
How to reproduce: connect to db, delete other connection. Connected connection says "connected" but actually behaves like disconnected

packages/vscode-extension/tests/specs/webviews/queryExecution.spec.ts
```diff
@@ -1,76 +1,111 @@
...
+    if (os.platform() === 'darwin') {
```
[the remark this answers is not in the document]
ncordon: These are not included in the `test:apiAndUnit`, they are webview tests, so they use Playwright inside to be able to do more user-like interactions. These webview tests are called with `test:webviews`.

The reason why they don't work on a Mac is because the little overlay which shows on the connections right click:


is native in the OS, which means is not part of the browser window, it's something controlled by the OS the tests have no access on.

package.json
```diff
@@ -35,6 +35,8 @@
   "devDependencies": {
     "@changesets/cli": "^2.27.8",
     "@typescript-eslint/eslint-plugin": "^7.8.0",
+    "@wdio/cli": "^8.41.0",
+    "@wdio/local-runner": "^8.41.0",
```
ncordon, line 39: This is to fix debugging the webview tests from VSCode

packages/vscode-extension/package.json
```diff
@@ -36,9 +36,15 @@
   "contributes": {
     "commands": [
       {
-        "command": "neo4j.manageConnection",
+        "command": "neo4j.editConnection",
         "category": "Neo4j",
-        "title": "Manage Connection",
+        "title": "Edit connection",
```
ncordon, line 41: From a call with Greg, I got the feedback we should be renaming the `Manage Connection` to `Edit connection`

ncordon: And also that we should be showing `Connection` in lower case in all of the menus, hence why in many places this PR rewrites that

packages/vscode-extension/src/commandHandlers.ts
```diff
@@ -74,9 +73,12 @@ export async function saveConnectionAndDisplayConnectionResult(
     connection,
     password,
   );
+  const errorDetail = result?.error?.friendlyMessage;
 
-  if (!result.success && result.retriable) {
```
ncordon, line 78: After a discussion with Greg: we should always be offering the option to store the connection when credentials or url are invalid.

The difference here was that if I was trying with an invalid url, I would be offered to store the connection regardless, whereas if I was trying with invalid credentials but the url was correct, it would not offer me to store.

packages/vscode-extension/src/commandHandlers.ts
```diff
@@ -74,9 +73,12 @@ export async function saveConnectionAndDisplayConnectionResult(
...
-    const result = await displaySaveConnectionAnywayPrompt();
+  if (!result.success) {
+    const result = await displaySaveConnectionAnywayPrompt(
+      errorDetail ? `${errorDetail}.` : errorDetail,
```
ncordon, line 80: Just adds a trailing `.` to the error message

packages/vscode-extension/src/connectionService.ts
```diff
@@ -23,10 +23,11 @@ export type State = 'inactive' | 'activating' | 'active' | 'error';
 export type Connection = {
   key: string;
   scheme: Scheme;
+  name?: string;
```
ncordon, line 26: Adds an optional display name to the connection

packages/vscode-extension/src/treeviews/connectionTreeDataProvider.ts
```diff
@@ -123,9 +123,15 @@ export class ConnectionTreeDataProvider
...
-    return connection.port
+    const connectionString = connection.port
       ? `${connection.user}@${connection.scheme}://${connection.host}:${connection.port}`
       : `${connection.user}@${connection.scheme}://${connection.host}`;
+
+    const nameString = connection.name ?? '';
+
+    return nameString
+      ? `${nameString} [${connectionString}]`
+      : connectionString;
```
ncordon, line 134: If we have a display name, we concatenate `displayName [url]` otherwise we show `url`

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -45,6 +46,7 @@ export class ConnectionPanel {
     this._extensionPath = extensionPath;
     this._connection = connection;
     this._password = password;
+    this._editConnection = connection ? true : false;
```
ncordon, line 49: Now that we retain information about the connection when we change to another tab,  we need this to distinguish between a connection that is being edited vs one that we are adding.

packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -265,7 +276,7 @@ export class ConnectionPanel {
                     }" data-invalid="${this.authIsInvalid()}" />
                   </div>
                   <div class="form--actions">
-                    <input id="save-connection" type="submit" value="Save Connection" />
+                    <input id="save-connection" type="submit" value="Save & Connect" />
```
ncordon, line 279: More feedback from Greg: the button should say `Save & Connect` to make clear the action we are about to perform.

This is similar to the MongoDB extension.

packages/vscode-extension/syntaxes/cypher.json
```diff
@@ -38,7 +38,7 @@
...
...T|FUNCTION|FUNCTIONS|GRANT|GRAPH|GRAPHS|GROUP|GROUPS|HEADERS|HOME|ID|IF|IMMUTABLE|IMPERSONATE|IN|INDEX|INDEXES|INF|INFINITY|INSERT|INT|INTEGER|IS|JOIN|KEY|LABEL|LABELS|LANGUAGE|LEADING|LIMIT|LIST|LOAD|LOCAL|LOOKUP|MANAGEMENT|MAP|MATCH|MERGE|NAME|NAMES|NAN|NEW|NFC|NFD|NFKC|NFKD|NODE|NODETACH|NODES|NONE|NORMALIZE|NORMALIZED|NOT|NOTHING|NOWAIT|NULL|OF|OFFSET|ON|ONLY|OPTION|OPTIONAL|OPTIONS|OR|ORDER|PASSWORD|PASSWORDS|PATH|PATHS|PLAINTEXT|POINT|POPULATED|PRIMARY|PRIMARIES|PRIVILEGE|PRIVILEGES|PROCEDURE|PROCEDURES|PROPERTIES|PROPERTY|PROVIDER|PROVIDERS|RANGE|READ|REALLOCATE|REDUCE|REL|RELATIONSHIP|RELATIONSHIPS|REMOVE|RENAME|REPEATABLE|REPLACE|REPORT|REQUIRE|REQUIRED|RESTRICT|RETURN|REVOKE|ROLE|ROLES|ROW|ROWS|SCAN|SECONDARY|SECONDARIES|SEC|SECOND|SECONDS|SEEK|SERVER|SERVERS|SET|SETTING|SETTINGS|SHORTEST|shortestPath|SHOW|SIGNED|SINGLE|SKIP|START|STARTS|STATUS|STOP|VARCHAR|STRING|SUPPORTED|SUSPENDED|TARGET|TERMINATE|TEXT|THEN|TIME|TIMESTAMP|TIMEZONE|TO|TOPOLOGY|TRAILING|TRANSACTION|TRANSACTIONS|TRAVERSE|TRIM|TRUE|TYPE|TYPED|TYPES|UNION|UNIQUE|UNIQUENESS|UNWIND|URL|USE|USER|USERS|USING|VALUE|VECTOR|VERTEX|WAIT|WHEN|WHERE|WITH|WITHOUT|WRITE|XOR|YIELD|ZONE|ZONED|EXPLAIN|PROFILE|CYPHER)\\b",
```
ncordon, line 41: Unrelated to this pr, when we update the grammar we keep forgetting to update the textmate one 🙈

packages/vscode-extension/tests/specs/unit/executeCommands.spec.ts
```diff
@@ -160,7 +188,92 @@ suite('Execute commands spec', () => {
       );
     });
 
-    test('Dismissing delete Connection prompt should not show any messages', async () => {
+    test('Deleting another connection should not disconnect the current one', async () => {
```
ncordon, line 191: This test I've added because of the bug you caught @anderson4j

packages/vscode-extension/tests/specs/unit/executeCommands.spec.ts
```diff
@@ -160,7 +188,92 @@ suite('Execute commands spec', () => {
...
+        },
+      );
+
+      sandbox.assert.neverCalledWith(
+        showInformationMessageStub,
+        CONSTANTS.MESSAGES.DISCONNECTED_MESSAGE,
+      );
+    });
+
+    test('Deleting the in-use connection should disconnect that connection', async () => {
```
ncordon, line 241: This test I've also added because of the bug you caught @anderson4j

ncordon:
The new changes since you last reviewed are contained in commit [0dd35f1](https://github.com/neo4j/cypher-language-support/pull/357/commits/0dd35f13fdf262744f21dd9ff967b785ab60200b) @anderson4j

anderson4j:
Nice fixes/improvements! Looks like 1 thing is still broken though - storing connections when credentials are wrong. I now get the same pop-up as when connecting to nonexisting server, but when the url is valid, it does not store the connection (even though I get the notification "Neo4j connection saved")
image

ncordon:
> Nice fixes/improvements! Looks like 1 thing is still broken though - storing connections when credentials are wrong. I now get the same pop-up as when connecting to nonexisting server, but when the url is valid, it does not store the connection (even though I get the notification "Neo4j connection saved")

Good catch! I think I've solved that in [e1abfc9](https://github.com/neo4j/cypher-language-support/pull/357/commits/e1abfc96a0c973f743791ef21ceb3739968f716a)

anderson4j (approved):
Looks like it works! One nice final thing would be to have a test for the final bug too, something like "create connection for inactive db, then activate this db and see that we can use existing connection to connect". Not so experienced with these tests though, so not sure how doable it is. Will approve and leave it to you to decide :)

PR #361, opened 2025-02-17
Fixes function invocation

simonthuresson:
# Description
Adds visitor for functionInvocation. 
Formater before:
MATCH (n)
RETURN count( DISTINCT n) 
Formater after:
MATCH (n)
RETURN count(DISTINCT n) 

# Tests
 - One test added which 
 - All tests passes


packages/language-support/src/tests/formatting/formatting.test.ts
```diff
@@ -447,6 +447,14 @@ WHERE (n[listItem] IS FLOAT OR n[listItem] IS INTEGER)
 RETURN n`;
     verifyFormatting(query, expected);
   });
+
+  test('test for function invocation', () => {
+    const query = `MATCH (n)
```
OskarDamkjaer, line 452: The before and after look identical here, did you forget the space that you have in the description of the PR?

packages/vscode-extension/syntaxes/cypher.json
```diff
@@ -38,7 +38,7 @@
           "name": "keyword.operator"
         },
         {
```
OskarDamkjaer, line 40: What's the change here? is it intentional?

packages/language-support/src/tests/formatting/formatting.test.ts
OskarDamkjaer: would also be nice to change the "before" to cover part of the implementation that adds spaces between arguments as well

OskarDamkjaer (approved):

packages/language-support/src/tests/formatting/formatting.test.ts
simonthuresson: Totally forgot. I also improved the test to cover more of what was implemented.

packages/vscode-extension/syntaxes/cypher.json
simonthuresson: Definitely not intentional, is fixed now.

PR #364, opened 2025-02-17
Solves issue when filling in data in connection tab

ncordon:
This solves a bug when opening a connection form, which would lose the details when switching to another tab.

https://github.com/user-attachments/assets/20d247ca-37a8-4155-89a5-21f65393e764

### DIFF
--- a/packages/vscode-extension/src/webviews/connectionPanel.ts
+++ b/packages/vscode-extension/src/webviews/connectionPanel.ts
@@ -78,6 +78,7 @@ export class ConnectionPanel {
           Uri.file(path.join(extensionPath, 'resources')),
         ],
         enableScripts: true,
+        retainContextWhenHidden: true,
       },
     );
 
--- a/packages/vscode-extension/tests/webviewUtils.ts
+++ b/packages/vscode-extension/tests/webviewUtils.ts
@@ -1,3 +1,5 @@
+import { WebView } from 'wdio-vscode-service';
+
 export async function waitUntilNotification(
   browser: WebdriverIO.Browser,
   notification: string,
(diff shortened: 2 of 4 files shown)


packages/vscode-extension/src/webviews/connectionPanel.ts
```diff
@@ -78,6 +78,7 @@ export class ConnectionPanel {
           Uri.file(path.join(extensionPath, 'resources')),
         ],
         enableScripts: true,
+        retainContextWhenHidden: true,
```
ncordon, line 81: As mentioned [here](https://code.visualstudio.com/api/extension-guides/webview#persistence) this is not optimal because it leaves the tab running, but it's the most easy to maintain solution. I've tried other solutions and this seems the most easy one to maintain and we can accept the cost of leaving that tab running because when you fill the connection it will always close automatically.

packages/vscode-extension/tests/webviewUtils.ts
```diff
@@ -49,3 +51,33 @@ export async function openFixtureFile(
...
+  await webview.close();
+}
+
+export async function setText(webview: WebView, elem: string, value: string) {
+  await webview.open();
+  await (await $(elem)).clearValue();
+  await webview.close();
+  await webview.open();
+  await (await $(elem)).addValue(value);
+  await webview.close();
```
ncordon, line 72: This seems better to use than `elem.setValue()` because that internally clears the value and writes the new one, with the difference that it does not reopen the webview, so the locators can get outdated and the tests would fail.

anderson4j (approved):
Looks to work well!

Don't think I understand the comment on setText(), but maybe we can cover that next time we pair

PR #366, opened 2025-02-18
adds setting darkmode according to system pref

simonthuresson:
### Description
This PR updates react-codemirror-playground to check the system setting for darkmode. If darkmode is enabled on system, react-codemirror playground will default to darkmode. 

You might have talked about having this already and concluded you did not want to follow system settings, in that case remove this PR. 

### Tests
 - All tests passes


OskarDamkjaer (approved):
Nice! We should add this in Browser as well!

PR #372, opened 2025-02-26
Ports bug fixes from v2 formatter to v1

tomasnyberg:
## Description

There are a few bugs that we have fixed in the v2 version of the formatter but have not fixed in v1. In general, we don't necessarily care about porting fixes for all bugs, as we instead want to prioritize getting v2 out ASAP. But these bugs got extra attention and seem to be particularly painful, and were also pointed out in the #team-browser slack channel where we asked for feedback. I've ported the following fixes to v1 (from v2)

1. Newlined comments should not get moved to the end of the previous line
2. Label expressions should have correct spacing
3. QPP's should not have spaces around them.

## Testing
- Ran the AST integrity + idempotency check on 100k queries, and all passed.
- npm run test
- 7 new tests for the cases above

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -385,11 +385,14 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       return;
     }
     this.visit(ctx.LCURLY());
-    const n = ctx.UNSIGNED_DECIMAL_INTEGER_list().length;
-    this.visitTerminalRaw(ctx.UNSIGNED_DECIMAL_INTEGER(0));
+    let idx = 0;
+    if (ctx._from_) {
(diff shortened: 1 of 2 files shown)


packages/language-support/src/formatting/formatting.ts
```diff
@@ -333,6 +378,25 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     this.visitIfNotNull(ctx.rightArrow());
   };
 
+  visitQuantifier = (ctx: QuantifierContext) => {
```
tomasnyberg, line 381: Sidenote: This is actually incorrect in V2, as it might move an ending quantifier to the start. I've noted this and will fix it soon.

OskarDamkjaer:
@tomasnyberg also add a changeset so the changes get released :+1:

PR #374, opened 2025-02-27
Adds avoid breaking

simonthuresson:
### Description
This PR adds the functionality to tell the algorithm that between these chunks, you are never allowed to break. For example  `RETURN 1`, here it should be a avoidBreak() between the return and 1.

### DIFF
--- a/packages/language-support/src/formatting/formattingv2.ts
+++ b/packages/language-support/src/formatting/formattingv2.ts
@@ -187,7 +187,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     }
     const chunk = this.currentBuffer()[idx];
     if (!(chunk.type === 'REGULAR')) {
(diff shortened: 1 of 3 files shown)


packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -290,6 +308,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
 
   visitReturnClause = (ctx: ReturnClauseContext) => {
     this.visit(ctx.RETURN());
+    this.avoidBreakBetween();
```
tomasnyberg, line 311: can you add this for all clauses? Currently the tests pass just because the memoization happens to avoid this case for all clauses, but it's not technically correct. If you add this to all clauses it should make them all correct

tomasnyberg: thanks 🔥

tomasnyberg (approved):
LGTM

PR #375, opened 2025-02-27
Removes groups and indentation chunks from chunk list in formatter

simonthuresson:
### Description

This PR resolves a technical debt issue identified in the [main V2 PR](https://github.com/neo4j/cypher-language-support/pull/369#:~:text=(tech%20debt)%20Splits%20can%20happen%20between%20%22special%22%20chunks%2C%20which%20leads%20to%20weird%20cases%20like%20trailing%20spaces%20at%20the%20end%20of%20lines).

**Problem:**  
Previously, group start, end, add indentation and remove indentation markers were part of the chunk list, causing unnecessary choices to be made. This led to unexpected behavior such as trailing spaces at the end of lines.

**Solution:**  
Markers such as group start, end, increase indentation and remove indentation are now contained within a chunk rather than being standalone entries in the chunk list:
- When a group start is encountered, the a general group start counter is incremented
- Once a regular chunk is added, the groupStart property of the chunk will be what group counter is. The group counter is then set to zero.  
- Group end markers are directly incremented on previous chunk.
- Add/remove indentation is directly changed on the previous chunk once encountered. 

This approach prevents splits from occurring between "special" chunks, resolving the underlying issue.


tomasnyberg:
Very nice 🔥  just had a few nits about some naming things etc but great overall. This is very important for other bugfixes so will be super helpful

packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -189,19 +187,31 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    this.currentBuffer().at(-1).group.push(groupEndChunk);
   };
 
-  endGroup = () => {
-    this.currentBuffer().push(groupEndChunk);
+  addGroups = (): GroupChunk[] => {
+    if (this.currentBuffer().length === 0) {
+      return [];
+    }
+    const groupChunk: GroupChunk[] = [];
```
tomasnyberg, line 209: this does not convey that it is a list

tomasnyberg (approved):
Looked through the changes since I last reviewed and it looks great! As soon as you get the tests to pass this should be merged asap 🔥 

nit: This reverts the changes we made to CASE expressions, can that be documented in the description as well?

packages/language-support/src/tests/formatting/formattingv2.test.ts
```diff
@@ -644,16 +646,17 @@ WHERE b.name = "XGyhUMQO"
 RETURN u, r, b, c`;
     verifyFormatting(query, expected);
   });
-
+  // TODO
```
OskarDamkjaer, line 649: Is this still applicable? If so, i'm thinking it make sense to expand on it here or just keep a card in trello

OskarDamkjaer (approved):
It'd be nice with a test showing the "trailing spaces at the end of lines" and other similar problems that get solved from this change

tomasnyberg:
> It'd be nice with a test showing the "trailing spaces at the end of lines" and other similar problems that get solved from this change

It's not really possible to test it since we would need to explicitly break the `decisionsToFormatted` function for it not to work. This PR allows us to remove some other bugs which wouldn't otherwise be possible which IMO is the real test

packages/language-support/src/tests/formatting/formattingv2.test.ts
tomasnyberg: We have a card for fixing the CASE statements, I'll add a point to remove this there

tomasnyberg: And I think it's fine to keep it there for now since CASE statements are not working fully as intended

PR #376, opened 2025-02-28
Adds .cy, .cyp and .cql extensions to the VSCode extension

### LINKED ISSUE #367: Add support for .cy file extension for vscode
Hi all,

I've just recently migrated from the old Neo4j extension to the new extension (Neo4j for VS Code). For the old extension, both `.cy` and `.cypher` files would be recognized. However, that no longer seems to be the case, and I've named all my queries in `.cy` format. It is difficult to migrate every saved query to `.cypher`, so please add back the `.cy` file extension support.

ncordon:
Adds `.cql, .cy, .cyp` as possible extensions for Cypher files in the VSCode plugin.

Closes #367

### DIFF
--- a/packages/vscode-extension/CHANGELOG.md
+++ b/packages/vscode-extension/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Adds autocompletions for Cypher version
 - Adds support for CYPHER preparser options (runtime, planner, etc)
 - Adds support for multiple Neo4j connections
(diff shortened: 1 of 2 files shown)


anderson4j (changes requested):
Same comment as on the other PR, 1.7.0 is the already released version, other than that nice to match files supported by the old extension :)

anderson4j (approved):
Looks like something went wrong when I tried to release 1.7.0 a long time ago, so current can be 1.7.0

PR #377, opened 2025-02-28
Fixes most bugs related to comments in V2 formatter

tomasnyberg:
## Description
Comments are not handled very well in the current version of the cypher (v2) formatter, and it frequently panics when encountering them. This PR _significantly_ improves its ability to handle comments. For instance, the "allTokenTypes" example in the playground currently gets formatted like this: 
```
MATCH (variable:Label)-[:REL_TYPE]->()
WHERE variable.property = "String" OR namespaced.function() = false
// comment
                                      OR $parameter > 2
RETURN variable;
```
but with the introduction of this PR, it will get formatted like this:
```
MATCH (variable:Label)-[:REL_TYPE]->()
WHERE variable.property = "String" OR namespaced.function() = false
      // comment
      OR $parameter > 2
RETURN variable;
```
### High-level overview of changes
- When we encounter a "hard-break" comment, i.e. one that has a newline before it, we end all currently active groups except for the last one.
    - We keep the last one since we want to maintain indentation for clauses (like the "// comment" and OR expr aligning with the WHERE above)
- When we start groups, we now keep track of their IDs, and maintain a stack of currently "active" groups. This ID is used when ending groups, such that we only end a group if the supplied id matches the one currently on the top of the stack. In the case that a comment has already ended the group we intended to end like above, we simply do nothing, since the group is already gone.
    - This avoids groups "wrapping" hard-break comments, which caused a lot of the weird behavior from before.
    - This comes with the added benefit of being able to name our groups (through the variable name used when saving the group id).
 - We no longer consider comments that go above the 80 character limit to be overflowing. This seems consistent with other formatters, which allow you to have long comments without complaining. In most cases when a comment overflows, the user wants the comment to be exactly there anyways. If they don't like the fact that it overflows, they can move the comment to e.g. the line before.
- Removes unnecessary code in the decisionsToFormatted function
   - Doesn't technically have anything to do with this PR but it came up in a merge conflict because I was adding some additional hacks to it. The PR which removes groups from the decision tree makes all the hacks unnecessary.

## Testing
- Adds 20 or so tests with line breaks and comments, and re-adds the TODO-commented one we have had for a while
- Ran the verificiation check on 100k queries. Out of these, all but 2 passed both the integrity and idempotency check. The ones that didn't aren't broken because of comments (we're tracking the issues that caused them to break).


packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -178,20 +180,37 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    while (idx >= 0 && this.currentBuffer()[idx].type === 'COMMENT') {
+      idx--;
+    }
+    return idx;
   };
 
-  endGroup = () => {
-    this.currentBuffer().at(-1).groupsEnding += 1;
+  endGroup = (id: number) => {
+    if (this.groupStack.length === 0 || this.groupStack.at(-1) !== id) {
```
simonthuresson, line 192: Is it possible for a groupstack to have length 0 when endGroup comes?

tomasnyberg: I tried removing it and the tests still pass. Assuming no incorrect endGroup() calls, I would assume no? Since we require an explicit id I think the odds of introducing an incorrect endGroup() are low enough to warrant removing it.

Nice catch!

packages/language-support/src/formatting/formattingSolutionSearchv2.ts
```diff
@@ -92,8 +92,11 @@ function getIndentations(curr: State, choice: Choice): [number, number] {
     finalIndent = curr.activeGroups.at(-1).align;
   }
 
-  if (choice.left.type === 'COMMENT') {
-    finalIndent = curr.column === 0 ? nextBaseIndent : 0;
+  // Align hard-break comments with the outermost group (usually the one that
+  // aligns things with a clause)
+  if (choice.left.type === 'COMMENT' && choice.left.breakBefore) {
+    const lastGroup = curr.activeGroups.at(0);
```
simonthuresson, line 98: Can activeGroups be empty at this point?

tomasnyberg: Yes, but in that case .at(0) returns undefined. The next line checks if it's undefined and picks `nextBaseIndent` in that case instead.

packages/language-support/src/formatting/formattingHelpersv2.ts
```diff
@@ -79,19 +79,24 @@ const traillingCharacters = [
   CypherCmdLexer.RBRACKET,
 ];
 
+// TODO: This function should probably not exist; we're not really fans of
+// shuffling around the AST like we're doing right now...
```
ncordon, line 83: Not related to the PR but just for your own understanding: we don't have an AST, we have a parsing tree (CST)

packages/language-support/src/tests/formatting/formattingv2.test.ts
```diff
@@ -1206,3 +1208,346 @@ RETURN dmk`.trim();
...
+  });
+
+  test('should not put the arrow on a newline for no reason', () => {
+    const query = `match (z:Consumer {zen_id: "T1M3wiuA"})-[:HAS_INTERACTION]-(i)
+MATCH (i:Interaction:PageView)-[:HAS_TARGET]->(t)
+//MATCH (t:Target)-[:BELONGS_TO]->(s)
+return z`;
+    const expected = `MATCH (z:Consumer {zen_id: "T1M3wiuA"})-[:HAS_INTERACTION]-(i)
+MATCH (i:Interaction:PageView)-[:HAS_TARGET]->(t)
+//MATCH (t:Target)-[:BELONGS_TO]->(s)
```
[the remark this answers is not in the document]
ncordon: What you have already is fine IMO. This is another option:

```ts
const expected = query;
verifyFormatting(query, expected)
```

packages/language-support/src/tests/formatting/formattingv2.test.ts
```diff
@@ -1206,3 +1208,346 @@ RETURN dmk`.trim();
...
+    const query = `MATCH (p:Product)
+WHERE p.price > 100 // price threshold for premium items
+      AND p.stock < 50 // low stock warning
+          OR
+      p.discount > 0 // consider discounted products even if stock is high
+RETURN p.name, p.price, p.stock, p.discount;`;
+    const expected = `MATCH (p:Product)
+WHERE p.price > 100 // price threshold for premium items
+      AND p.stock < 50 // low stock warning
+      OR p.discount > 0 // consider discounted products even if stock is high
```
ncordon, line 1395: really nice

ncordon (approved):
LGTM

PR #382, opened 2025-03-03
Adds preservation of explicit newlines made by user to formatter

tomasnyberg:
## Description

Multiple people who have tried out the V1 formatter commented on the fact that it does not preserve the explicit newlines they have placed in their queries. So, for instance, this query
```
MATCH (n)

// Return the node
RETURN n
```
would get formatted to this:

```
MATCH (n)
// Return the node
RETURN n
```
Which can be quite annoying if you've tried to separate some things logically by inserting that newline. Prettier also preserves these explicit newlines, e.g. if you have this 

```
 const traillingCharacters = [
  CypherCmdLexer.SEMICOLON,

  CypherCmdLexer.COMMA,
  CypherCmdLexer.COLON,
  CypherCmdLexer.RPAREN,
  CypherCmdLexer.RBRACKET,
];
```
it will not remove the newline after the first list item. If there are two or more newlines however, they are reduced to just one.

This PR adds support for preserving these explicit newlines, by checking for more than one \n in these three places

1. After the end of a clause
2. Before a comment
3. After a comment

If two or more are found, we set the chunk to "double break", meaning its split choices will now contain \n\n instead of \n.

### Testing 

- ~13 new tests for these cases, had to fix some old tests as well
- Ran the verify check on 100k queries and all but two passed, which is the same result as on main. There does seem to be a bug with comments where they get duplicated or disappear however. We are tracking that on trello and will fix it in a separate PR since it's not related to this.

### DIFF
--- a/packages/language-support/src/formatting/formattingSolutionSearchv2.ts
+++ b/packages/language-support/src/formatting/formattingSolutionSearchv2.ts
@@ -73,6 +73,34 @@ interface FinalResultWithPos {
 type FinalResult = string | FinalResultWithPos;
 
 const openingCharacters = [CypherCmdLexer.LPAREN, CypherCmdLexer.LBRACKET];
+const standardSplits: Split[] = [
+  { splitType: ' ', cost: 0 },
+  { splitType: '\n', cost: 1 },
+];
+const doubleBreakStandardSplits: Split[] = [
+  { splitType: ' ', cost: 0 },
+  { splitType: '\n\n', cost: 1 },
+];
+const noSpaceSplits: Split[] = [
+  { splitType: '', cost: 0 },
+  { splitType: '\n', cost: 1 },
+];
+const noSpaceDoubleBreakSplits: Split[] = [
+  { splitType: '', cost: 0 },
+  { splitType: '\n\n', cost: 1 },
+];
+const noBreakSplit: Split[] = [{ splitType: ' ', cost: 0 }];
+const noSpaceNoBreakSplit: Split[] = [{ splitType: '', cost: 0 }];
+const onlyBreakSplit: Split[] = [{ splitType: '\n', cost: 0 }];
+const onlyDoubleBreakSplit: Split[] = [{ splitType: '\n\n', cost: 0 }];
+
+const emptyChunk: RegularChunk = {
+  type: 'REGULAR',
+  text: '',
+  groupsStarting: 0,
+  groupsEnding: 0,
+  modifyIndentation: 0,
+};
 
 export function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
   return (
@@ -356,32 +384,3 @@ export function buffersToFormattedString(
   }
   return { formattedString: formatted.trimEnd(), cursorPos: cursorPos };
 }
-
-const standardSplits: Split[] = [
-  { splitType: ' ', cost: 0 },
-  { splitType: '\n', cost: 1 },
-];
-const doubleBreakStandardSplits: Split[] = [
-  { splitType: ' ', cost: 0 },
-  { splitType: '\n\n', cost: 1 },
-];
-const noSpaceSplits: Split[] = [
-  { splitType: '', cost: 0 },
-  { splitType: '\n', cost: 1 },
-];
-const noSpaceDoubleBreakSplits: Split[] = [
-  { splitType: '', cost: 0 },
-  { splitType: '\n\n', cost: 1 },
-];
-const noBreakSplit: Split[] = [{ splitType: ' ', cost: 0 }];
-const noSpaceNoBreakSplit: Split[] = [{ splitType: '', cost: 0 }];
-const onlyBreakSplit: Split[] = [{ splitType: '\n', cost: 0 }];
-const onlyDoubleBreakSplit: Split[] = [{ splitType: '\n\n', cost: 0 }];
-
-const emptyChunk: RegularChunk = {
-  type: 'REGULAR',
-  text: '',
-  groupsStarting: 0,
-  groupsEnding: 0,
-  modifyIndentation: 0,
-};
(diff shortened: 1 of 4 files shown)


packages/language-support/src/formatting/formattingSolutionSearchv2.ts
```diff
@@ -74,6 +74,35 @@ type FinalResult = string | FinalResultWithPos;
 
 const openingCharacters = [CypherCmdLexer.LPAREN, CypherCmdLexer.LBRACKET];
 
+const standardSplits: Split[] = [
```
tomasnyberg, line 77: Took the opportunity to move these consts to the top since I think that makes more sense. Not sure why I put them at the bottom to begin with.

packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -243,11 +292,22 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
       token.tokenIndex,
     );
-    const commentTokens = (hiddenTokens || []).filter((token) =>
-      isComment(token),
-    );
     const nodeLine = node.symbol.line;
-    for (const commentToken of commentTokens) {
+    let breakCount = 0;
+    let includesComment = false;
+    for (const hiddenToken of hiddenTokens || []) {
```
simonthuresson, line 298: Why do you need the empty array?

simonthuresson: Right, getHiddenTokensToRights return type is Token[], but in practice it can be null

tomasnyberg:
We noticed newlines aren't preserved between MATCH and WHERE for some reason:

```
MATCH (n:Person)


WHERE n.name = "Steve"

RETURN n
LIMIT 12;
```
turns into 

```
MATCH (n:Person)
WHERE n.name = "Steve"

RETURN n
LIMIT 12;
```

And I'm not entirely sure why. I'll sort this out before merging and add a test for it.

packages/language-support/src/tests/formatting/formattingv2.test.ts
```diff
@@ -1551,3 +1547,476 @@ LIMIT "6pkMe6Kx"`;
...
+    verifyFormatting(query, expected);
+  });
+
+  test('this comment should leave a newline before it', () => {
+    const query = `
+MATCH (n)
+
+// Comment
+RETURN n`.trimStart();
+    const expected = query;
```
simonthuresson, line 1791: Nice

tomasnyberg:
> We noticed newlines aren't preserved between MATCH and WHERE for some reason:
> 
> ```
> MATCH (n:Person)
> 
> 
> WHERE n.name = "Steve"
> 
> RETURN n
> LIMIT 12;
> ```
> 
> turns into
> 
> ```
> MATCH (n:Person)
> WHERE n.name = "Steve"
> 
> RETURN n
> LIMIT 12;
> ```
> 
> And I'm not entirely sure why. I'll sort this out before merging and add a test for it.


Fixed now, it had to do with the `WHERE` not being a real clause. I included breaking before `LIMIT` as well.

packages/language-support/src/formatting/formattingv2.ts
tomasnyberg: Yeah not sure why it can return null 🤯  but all tests fail if we don't do this

packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -213,6 +229,53 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+  preserveExplicitNewline = (
+    ctx: ParserRuleContext,
+    side: 'before' | 'after',
+  ) => {
+    const bottomChild = this.findBottomChild(ctx, side);
+    const token = bottomChild.symbol;
+    const hiddenTokens =
+      side === 'before'
+        ? this.tokenStream.getHiddenTokensToLeft(token.tokenIndex)
+        : this.tokenStream.getHiddenTokensToRight(token.tokenIndex);
```
ncordon, line 266: Just to make sure, this `getHiddenTokensToRight` only returns hidden tokens to the right of a given one until it finds one that is not in the hidden channel?

packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -269,6 +343,9 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       }
       this.currentBuffer().push(chunk);
     }
+    if (breakCount > 1 && includesComment) {
+      this.doubleBreakBetween();
+    }
```
ncordon, line 348: When is this condition supposed to kick in? I've put a breakpoint inside this if because I don't understand its purpose and from the tests we have none hit it

packages/language-support/src/formatting/formattingv2.ts
tomasnyberg: Exactly

packages/language-support/src/formatting/formattingv2.ts
tomasnyberg: I'm not sure why the breakpoint did not hit for you, perhaps your VSCode acting up? If I comment it out 4 tests fail.

It's a check for if the last comment has two or more newlines after it, since then we won't add the explicit newline in the for loop (the loop only adds explicit newlines when it sees a  comment). So if we have 

`[C, \n, \n]` as the hidden tokens array, we would forget about adding the newline after C without it. 

An example would be this query 
```
  MATCH (n)
  // Comment

  RETURN n
```
The hidden tokens array looks exactly like I've described above, and so without the check we would just get  
```
  MATCH (n)
  // Comment
  RETURN n
```

It's definitely a bit unclear what its purpose is when reading it though, it took me a moment to remember as well (even though I wrote it yesterday XD). I've added a comment for context.

ncordon: > I'm not sure why the breakpoint did not hit for you, perhaps your VSCode acting up? If I comment it out 4 tests fail.

Because I wasn't trying with the right suite 🙈 

Thanks for the explanation!

ncordon (approved):

PR #386, opened 2025-03-05
Adds completions to yield of show commands

anderson4j:
Before:
```
SHOW SETTINGS YIELD ^
SHOW DATABASE YIELD ^
```
Gave no completions.
Now, we have a list of yield-items for these, resulting in:
```
SHOW SETTINGS YIELD ^ <- [defaultValue, description, ...]
SHOW DATABASE YIELD ^ <- [access, address, aliases, ...]
```
This approach would have problems with old/new versions of databases if these lists are version-dependent (I suppose mostly things should remain, but for example SHOW DATABASES will get "defaultLanguage" soon). I got mine from a 5.26 db iirc. Also haven't gotten around to adding tests yet, but looks to work as expected in playground. Will proceed with connections stuff for now.

Alternative easy approach - polling. Might be slow. Alternative approach that would be quicker but probably slow to implement - create map<db-version, yieldItemArray> (I suppose we could have something like "only add an entry when the version is updated" and check what interval a given version is). Doing manually would be a little annoying, but maybe it could be scripted?


ncordon:
I'll close this, I'll add it to the Trello card so we keep track of it

PR #388, opened 2025-03-05
Fixes function invocation alignment in formatter

tomasnyberg:
## Description
As mentioned in PR #384, there is a bug with function invocation alignment where function arguments do not align with the opening `(`, and instead align with the namespace or the function name. This PR fixes the issue described, meaning that for instance this query goes from being formatted like this:
```
RETURN collect(create_this1 {datetime: apoc.date.convertFormat(
               toString(create_this1.datetime), "OZQvXyoU", "EhpkDy8g")})
       AS data
```
to being formatted like this:

```
RETURN collect(create_this1 {datetime:
               apoc.date.convertFormat(toString(create_this1.datetime),
                                       "OZQvXyoU", "EhpkDy8g")}) AS data
```
which is more consistent with how we usually format things.

### Testing
- Three new tests which ensure that the best solution for very tricky namespaced functions is found.
- Fixes the test described above

### DIFF
--- a/packages/language-support/src/formatting/formattingSolutionSearchv2.ts
+++ b/packages/language-support/src/formatting/formattingSolutionSearchv2.ts
@@ -188,7 +188,9 @@ function reconstructBestPath(state: State): Result {
 }
 
 function getStateKey(state: State): string {
-  return `${state.column}-${state.choiceIndex}`;
+  return `${state.column}-${state.choiceIndex}-${
+    state.activeGroups.at(-1)?.align
+  }`;
 }
 
 function bestFirstSolnSearch(
--- a/packages/language-support/src/formatting/formattingv2.ts
+++ b/packages/language-support/src/formatting/formattingv2.ts
@@ -926,9 +926,9 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     this.endGroup(functionNameGrp);
     this.avoidSpaceBetween();
     this.avoidBreakBetween();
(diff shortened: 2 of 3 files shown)


packages/language-support/src/tests/formatting/edgecasesv2.test.ts
```diff
@@ -309,6 +309,59 @@ YIELD *`;
...
+RETURN userAccountInfo;`;
+    const expected = query;
+    verifyFormatting(query, expected);
+  });
+
+  test('aligns and breaks long namespaced functions well 3', () => {
+    const query = `MATCH (inventoryRecord:ProductInventoryTrackingInformation)
+WITH inventoryRecord,
+     apoc.util.
+     validate(inventoryRecord.currentStock
```
ncordon, line 349: Uhhh, interesting, I think we should keep the name of the function consistent in one line? Probably not related to this pr

ncordon (approved):

packages/language-support/src/tests/formatting/edgecasesv2.test.ts
tomasnyberg: Do you mean not allowing a split on the dots like above? 

It leads to some weird results if the namespace/ function name is too long, e.g.

`apoc.util.validateResults(`

or some even longer ones. If we then have an alignment of e.g. 10 spaces before it, we now have to format everything within it in the remaining 40 or so characters, leaving a ton of whitespace

PR #389, opened 2025-03-06
Improves CASE statements

simonthuresson:
### Description
This PR overhauls how CASE statements are formatted. The previous rules always did a line break before CASE. This did work okey but for example a CASE statement inside an expression would be put on column 0:
```
WITH s, t.name AS tableName, collect({name: c.name, pk:
CASE
  WHEN true  THEN "(PK)"
  ELSE ""
END
```

Now instead, CASE always has an indentation to better showcase it is part of the previous expression:
```
WITH s, t.name AS tableName, collect({name: c.name, pk:
  CASE 
    WHEN true  THEN "(PK)"
    ELSE ""
  END
``` 

### Note
We have discussed quite a lot if CASE should break and align with the outer group. This does sometimes better show what CASE is a part of. However, it does also introduce "unnecessary" empty space. Previous example would look like this:
```
WITH s, t.name AS tableName, collect({name: c.name, pk:
     CASE 
       WHEN true  THEN "(PK)"
       ELSE ""
     END
``` 
Which works fine, but another example would be:
```
MATCH (n:Person)
RETURN
       CASE
         WHEN n.eyes = 'blue' THEN 1
         WHEN n.age < 40 THEN 2
         ELSE 3
       END AS result, n.eyes, n.age
```
In which we think there is definitely too much empty space, therefore we settled on an in between which is always having one indentation.


packages/language-support/src/formatting/formattingHelpersv2.ts
```diff
@@ -53,6 +53,7 @@ export interface BaseChunk {
   groupsStarting: number;
   groupsEnding: number;
   modifyIndentation: number;
+  specialIndentation: number;
```
tomasnyberg, line 56: Could this be renamed? `special` does not really convey what it's doing. Perhaps just `caseIndentation` since `CASE` is the only construct that actually uses it

packages/language-support/src/formatting/formattingSolutionSearchv2.ts
```diff
@@ -382,7 +422,7 @@ export function buffersToFormattedString(
       formatted += formattingResult.formattedString + '\n';
     }
   }
-  if (indentation > 0) {
+  if (indentation !== 0 || specialIndentation !== 0) {
```
tomasnyberg, line 425: nice catch 👍

packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -213,6 +223,10 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     this.groupStack.pop();
   };
 
+  removeGroup = () => {
```
tomasnyberg, line 226: I think we count starting groups as well. What happens to those? The `startGroupCounter` I mean.

packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -229,6 +243,18 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     this.currentBuffer().at(idx).modifyIndentation -= 1;
   };
 
+  mustBreakBetween = () => {
+    this.setAvoidProperty('mustBreak');
```
tomasnyberg, line 247: nit: `setAvoidProperty` is kind of a bad name. Not the fault of this PR but could perhaps be changed with it. Maybe `setChunkSplitProperty`

packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -949,51 +980,82 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
 
   // Handled separately since cases want newlines
   visitCaseExpression = (ctx: CaseExpressionContext) => {
-    this.breakLine();
+    while (this.groupStack.length > 1) {
```
tomasnyberg, line 983: Could this be a method instead?

packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -949,51 +980,82 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+      this.endGroup(this.groupStack.at(-1));
+    }
+    this.addSpecialIndentation();
+    this.mustBreakBetween();
     this.visit(ctx.CASE());
-    const caseGrp = this.startGroup();
+    for (let i = 0; i < this.currentBuffer().at(-1).groupsStarting; i++) {
```
tomasnyberg, line 989: same thing here

packages/language-support/src/tests/formatting/edgecasesv2.test.ts
```diff
@@ -156,21 +156,127 @@ WHERE b.name = "XGyhUMQO"
...
-END
+  CASE
+    WHEN n["asdf"] IS STRING THEN n.prop
+    ELSE 'default'
+  END
 RETURN n`;
     verifyFormatting(query, expected);
   });
 
+  test('multiple case statments after each other', () => {
```
tomasnyberg, line 174: Can you add a test where there are nested CASE expressions? It would be good to verify that the specialIndentation accumulates

tomasnyberg (commented):
💯

packages/language-support/src/formatting/formattingHelpersv2.ts
simonthuresson: SpecialIndentation might not be the perfect name. Though it is for special cases of where indentation overrides groups? Also it might be used later, im thinking FOREACH. Sure it is only used for CASE right now, but could perhaps be used somewhere else later?

tomasnyberg: Sure, I'm struggling to come up with a good alternative as well. I suppose it really is `special` in the sense that it overrides other indentaiton / alignment

packages/language-support/src/formatting/formattingv2.ts
simonthuresson: When we do this.visit(ctx.CASE()) it takes up all previous started groups and sets startGroupCounter to zero. Therefore we only need to modify the CASE chunk, because CASE has already "eaten" the previous group starts

packages/language-support/src/formatting/formattingv2.ts
simonthuresson: Yepp, I like setChunkSplitProperty better

packages/language-support/src/tests/formatting/edgecasesv2.test.ts
simonthuresson: Yepp, added, make sense

packages/language-support/src/formatting/formattingHelpersv2.ts
OskarDamkjaer: overridingIntentation perhaps?

packages/language-support/src/formatting/formattingv2.ts
```diff
@@ -154,13 +156,22 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    propertyName: 'noSpace' | 'noBreak' | 'mustBreak',
+  ): void => {
     let idx = this.currentBuffer().length - 1;
 
     while (idx >= 0 && this.currentBuffer()[idx].type === 'COMMENT') {
       idx--;
     }
-
+    // After a comment we always hardBreak, so if the loop finds a comment
+    // we should not also set mustBreak, it will cause breaking twice
```
OskarDamkjaer, line 168: I may be out bicycling here, but would it be possible to re-use the `mustBreak` concept for the comment as well (instead of the current mechanism) to avoid this check?

packages/language-support/src/formatting/formattingHelpersv2.ts
OskarDamkjaer: or maybe extraGroupIndent perhaps?

OskarDamkjaer (approved):

packages/language-support/src/formatting/formattingHelpersv2.ts
tomasnyberg: > `overridingIntentation` 

This one's okay but I'm not sure it helps significantly in terms of understanding

> `extraGroupIndent`

It doesn't really add indentation to groups

We're planning on encapsulating the indentation rules in a more sensible manner since right now it's a bit all over the place (as a result of the rules being all over the place). I think the only real solution is encapsulating it all in one or two methods and adding tons of comments. Indentation has to be decided based on three (maybe even more going forward) different things, and I don't think good variable names will be enough to explain it all no matter what.

packages/language-support/src/formatting/formattingv2.ts
tomasnyberg: Comments implicitly have a `mustBreak` with how we handle them currently (see the `determineSplits` function in the solution search). If we attached the mustBreak to a comment also in this funciton we would have to have an `if` for that instead, so no real difference to what we're doing here. 

There's an argument to be made that perhaps they should be different methods since mustBreak doesn't necessarily need to avoid comments like the other two properties, but I think it's fine as is unless we e.g. start allowing inline comments without a break after them

tomasnyberg (approved):

ncordon (approved):

PR #390, opened 2025-03-06
Adds parameter pane to VSCode extension

ncordon:
Adds a parameter pane to the VSCode extension:

https://github.com/user-attachments/assets/51ddfc53-031f-4e55-aacf-d0f48d2bc5d3

Right now those parameters can only be added individually or collectively cleaned through the context menu or the VSCode commands:

## How

The parameters are stored in the VSCode global store (similar to the connections) so they are available after restarting the editor / extension. They also need to be transmitted to the language server on startup or on parameter modification.

When we set a parameter, we are prompted for the key and separately for the value. If the key includes any non-alpha numerical characters, it will get escaped when evaluated. The value is anything that could be used within a `RETURN` statement:

```
RETURN datetime() // We can complete value with datetime()
RETURN 1 + 5 - 4 // We can complete this in the value
```

What we input in the value box gets evaluated against neo4j (therefore **we need an active neo4j connection to set parameters**). We then send it to the driver with a `RETURN ${value} as param` query, we get the driver object back, we serialize it (to be able to store it) and any time we need to use those parameters, we deserialize the stored value. The code to do that has been lifted from upx and the plan is to eventually move the whole neo4j-sdk here, so those plans align well with just using that code.

## Testing

There are also a few rough edges that should be ironed out but bear in mind this is a first iteration. Namely:

* Better size for the icons (already talking to the design team about that.
* Better error messages
* More tests: I've added a few e2e tests, but this PR is already really big, so I will add more low level (mock) tests in a separate PR.


packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -331,14 +332,33 @@ const parameterCompletions = (
...
+      let maybeInsertText = backtickedName
+        ? { insertText: `$${backtickedName}` }
+        : {};
+      if (previousToken) {
+        const suffix = maybeInsertText.insertText ?? `$${paramName}`;
+        maybeInsertText = {
+          // We need to complete parameters correctly in VSCode,
+          // otherwise when we have 'RETURN $' and we get offered $param
+          // we would complete RETURN $$param, which is not what we want
+          insertText: computeSuffix(previousToken.text, suffix),
```
ncordon, line 344: Probably this needs tests in the language support side and the react codemirror side

packages/language-server/src/server.ts
```diff
@@ -132,6 +135,11 @@ connection.onNotification(
   },
 );
 
+connection.onNotification('updateParameters', (parameters: Neo4jParameters) => {
+  neo4jSchemaPoller.setParameters(parameters);
+  relintAllDocuments();
+});
+
```
ncordon, line 142: When we set the parameters in the UI layer, we want them to be available for the language server to complete

packages/language-support/src/tests/autocompletion/parameterCompletions.test.ts
```diff
@@ -19,9 +19,21 @@ describe('Completes parameters outside of databases, roles, user names', () => {
...
+        {
+          label: '$intParam',
+          insertText: 'intParam',
+          kind: CompletionItemKind.Variable,
+        },
+        {
+          label: '$mapParam',
+          insertText: 'mapParam',
+          kind: CompletionItemKind.Variable,
+        },
```
ncordon, line 36: This part tests the `insertText`

packages/schema-poller/src/cypher-execution/query-result.ts
```diff
@@ -0,0 +1,348 @@
+import type { QueryResult, Record as Neo4jRecord } from 'neo4j-driver';
```
ncordon, line 1: This file is copied from upx

packages/schema-poller/src/data-transforms/cypher-type-names.ts
```diff
@@ -0,0 +1,147 @@
...
+  if (isRelationship(val)) {
+    return 'Relationship';
+  }
+
+  if (isPath(val)) {
+    return 'Path';
+  }
+
+  // TODO Nacho: this was `Unknown` in the upx code. What should we do here?
+  return 'Map';
```
ncordon, line 146: This file is copied from upx except for this last line which I've modified so we can get in VSCode the type of a `Map` shown right

packages/schema-poller/src/data-transforms/record-to-string.ts
```diff
@@ -0,0 +1,254 @@
+import type { Node, Path, Point, Relationship } from 'neo4j-driver-core';
```
ncordon, line 1: Also copied from upx, probably some parts of this file are not needed

packages/schema-poller/src/data-types/cypher-data-types.ts
```diff
@@ -0,0 +1,60 @@
+import type {
```
ncordon, line 1: File copied from upx

packages/vscode-extension/src/commandHandlers/connection.ts
```diff
@@ -216,3 +217,20 @@ export async function cypherFileFromSelection(): Promise<void> {
...
+
+export async function forceConnect(i: number): Promise<void> {
+  const connections = getConnections();
+  const connectionToToggle = Object.values(connections).at(i);
+  if (connectionToToggle) {
+    const { result, connection } =
+      await toggleConnectionAndUpdateDatabaseConnection(connectionToToggle);
+    displayMessageForConnectionResult(connection, result);
+  }
+}
```
ncordon, line 236: These two are for testing purposes

packages/vscode-extension/src/constants.ts
```diff
@@ -10,6 +10,13 @@ export const CONSTANTS = {
     SWITCH_DATABASE_COMMAND: 'neo4j.switchDatabase',
     RUN_CYPHER: 'neo4j.runCypher',
     CYPHER_FILE_FROM_SELECTION: 'neo4j.cypherFileFromSelection',
+    ADD_PARAMETER: 'neo4j.addParameter',
+    CLEAR_PARAMETERS: 'neo4j.clearParameters',
+    INTERNAL: {
+      EVAL_PARAMETER: 'neo4j.internal.evalParam',
+      FORCE_DISCONNECT: 'neo4j.internal.forceDisconnect',
+      FORCE_CONNECT: 'neo4j.internal.forceConnect',
+    },
```
ncordon, line 19: I've added a few commands that are internal, therefore they can only be used when having access to the vscode api (i.e. in tests)

packages/language-support/src/helpers.ts
```diff
@@ -71,12 +71,6 @@ export function findParent(
   return current;
 }
 
-export function getTokens(tokenStream: CommonTokenStream): Token[] {
```
ncordon, line 74: Just a cleanup

packages/schema-poller/src/data-transforms/cypher-type-names.ts
```diff
@@ -0,0 +1,146 @@
+import {
```
ncordon, line 1: I lifted this file from upx

packages/vscode-extension/src/connectionService.ts
```diff
@@ -503,7 +503,7 @@ async function disconnectAllDatabaseConnections(): Promise<void> {
  * Gets all Connections from the global state.
  * @returns A Connections object.
  */
-function getConnections(): Connections {
+export function getConnections(): Connections {
```
ncordon, line 506: Exports this for the `forceConnect` functionality that I need for the tests

anderson4j (commented):
Will look more at the code tomorrow, but from playing around with the UI I found two things:
1. In UI-creation of params we can assign names starting with symbols that arent allowed in browser/the grammar. (Might be annoying to change for not small inconsistency, so could be future paper-cut?)

2. This one's more annoying - When suggesting parameters for alias/db-names we get (valid) parameters twice. Example query: CREATE ALIAS x FOR DATABASE ^
Otherwise the UI felt natural and easy to use :)

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -738,11 +760,13 @@ export function completionCoreCompletion(
 type CompletionHelperArgs = {
   parsingResult: ParsedStatement;
   dbSchema: DbSchema;
+  previousToken?: Token;
   candidateRule: CandidateRule;
 };
 function completeAliasName({
   candidateRule,
   dbSchema,
+  previousToken,
```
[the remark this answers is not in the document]
ncordon: Good catch

packages/vscode-extension/tests/specs/webviews/connection.spec.ts
```diff
@@ -39,12 +39,18 @@ suite('Connection testing', () => {
   });
 
   test('should connect when selecting new connection', async function () {
+    if (os.platform() === 'darwin') {
```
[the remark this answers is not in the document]
ncordon: This test includes a `await clickOnContextMenuItem(connectionSection, 'Disconnect', 0);` and the problem is precisely that menu is native in MacOS, so we do need this

packages/vscode-extension/tests/specs/webviews/params.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+
+suite('Params panel testing', () => {
+  let workbench: Workbench;
+
+  before(async () => {
+    workbench = await browser.getWorkbench();
+  });
+
+  async function addParam() {
+    await browser.executeWorkbench((vscode) => {
```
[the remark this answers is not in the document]
ncordon: I feel it's because it will fail and we get a Remote command timeout (you can try yourself), so either we make this void or we make the call to it void, I've switched to the second approach now so it's clearer.

packages/vscode-extension/src/commandHandlers/params.ts
```diff
@@ -0,0 +1,103 @@
...
+  paramName: string,
+  paramValue: string,
+): Promise<void> {
+  const schemaPoller = getSchemaPoller();
+  const dbSchema = schemaPoller.metadata.dbSchema;
+
+  try {
+    const maybeEscapedName = backtickIfNeeded(paramName) ?? paramName;
+    const errors = lintCypherQuery(
+      `:param ${maybeEscapedName} => ${paramValue}`,
```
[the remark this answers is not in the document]
ncordon: I don't follow what you are suggesting here?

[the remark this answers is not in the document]
ncordon: Or are you saying `10` does not need escaping?

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -337,13 +337,12 @@ const parameterCompletions = (
...
-          // otherwise when we have 'RETURN $' and we get offered $param
-          // we would complete RETURN $$param, which is not what we want
-          insertText: computeSuffix(previousToken.text, suffix),
-        };
+        const param = maybeInsertText.insertText ?? `$${paramName}`;
+        const suffix = computeSuffix(previousToken.text, param);
+        // We need to complete parameters correctly in VSCode,
+        // otherwise when we have 'RETURN $' and we get offered $param
+        // we would complete RETURN $$param, which is not what we want
+        maybeInsertText = suffix ? { insertText: suffix } : {};
```
ncordon, line 345: Avoids having insertText if the token is not a prefix of the parameter

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -665,11 +666,20 @@ export function completionCoreCompletion(
       }
 
       if (ruleNumber === CypherParser.RULE_symbolicAliasName) {
-        return completeAliasName({ candidateRule, dbSchema, parsingResult });
+        return completeAliasName({
+          candidateRule,
+          dbSchema,
+          parsingResult,
+        });
```
ncordon, line 673: I don't know if the grammar changed (possibly) but this looks like it can never be a parameter

packages/language-support/src/tests/autocompletion/databasenameCompletion.test.ts
```diff
@@ -301,15 +301,15 @@ describe('Can complete database names', () => {
...
       ],
       excluded: [
         { label: 'db1', kind: CompletionItemKind.Value },
         { label: 'db2', kind: CompletionItemKind.Value },
         { label: 'movies', kind: CompletionItemKind.Value },
-        // do not suggest non-string parameters
+        // do not suggest parameters, it's clear at this point we were not after a parameter in the completions
         { label: '$param2', kind: CompletionItemKind.Variable },
         { label: '$param3', kind: CompletionItemKind.Variable },
+        { label: '$param1', kind: CompletionItemKind.Variable },
```
ncordon, line 312: The reason why this test has changed is because `ALTER ALIAS a.b.c.` is well within the symbolic alias name completion, so the completion engine doesn't try to complete with parameter rule because after all of the namespaces you can only have another namespace basically. 

I've tried other approaches and the code ends up super convoluted so I'm fine with that tradeoff in this case.

For cases like `ALTER ALIAS blablablah` we are still getting parameter completions as I would expect.

packages/vscode-extension/src/commandHandlers/params.ts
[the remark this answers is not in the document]
anderson4j: Well im not sure what is actually intended to be allowed, but in Browser (I launch it from Desktop) I can't use 10 as a parameter name, and in vscode/workspace I can. And as you say, it looks like it doesnt need escaping either, but gets it on completions anyway.
image
image

packages/vscode-extension/tests/specs/webviews/connection.spec.ts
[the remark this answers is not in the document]
anderson4j: Oh right, forgot I added the disconnecting in the end to "reset state"

anderson4j (approved):
nice! (approving since I dont think the $10 issue is with vscode, and if it is, we can do it separately)

PR #391, opened 2025-03-06
Fixes map projection alignment in formatter

tomasnyberg:
## Description
Map projections, unlike map literals, currently do not get a wrapping group. This means that splits wihtin them look a bit weird:

```
MATCH (p:Person {name: "Alice"})
RETURN p {.name, .age, .email, .phone, .address, .occupation, .nationality,
       .birthdate, .gender} AS personInfo
```
This PR adds a wrapping group within map projections, meaning they now get formatted like this:

```
MATCH (p:Person {name: "Alice"})
RETURN p {.name, .age, .email, .phone, .address, .occupation, .nationality,
          .birthdate, .gender} AS personInfo
```

### Testing

2 new tests added
- Note: One test fails right now due to an unrelated bug with function invocation alignment that will be resolved with #388


ncordon (approved):
lgtm

PR #393, opened 2025-03-06
Fetches version explicitly in VSCode to fix cypher 25 polling

ncordon:
This PR addresses a couple of bugs.

## First bug
First of all we were relying on `SHOW DATABASES` surfacing the `defaultLanguage` column, but the query we were using was `SHOW DATABASES` and not `SHOW DATABASES YIELD *`, so that column was not among the surfaced ones.

## Second bug
The thesis was that column would only appear when Cypher 25 support was enabled, which is not true, for `2025.02` that column is there but the value cannot be other than `CYPHER 5`.

Had that been enabled, it would have meant that for neo4j `2025.02` we would have seen polling errors in the debug console:
 

which is the second bug we are correcting.

We are now polling for the version explicitly from the server, relying on the fact that from `2025.03` onwards `CALL dbms.components()` should be returning `5.27.0-2025030` 🙈 .


packages/schema-poller/src/schemaPoller.ts
```diff
@@ -163,7 +164,20 @@ export class Neo4jSchemaPoller {
       database,
     );
 
-    this.metadata = new MetadataPoller(databases, this.connection, this.events);
+    const { query: versionQuery, queryConfig: versionQueryConfig } =
+      getVersion();
+    const { serverVersion } = await this.driver.executeQuery(
```
anderson4j, line 169: Since this is a property of the dbms we are connected to, maybe we should keep it as a field in connection?

packages/schema-poller/src/queries/version.ts
```diff
@@ -0,0 +1,29 @@
+import { resultTransformers } from 'neo4j-driver';
+import { ExecuteQueryArgs } from '../types/sdkTypes';
+
+/**
+ * List available databases in your dbms
+ * https://neo4j.com/docs/cypher-manual/current/administration/databases/#administration-databases-show-databases
+ */
```
anderson4j, line 7: looks like a copypaste leftover

anderson4j (changes requested):
Nice! This way of checking version looks less hacky then the old solution of "checking if feature x exists" :)

Had some small suggestions and found one comment that was copypasted. Also a question on updating version nbr

.github/workflows/ci.yaml
```diff
@@ -51,8 +51,7 @@ jobs:
       - name: Builds and run tests
         # Tests with the same flags that VSCode does
         run: |
-          npm test -- --filter=!antlr4 -- --passWithNoTests
-          npm test -- --filter=antlr4
+          npm test
```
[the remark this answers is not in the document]
ncordon: It was just because we were using a different test extension in the past (`jest`) and we didn't have the antlr4 tests in the VSCode menu so at some point we broke the VSCode menu

packages/vscode-extension/package.json
```diff
@@ -6,7 +6,7 @@
   "publisher": "neo4j-extensions",
   "author": "Neo4j Inc.",
   "license": "Apache-2.0",
-  "version": "1.7.0",
+  "version": "1.7.1",
```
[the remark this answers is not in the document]
ncordon: Probably both

packages/schema-poller/src/schemaPoller.ts
ncordon: We already have it as field in the metadata of the connection

packages/language-support/src/tests/version.test.ts
```diff
@@ -0,0 +1,20 @@
...
+    expect(cypher25Supported('5.27.0-2025030')).toBe(false);
+  });
+
+  test('cypher 25 should be supported 5.27.0-2025040 onwards', () => {
+    expect(cypher25Supported('5.27.0-2025040')).toBe(true);
+    expect(cypher25Supported('5.27.0-2025050')).toBe(true);
+    expect(cypher25Supported('5.27.0-2025060')).toBe(true);
+    expect(cypher25Supported('5.28')).toBe(true);
+    expect(cypher25Supported('5.28.0')).toBe(true);
+    expect(cypher25Supported('6.1.0')).toBe(true);
```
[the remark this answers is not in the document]
ncordon: 👍 yeah makes sense

packages/schema-poller/src/schemaPoller.ts
anderson4j: Yeah I was thinking of shifting it so the field was in the connection, kind of like in the cypher shell, where it's in the BoltStateHandler. But either way no strong opinions, so if you prefer this way lets keep it

anderson4j (approved):
Clarified my other comment, but up to you to agree or not. In general LGTM

PR #394, opened 2025-03-07
Connections tree improvements

anderson4j:
Updates the connection tree with icons like this:
image
Also adds expanding on left click of a connection, and automatic expansion of the connection node on connection as seen here:

https://github.com/user-attachments/assets/42d5cfee-5ce8-487a-90a3-396cbcdaa220


packages/vscode-extension/resources/images/Neo4jActiveDark.svg
anderson4j, line 1: The trello card/instruction from Greg mentions turning it blue on being active, but from the design folks I heard we couldnt mess with "official logo" stuff too much, so this was the best he found

packages/vscode-extension/src/treeviews/connectionTreeDataProvider.ts
```diff
@@ -155,16 +157,103 @@ export class ConnectionItem extends TreeItem {
...
+            path.join(
+              __dirname,
+              '..',
+              'resources',
+              'images',
+              'Neo4jInactiveDark.svg',
+            ),
+          ),
+        };
+        this.id = `${key}-${collapsibleState}`;
```
anderson4j, line 210: Updating the id to change when collapsibleState does (but key is the same) was needed to update the UI when changing only collapsibleState in the tree

packages/vscode-extension/src/treeviews/connectionTreeDataProvider.ts
```diff
@@ -155,16 +157,103 @@ export class ConnectionItem extends TreeItem {
     super(label, collapsibleState);
     this.tooltip = label;
     this.contextValue = this.type;
-
+    this.command = {
+      title: 'onClickConnect',
+      command: CONSTANTS.COMMANDS.CONNECT_COMMAND,
```
anderson4j, line 162: CONNECT_COMMAND actually toggles "connected" to the other state, so clicking when disconnected connects, but clicking again when connected disconnects. I quite like that so I didnt mess with it

ncordon: The feedback Greg gave as well as me is that we'd expect clicking the first time to actually connect, but the second time it should toggle / untoggle the dropdown for databases instead?

packages/vscode-extension/src/treeviews/connectionTreeDataProvider.ts
```diff
@@ -155,16 +157,103 @@ export class ConnectionItem extends TreeItem {
...
       case 'activeConnection':
+        this.iconPath = {
+          light: Uri.file(
+            path.join(
+              __dirname,
+              '..',
+              'resources',
+              'images',
+              'Neo4jActiveLight.svg',
+            ),
```
ncordon, line 175: Have you tried if the icons are visible in Windows?

You can generate a `.vsix` package doing:

```
cd packages/vscode-extension
npx vsce package
```

and then installing that VSIX manually when you click on the Marketplace icon in VSCode:

packages/vscode-extension/src/treeviews/connectionTreeDataProvider.ts
anderson4j: (resolving since this has been adjusted since feedback)

packages/vscode-extension/src/treeviews/connectionTreeDataProvider.ts
```diff
@@ -158,13 +160,106 @@ export class ConnectionItem extends TreeItem {
...
+            path.join(
+              __dirname,
+              '..',
+              'resources',
+              'images',
+              'Neo4jInactiveDark.svg',
+            ),
+          ),
+        };
+        this.id = `${key}${collapsibleState}`;
```
ncordon, line 206: Why do you need the collapsible state in the id?

anderson4j: Without changing this, it seems vscode thought the item hadnt changed when collapsibleState did, so even though the underlying object had changed values, the UI didnt update to auto-expand when connecting.

PR #397, opened 2025-03-07
Upgrades formatter from V1 to V2

tomasnyberg:
## Description
This PR removes the V1 formatter and replaces it with the new and improved V2 formatter. The different versions differ significantly in terms of features, and not all changes will be listed in this description. For a thorough overview of what has changed in the formatter, look through all commits related to formatting from #369 onwards. A quick overview of the new functionality would be:

1. Line breaks (!!!)
    - Any expression or clause that exceeds 80 characters is broken into multiple lines, such that the formatter tries to fit as much as possible on one line. See the [architecture overview](https://github.com/neo4j/cypher-language-support/blob/main/packages/language-support/src/formatting/architecture.md) for how it works, or the [planned iterations](https://docs.google.com/document/d/1V3dBj-2HUAvSnWpp2GNowmpnOrmaV3t2k5_Q_giLcqI/edit?tab=t.0#heading=h.6poegv7ypu63) doc for an overview of what this entails.
1.  Preservation of explicit newlines
    - If a user has made two newlines (`\n\n`) after a clause or a comment, they will not be removed by the formatter. More than one blank line is not allowed however.
1. Loads of small bugs have been sorted out, e.g. spacing in map projections.

### "New" tests
Note that there are a few "new" tests added in this PR; they are simply tests that were present in the test file for V1, but had not made it into any of the V2 test files (because we had forgotten to port them over). I ran a script (written by GPT) to check for any missing tests, and all the old ones from V1 should be included now.

### Experimental formatting removed
This PR removes the experimental formatting checkbox and export since there is no longer an "experimental" formatter with the introduction of this PR:

image

### Versioning
I've marked this PR with a minor version change bump rather than just a patch, since the output of the formatter changes quite significantly. While I think it's unlikely that anyone would be affected by it, it feels like the reasonable thing to do. I'm happy to change it to just a patch however.

## Testing
- `npm run test`
- built the package with `npm run build` and tried the formatting in the codemirror playground. It provides the V2 formatting by default.


packages/language-support/src/formatting/formatting.ts
```diff
@@ -3,92 +3,318 @@ import { default as CypherCmdLexer } from '../generated-parser/CypherCmdLexer';
...
+      ...(hasCursor && { isCursor: true }),
+    };
+    this.currentBuffer()[indices[1]] = chunk;
+  };
+
+  /**
+   * Sets that if the previous token should not choose between the argument ('noSpace' or 'noBreak')
+   * Skips any preceding comments or special chunks we did not expect.
+   */
+  setChunkSplitProperty = (
```
simonthuresson, line 156: Rename to setChunkProperty?

packages/language-support/src/formatting/formatting.ts
```diff
@@ -3,92 +3,318 @@ import { default as CypherCmdLexer } from '../generated-parser/CypherCmdLexer';
...
+    this.setChunkSplitProperty('noBreak');
+  };
+
+  mustBreakBetween = () => {
+    this.setChunkSplitProperty('mustBreak');
+  };
+
+  doubleBreakBetween = (): void => {
+    if (this.currentBuffer().length > 0) {
+      this.currentBuffer().at(-1).doubleBreak = true;
```
simonthuresson, line 197: We are using this.currentBuffer().at(-1) quite a lot, possibly introduce a function for this?

packages/language-support/src/formatting/formatting.ts
```diff
@@ -3,92 +3,318 @@ import { default as CypherCmdLexer } from '../generated-parser/CypherCmdLexer';
...
+
+  addSpecialIndentation = () => {
+    this.currentBuffer().at(-1).specialIndentation += 1;
+  };
+
+  removeSpecialIndentation = () => {
+    this.currentBuffer().at(-1).specialIndentation -= 1;
+  };
+
+  findBottomChild = (
```
simonthuresson, line 274: nit: getBottomChild?

OskarDamkjaer (approved):
I don't see any changes to the playground here. Does it still have the `use experimental` checkbox? Would be good to clear that out if not.  Edit:

>This PR does not remove the formatQueryExperimental export, nor does it remove the checkbox for it from the codemirror playground. My reasoning is that we are likely to soon be adding an experimental formatter once again, and it would be unnecessary to remove it only to add it back soon thereafter.

I see you've addressed this in the description already. However, I think it makes sense to remove the experimental toggle as you've been looking into doing the next version without a fully copy / paste. Ideally the v3 improvements would go straight into the normal one, without the need for an "experimental" tag

packages/language-support/src/formatting/formatting.ts
OskarDamkjaer: `this.lastInCurrentBuffer()` could perhaps read better

packages/language-support/src/formatting/formatting.ts
```diff
@@ -147,26 +403,161 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    this.visit(ctx.AS());
+    this.visit(ctx.variable());
+    this.endGroup(asGrp);
+    this.endGroup(unwindClauseGrp);
+  };
+
+  visitLimit = (ctx: LimitContext) => {
+    this.preserveExplicitNewlineBefore(ctx);
+    this.breakLine();
+    this.visitChildren(ctx);
```
simonthuresson, line 529: I think we could change this to use the explicit functions, and not just call visitChildren

packages/language-support/src/formatting/formatting.ts
```diff
@@ -177,6 +568,23 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     this.visit(ctx.RPAREN());
   };
 
+  visitReduceExpression = (ctx: ReduceExpressionContext) => {
+    const reduceExprGrp = this.startGroup();
```
simonthuresson, line 572: I think this group start should be moved down to below LPAREN, to align properly

packages/language-support/src/formatting/formatting.ts
tomasnyberg: 💯  will change this

packages/language-support/src/formatting/formatting.ts
tomasnyberg: `this.lastInCurrentBuffer()` sounds great, will change it to this.

packages/language-support/src/formatting/formatting.ts
tomasnyberg: sure

packages/language-support/src/formatting/formatting.ts
```diff
@@ -3,92 +3,318 @@ import { default as CypherCmdLexer } from '../generated-parser/CypherCmdLexer';
...
   lowerCase?: boolean;
   upperCase?: boolean;
+  spacingChoice?: SpacingChoice;
 }
 
+type SpacingChoice = 'SPACE_AFTER' | 'EXTRA_SPACE';
+
 export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
-  buffer: string[] = [];
+  buffers: Chunk[][] = [];
```
simonthuresson, line 85: Is the name buffers missleading?

packages/language-support/src/formatting/formatting.ts
tomasnyberg: I think it's fine to just have visitChildren, since it's more clear that we don't actually want to handle the children in a special way, we just want to have the special behavior with preserving explicit newlines and a breakline before. If we explicitly visit all children it's more code for no real reason and it might take longer to figure out what the method is trying to achieve.

packages/language-support/src/formatting/formatting.ts
tomasnyberg: Yes, nice catch! Changing this

packages/language-support/src/formatting/formatting.ts
```diff
@@ -187,11 +595,13 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   };
 
   // Handled separately since otherwise they will get weird spacing
-  // TODO: doesn't handle the special label expressions yet
-  // (labelExpression3 etc)
   visitLabelExpression = (ctx: LabelExpressionContext) => {
     this.visitRawIfNotNull(ctx.COLON());
-    this.visitIfNotNull(ctx.IS());
+    if (ctx.IS()) {
+      this.visitIfNotNull(ctx.IS());
```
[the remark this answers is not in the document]
tomasnyberg: not needed indeed, changing it to a regular visit

tomasnyberg:
> > This PR does not remove the formatQueryExperimental export, nor does it remove the checkbox for it from the codemirror playground. My reasoning is that we are likely to soon be adding an experimental formatter once again, and it would be unnecessary to remove it only to add it back soon thereafter.
> 
> I see you've addressed this in the description already. However, I think it makes sense to remove the experimental toggle as you've been looking into doing the next version without a fully copy / paste. Ideally the v3 improvements would go straight into the normal one, without the need for an "experimental" tag

Yeah I thought about it a bit more and I think you're right. @simonthuresson agrees as well.

I'll remove the flag and just make it the default. Let's just add it back in later if needed.

tomasnyberg:
I've removed the experimental formatting flag now, and the default is now simply the V2 formatter

simonthuresson (approved):

PR #400, opened 2025-03-10
Fixes EXISTS expressions

simonthuresson:
### Description
This PR improves the indentation of EXISTS expressions in Cypher queries. Similar to CALL expressions, the content inside EXISTS expressions should be properly indented, but when these expressions appear within WHERE clauses, the indentation requires special handling:

```
MATCH (user:Actor {actor_type: "EFXxFHob"})
WITH user,
     apoc.coll.intersection(labels(user), idp_label_list) AS user_idp_labels
WHERE (size(user_idp_labels) = "3INQ6teR" OR user.service_type = "UlAAMmmD") AND
      NOT EXISTS {
  MATCH (user)-[:PART_OF]->(group:Actor {actor_type: "A0cFHwrB"})
  WITH apoc.coll.intersection(labels(group), idp_label_list)
       AS group_idp_labels
  WHERE size(group_idp_labels) = "7SOM63mX" OR
        group.service_type = "11gaOJfr" AND user_idp_labels = group_idp_labels
} AND toLower(user.status) = "p6MXYsvG"
RETURN user
``` 

This PR fixes this by aligning whats inside a EXISTS expression with its base group and adding one indentation step. Previous example becomes:
```
MATCH (user:Actor {actor_type: "EFXxFHob"})
WITH user,
     apoc.coll.intersection(labels(user), idp_label_list) AS user_idp_labels
WHERE (size(user_idp_labels) = "3INQ6teR" OR user.service_type = "UlAAMmmD") AND
      NOT EXISTS {
        MATCH (user)-[:PART_OF]->(group:Actor {actor_type: "A0cFHwrB"})
        WITH apoc.coll.intersection(labels(group), idp_label_list)
             AS group_idp_labels
        WHERE size(group_idp_labels) = "7SOM63mX" OR
              group.service_type = "11gaOJfr" AND
              user_idp_labels = group_idp_labels
      } AND toLower(user.status) = "p6MXYsvG"
RETURN user

```
When nested EXISTS appear, the rules still follows, aligning by the base group and adding one indentation step:
```
MATCH (user:Actor {actor_type: "EFXxFHob"})
WITH user,
     apoc.coll.intersection(labels(user), idp_label_list) AS user_idp_labels
WHERE (size(user_idp_labels) = "3INQ6teR" OR user.service_type = "UlAAMmmD") AND
      NOT EXISTS {
        MATCH (user)-[:PART_OF]->(group:Actor {actor_type: "A0cFHwrB"})
        WITH apoc.coll.intersection(labels(group), idp_label_list)
             AS group_idp_labels
        WHERE (size(group_idp_labels) = "7SOM63mX" OR
               group.service_type = "11gaOJfr" AND
               user_idp_labels = group_idp_labels) AND EXISTS {
                MATCH (user)-[:HAS_ACCESS]->(resource:Resource)
                WHERE resource.status = "active"
              }
      } AND toLower(user.status) = "p6MXYsvG"
RETURN user
```
### Technical challenge
A notable subtlety is that the base group for the second EXISTS expression is after the second WHERE clause, not the first WHERE. This occurs because clauses like MATCH have a breakLine() method that terminates all active groups, meaning that after a line break, we need use the established groups.

### Refactorisation
This PR refactors the state management by introducing an IndentationState that:
 - Encapsulates indentation-related state in a single object
 - Allows for easier state passing between different components

### Tests
 - Modifies two tests
 - Adds one test
 - All tests passes

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -61,6 +61,7 @@ import {
 } from '../generated-parser/CypherCmdParser';
 import CypherCmdParserVisitor from '../generated-parser/CypherCmdParserVisitor';
 import {
+  AlignIndentationOptions,
   Chunk,
   CommentChunk,
   findTargetToken,
@@ -151,7 +152,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       groupsEnding: prefix.groupsEnding + suffix.groupsEnding,
       modifyIndentation: prefix.modifyIndentation + suffix.modifyIndentation,
       specialIndentation: prefix.specialIndentation + suffix.specialIndentation,
-      alignIndentation: prefix.alignIndentation + suffix.alignIndentation,
+      alignIndentation: prefix.alignIndentation,
       ...(hasCursor && { isCursor: true }),
     };
     this.currentBuffer()[indices[1]] = chunk;
@@ -280,12 +281,13 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   };
 
   addAlignIndentation = () => {
-    this.lastInCurrentBuffer().alignIndentation += 1;
+    this.lastInCurrentBuffer().alignIndentation = AlignIndentationOptions.Add;
   };
 
   removeAlignIndentation = () => {
     const idx = this.getFirstNonCommentIdx();
-    this.currentBuffer().at(idx).alignIndentation -= 1;
+    this.currentBuffer().at(idx).alignIndentation =
+      AlignIndentationOptions.Remove;
   };
 
   getBottomChild = (
@@ -354,7 +356,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
         groupsEnding: 0,
         modifyIndentation: 0,
         specialIndentation: 0,
-        alignIndentation: 0,
+        alignIndentation: AlignIndentationOptions.Maintain,
       };
       this.startGroupCounter = 0;
       this.currentBuffer().push(chunk);
@@ -393,7 +395,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
         groupsEnding: 0,
         modifyIndentation: 0,
         specialIndentation: 0,
-        alignIndentation: 0,
+        alignIndentation: AlignIndentationOptions.Maintain,
       };
       // If we have a "hard-break" comment, i.e. one that has a newline before it,
       // we end all currently active groups. Otherwise, that comment becomes part of the group,
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -90,7 +90,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   groupID = 0;
   groupStack: number[] = [];
   startGroupCounter = 0;
-  shouldEndOnNext: number[] = [];
+  groupsToEndOnBreak: number[] = [];
 
   constructor(private tokenStream: CommonTokenStream) {
     super();
@@ -109,8 +109,8 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   lastInCurrentBuffer = () => this.currentBuffer().at(-1);
 
   breakLine = () => {
-    if (this.shouldEndOnNext.length > 0) {
-      this.endGroup(this.shouldEndOnNext.pop());
+    if (this.groupsToEndOnBreak.length > 0) {
+      this.endGroup(this.groupsToEndOnBreak.pop());
     }
     if (this.currentBuffer().length > 0) {
       this.buffers.push([]);
@@ -1032,7 +1032,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       const endOfExistGroup = this.startGroup();
       this.visit(ctx.RCURLY());
       this.removeAlignIndentation();
-      this.shouldEndOnNext.push(endOfExistGroup);
+      this.groupsToEndOnBreak.push(endOfExistGroup);
     } else {
       this.visitIfNotNull(ctx.matchMode());
       this.visit(ctx.patternList());
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -90,6 +90,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   groupID = 0;
   groupStack: number[] = [];
   startGroupCounter = 0;
+  shouldEndOnNext: number[] = [];
 
   constructor(private tokenStream: CommonTokenStream) {
     super();
@@ -108,6 +109,9 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   lastInCurrentBuffer = () => this.currentBuffer().at(-1);
 
   breakLine = () => {
+    if (this.shouldEndOnNext.length > 0) {
+      this.endGroup(this.shouldEndOnNext.pop());
+    }
     if (this.currentBuffer().length > 0) {
       this.buffers.push([]);
     }
@@ -146,6 +150,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       groupsEnding: prefix.groupsEnding + suffix.groupsEnding,
       modifyIndentation: prefix.modifyIndentation + suffix.modifyIndentation,
       specialIndentation: prefix.specialIndentation + suffix.specialIndentation,
+      alignIndentation: prefix.alignIndentation + suffix.alignIndentation,
       ...(hasCursor && { isCursor: true }),
     };
     this.currentBuffer()[indices[1]] = chunk;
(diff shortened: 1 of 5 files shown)


packages/language-support/src/formatting/formatting.ts
```diff
@@ -90,6 +90,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   groupID = 0;
   groupStack: number[] = [];
   startGroupCounter = 0;
+  shouldEndOnNext: number[] = [];
```
tomasnyberg, line 93: I don't think this conveys that it is groups that should end, and "next" is a bit vague. Maybe `groupsToEndOnBreak`

packages/language-support/src/formatting/formatting.ts
```diff
@@ -108,6 +109,9 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   lastInCurrentBuffer = () => this.currentBuffer().at(-1);
 
   breakLine = () => {
+    if (this.shouldEndOnNext.length > 0) {
```
tomasnyberg, line 112: This should be a `while` right? I'm assuming there can be more than one.

Also, I'm not sure how happy I am with using groups this way. I get that it is a workaround to get the alignment to work, but it feels like we should rather fix the underlying cause. I'm not totally opposed to introducing this workaround to make this work, but I think we should at least have some plan regarding rethinking how groups translate between different chunk lists. Perhaps allowing groups to transfer between chunkLists is not a bad idea.

packages/language-support/src/formatting/formatting.ts
```diff
@@ -146,6 +150,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       groupsEnding: prefix.groupsEnding + suffix.groupsEnding,
       modifyIndentation: prefix.modifyIndentation + suffix.modifyIndentation,
       specialIndentation: prefix.specialIndentation + suffix.specialIndentation,
+      alignIndentation: prefix.alignIndentation + suffix.alignIndentation,
```
tomasnyberg, line 153: should this be a `+` ? In what cases would we want to take the sum of them. I feel like it should only be `prefix.alignIndentation`

packages/language-support/src/formatting/formatting.ts
```diff
@@ -273,6 +278,15 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     this.lastInCurrentBuffer().specialIndentation -= 1;
   };
 
+  addAlignIndentation = () => {
+    this.currentBuffer().at(-1).alignIndentation += 1;
+  };
+
+  removeAlignIndentation = () => {
```
tomasnyberg, line 285: Could the indentation methods be collected into a method, similar to `setChunkProperty`? This is now 6 different methods for setting indentation on a chunk, and it's starting to feel a bit messy.

packages/language-support/src/formatting/formatting.ts
```diff
@@ -1005,17 +1023,26 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     this.visit(ctx.EXISTS());
     this.avoidBreakBetween();
     this.visit(ctx.LCURLY());
-    if (ctx.regularQuery()) {
-      this.addIndentation();
+    const isRegularQuery = ctx.regularQuery() ? true : false;
```
tomasnyberg: unncessary conversion to true/false. I also prefer the previous version, as I  think it's a bit verbose compared to just 

`if (ctx.regularQuery())`

It's also a bit misleading in terms of meaning; "isRegularQuery" indicates that it pertains to the whole EXISTS expression, when it just relates to what type of inner query it is.

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -48,6 +48,7 @@ export interface BaseChunk {
   groupsEnding: number;
   modifyIndentation: number;
   specialIndentation: number;
+  alignIndentation: number;
```
tomasnyberg, line 51: Can't these three be put into an `IndentationState`?

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -70,8 +68,13 @@ interface FinalResultWithPos {
 
 interface Indentations {
```
tomasnyberg, line 69: I think maybe the purpose of the  `Indentations` interface becomes a bit unclear when we also have `IndentationState`? The naming I mean

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -134,11 +142,18 @@ function getIndentations(curr: State, choice: Choice): Indentations {
...
     }
     // Case 2: Special indentation with active groups
-    else if (curr.specialIndentation !== 0) {
+    else if (curr.indentationState.special !== 0) {
       finalIndent =
         curr.activeGroups.length > 1
           ? curr.activeGroups.at(-1).align
-          : curr.specialIndentation;
+          : curr.indentationState.special;
+      // Case 42: align indentation
```
tomasnyberg: Case 3? 

I think the logic becomes a bit hard to follow here. I'm not against having the different cases but I think they warrant a bit of a longer explanation. Perhaps describing that this is only for cases where we have nested clauses, like in the `EXISTS` case?

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -149,11 +164,17 @@ function getIndentations(curr: State, choice: Choice): Indentations {
     // When not at the start of a line, no indentation
     finalIndent = 0;
   }
+  if (choice.left.alignIndentation < 0) {
```
tomasnyberg, line 167: shouldn't this pop as many as there are alignIndentation? I don't see why it is a number if we only pop one

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -149,11 +164,17 @@ function getIndentations(curr: State, choice: Choice): Indentations {
...
+    finalIndent = align.pop();
+  }
 
   return {
-    nextBaseIndentation: nextBaseIndent,
-    nextSpecialIndentation: nextSpecialIndent,
     finalIndentation: finalIndent,
+    indentationState: {
```
tomasnyberg, line 173: nice

tomasnyberg (commented):
Good improvement to EXISTS expressions, but I'm not sure if we should wait for this before releasing V2 in Query. Thoughts @ncordon @OskarDamkjaer ?

packages/language-support/src/formatting/formatting.ts
simonthuresson: I agree, `groupsToEndOnBreak`is better

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: Agree, the naming does not reflect the purpose of this interface, IndentationResult perhaps?

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -400,15 +419,19 @@ export function buffersToFormattedString(
...
+      },
       overflowingCount: 0,
       edge: null,
     };
     const result = bestFirstSolnSearch(initialState, choices);
-    indentation = result.indentation;
-    specialIndentation = result.specialIndentation;
+    indentation = result.indentation.base;
+    specialIndentation = result.indentation.special;
+    align = result.indentation.align;
```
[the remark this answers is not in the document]
simonthuresson: I'll fix, that would be much cleaner code

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: I agree that the logic has become difficult to follow. I have improved the comments to hopefully make the logic a bit more clear

packages/language-support/src/formatting/formatting.ts
simonthuresson: I do think there can only one each time. I even think it should be only one each time.

packages/language-support/src/formatting/formattingHelpers.ts
simonthuresson: I cant use the IndentationState introduced for state because that has list of numbers as type for aligning. For that reason im not sure it makes to create a new interface to put these three in.

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: I am quite sure that it can only pop one time. I use a number because it is an easy way to represent three states (add, remove or do nothing). I could introduce an ENUM perhaps?

ncordon (commented):
Could this improvement be made on `COUNT` and `COLLECT` as well? Is that on your radar already?

simonthuresson:
> Could this improvement be made on `COUNT` and `COLLECT` as well? Is that on your radar already?

Yes this improvement can definitely be applied to `COUNT` and `COLLECT` as well. This was not on our radar, good catch!

ncordon (approved):
I took a long look yesterday and it was working well for me 😄

PR #404, opened 2025-03-11
Fixes count and collect

simonthuresson:
### Description
This PR fixes correct indentation for `COUNT` and `COLLECT`. This was something that was discovered in this PR: https://github.com/neo4j/cypher-language-support/pull/400
as something which should have similar behavior to `EXISTS`.
The following are before and after for `COUNT` and `COLLECT` respectively:
Before:
```
MATCH (person:Person)
RETURN person.name AS name, COUNT {
MATCH (person)-[:HAS_DOG]->(dog:Dog)
RETURN dog.name AS petName
  UNION
MATCH (person)-[:HAS_CAT]->(cat:Cat)
RETURN cat.name AS petName } AS numPets
```
After:
```
MATCH (person:Person)
RETURN person.name AS name, COUNT {
         MATCH (person)-[:HAS_DOG]->(dog:Dog)
         RETURN dog.name AS petName
           UNION
         MATCH (person)-[:HAS_CAT]->(cat:Cat)
         RETURN cat.name AS petName
       } AS numPets
```
Before:
```
MATCH (person:Person)
RETURN person.name AS name, COLLECT {
MATCH (person)-[r:HAS_DOG]->(dog:Dog)
WHERE r.since > 2017
RETURN dog.name } AS youngDogs
```
After:
```
MATCH (person:Person)
RETURN person.name AS name, COLLECT {
         MATCH (person)-[r:HAS_DOG]->(dog:Dog)
         WHERE r.since > 2017
         RETURN dog.name
       } AS youngDogs
```

### Additional
This PR does not only add visitors for `COUNT` and `COLLECT`, but also makes sure that `UNION` is correctly indented in COUNT. This is made sure by using both align indentation and the old baseIndentation. 

### Tech debt
We acknowledge that this PR has tech debt that was introduced in: https://github.com/neo4j/cypher-language-support/pull/400. This is related to that there are mixed use of indentation properties. Chunks uses similar indentation properties as the indentation interface used by State. But it currently cant use that interface because of those differences. These will be cleared up in later PR.

### Tests
 - Two tests added for `COUNT`
 - One test added for `COLLECT`

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -7,6 +7,7 @@ import {
   CaseAlternativeContext,
   CaseExpressionContext,
   ClauseContext,
+  CollectExpressionContext,
   CountStarContext,
   CreateClauseContext,
   DeleteClauseContext,
@@ -1042,6 +1043,19 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       this.visitIfNotNull(ctx.whereClause());
       this.visit(ctx.RCURLY());
     }
+  };
+
+  visitCollectExpression = (ctx: CollectExpressionContext) => {
+    this.visit(ctx.COLLECT());
+    this.avoidBreakBetween();
+    this.visit(ctx.LCURLY());
+    this.addAlignIndentation();
+    this.visit(ctx.regularQuery());
+    this.breakLine();
+    const endOfCollectGroup = this.startGroup();
+    this.visit(ctx.RCURLY());
+    this.removeAlignIndentation();
+    this.groupsToEndOnBreak.push(endOfCollectGroup);
   };
 
   visitCaseAlternative = (ctx: CaseAlternativeContext) => {
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -8,6 +8,7 @@ import {
   CaseExpressionContext,
   ClauseContext,
   CollectExpressionContext,
+  CountExpressionContext,
   CountStarContext,
   CreateClauseContext,
   DeleteClauseContext,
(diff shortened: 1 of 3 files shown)


packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -150,13 +150,14 @@ function getIndentations(curr: State, choice: Choice): IndentationResult {
         curr.activeGroups.length > 1
           ? curr.activeGroups.at(-1).align
           : curr.indentationState.special;
-      // Case 3: Currently only for EXISTS,
-      // Aaligning with base group plus indentation
+      // Case 3: Currently for for EXISTS, COLLECT and COUNT,
+      // Aligning with base group plus indentation plus possible baseIndentation
+      // baseIndentation can happen with UNION
```
tomasnyberg, line 155: baseIndentation twice?

packages/language-support/src/formatting/formatting.ts
```diff
@@ -1044,6 +1046,42 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    this.visit(ctx.COUNT());
+    this.avoidBreakBetween();
+    this.visit(ctx.LCURLY());
+
+    if (ctx.regularQuery()) {
+      this.addAlignIndentation();
+      this.visit(ctx.regularQuery());
+      this.breakLine();
+      this.mustBreakBetween();
+      // This is a workaround because group does not translate between chunk lists
```
tomasnyberg, line 1072: nit: comment here but not in tthe comment above

tomasnyberg (approved):
Cool, can the description outline what we discussed about tech debt, so that we keep all context?

packages/language-support/src/tests/formatting/linebreaks.test.ts
```diff
@@ -463,6 +463,35 @@ WHERE ((size(apoc.coll.intersection(labels(user), idp_label_list)) = "3INQ6teR"
       } AND user.last_login > datetime() - duration({days: 30}))`;
     verifyFormatting(query, expected);
   });
+  test('test that clause under collect gets properly indented', () => {
+    const query = `MATCH (person:Person)
+RETURN person.name AS name, COLLECT {
+         MATCH (person)-[r:HAS_DOG]->(dog:Dog)
```
ncordon, line 469: What are we aligning this with? Shouldn't it be aligned with the column of `person`? Is it because we are adding an extra two spaces?

simonthuresson: Yes we are aligning with person plus two spaces. The idea is to signal it is part of RETURN but still its own sub clause.

PR #405, opened 2025-03-11
Adds preservation of newlines between statements in formatter

tomasnyberg:
## Description

PR #382 added preservation of explicit newlines, but only between different clauses. As a result, in queries like:

```
MATCH (a)
RETURN a;

MATCH (b)
RETURN b;
```

the blank line  after `a;` is removed. This PR allows preserving newlines between statements, so that users can separate statements with a blank line.

### Testing

4 new simple tests for statements with newlines after them

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -290,16 +290,16 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     throw new Error('Internal formatting error in findBottomChild');
   };
 
-  preserveExplicitNewlineAfter = (ctx: ParserRuleContext) => {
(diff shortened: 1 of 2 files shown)


packages/language-support/src/formatting/formatting.ts
```diff
@@ -414,6 +414,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
           this.currentBuffer().pop();
         }
         this.visit(ctx.SEMICOLON(i));
+        this.preserveExplicitNewlineAfter(ctx.SEMICOLON(i));
```
simonthuresson, line 417: Will the semicolon still be visited?

ncordon (approved):

packages/language-support/src/formatting/formatting.ts
tomasnyberg: Yes 👍

PR #409, opened 2025-03-13
Fixes tech debt (indentations, outdated code) for formatter

simonthuresson:
## Description
This PR primarily addresses technical debt related to how indentation is handled. Additionally, it removes some unnecessary checks, improving code clarity.

### Indentation
In PR: https://github.com/neo4j/cypher-language-support/pull/404 and https://github.com/neo4j/cypher-language-support/pull/400 a changes was made to how indentations was made. However, these also introduced tech debt in form that it became much more difficult to understand how indentation works. This is hopefully improved in this PR. The main idea is to create a clear distinction between how indentation is handled in the Chunk, and how the chunk affects the State:s indentation.

The function which calculates the indentation for the current chunk also got an overhaul. Instead of nested if statements we do a if check and return immediately. This hopefully helps with understanding what is responsible of what and what should be returned.

### Cleanup of Unnecessary Checks
This PR also removes redundant checks introduced in [#375](https://github.com/neo4j/cypher-language-support/pull/375). That PR changed the behavior so that only RegularChunk or CommentChunk can be present in the buffers and ChunkList, making certain checks unnecessary.

### Unnecessary exports
This PR also found some interfaces which was unnecessarily exported, these exports was removed. 

### Test
 - All tests passes
 - The sanity check with 10 000 queries passes

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -69,6 +69,7 @@ import {
   CommentChunk,
   findTargetToken,
   getParseTreeAndTokens,
+  initialIndentation,
   isComment,
   RegularChunk,
   wantsToBeConcatenated,
@@ -375,11 +376,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
         text,
         groupsStarting: this.startGroupCounter,
         groupsEnding: 0,
-        indentation: {
-          base: 0,
-          special: 0,
-          align: AlignIndentationOptions.Maintain,
-        },
+        indentation: { ...initialIndentation },
       };
       this.startGroupCounter = 0;
       this.currentBuffer().push(chunk);
@@ -416,11 +413,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
         text,
         groupsStarting: 0,
         groupsEnding: 0,
-        indentation: {
-          base: 0,
-          special: 0,
-          align: AlignIndentationOptions.Maintain,
-        },
+        indentation: { ...initialIndentation },
       };
       // If we have a "hard-break" comment, i.e. one that has a newline before it,
       // we end all currently active groups. Otherwise, that comment becomes part of the group,
@@ -714,11 +707,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       node,
       groupsStarting: this.startGroupCounter,
       groupsEnding: 0,
-      indentation: {
-        base: 0,
-        special: 0,
-        align: AlignIndentationOptions.Maintain,
-      },
+      indentation: { ...initialIndentation },
     };
     this.startGroupCounter = 0;
     if (node.symbol.tokenIndex === this.targetToken) {
@@ -759,11 +748,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       node,
       groupsStarting: this.startGroupCounter,
       groupsEnding: 0,
-      indentation: {
-        base: 0,
-        special: 0,
-        align: AlignIndentationOptions.Maintain,
-      },
+      indentation: { ...initialIndentation },
     };
     this.startGroupCounter = 0;
     if (node.symbol.tokenIndex === this.targetToken) {
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -272,8 +272,13 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     type: 'base' | 'special' | 'align',
   ) => {
     const chunk = this.lastInCurrentBuffer();
-    if (type === 'align' && chunk[type] !== AlignIndentationOptions.Maintain) {
-      new Error('Error: align should not be modified if it is not maintain');
+    if (
+      type === 'align' &&
+      chunk.indentation[type] !== AlignIndentationOptions.Maintain
+    ) {
+      throw new Error(
+        'Error: align should not be modified if it is not maintain',
+      );
     }
     if (modifier === 'add') {
       chunk.indentation[type] += 1;
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -278,11 +278,11 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     }
   };
 
-  addIndentation = () => {
+  addBaseIndentation = () => {
     this.setIndentationProperty('add', 'base');
   };
 
-  removeIndentation = () => {
+  removeBaseIndentation = () => {
     this.setIndentationProperty('remove', 'base');
   };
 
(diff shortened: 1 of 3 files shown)


packages/language-support/src/formatting/formatting.ts
```diff
@@ -265,31 +267,48 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    modifier: 'add' | 'remove',
+    type: 'base' | 'special' | 'align',
+  ) => {
+    const chunk = this.lastInCurrentBuffer();
+    if (
+      type === 'align' &&
+      chunk.indentation[type] !== AlignIndentationOptions.Maintain
+    ) {
+      throw new Error(
+        'Error: align should not be modified if it is not maintain',
```
simonthuresson, line 280: Could perhaps use a better error message?

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -108,12 +109,14 @@ const emptyChunk: RegularChunk = {
   text: '',
   groupsStarting: 0,
   groupsEnding: 0,
-  modifyIndentation: 0,
-  specialIndentation: 0,
-  alignIndentation: 0,
+  indentation: {
+    base: 0,
+    special: 0,
+    align: 0,
```
tomasnyberg, line 115: Shouldn't this be the maintain type? Or replaced by some standard `noIndentation` object (see previous comment)

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -122,64 +125,100 @@ export function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
-    curr.indentationState.special +
-    choice.left.specialIndentation * INDENTATION;
-  let finalIndent = currBaseIndent;
-  const align = [...curr.indentationState.align];
-  if (choice.left.alignIndentation === AlignIndentationOptions.Add) {
-    align.push(curr.activeGroups.at(0).align);
+function deriveNextIndentationState(
+  chunkIndentation: ChunkIndentation,
+  indentationState: IndentationState,
+): IndentationState {
```
tomasnyberg, line 131: 🔥

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -122,64 +125,100 @@ export function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
+  return {
+    base: indentationState.base + chunkIndentation.base * INDENTATION,
```
tomasnyberg, line 133: nit: the `INDENTATION` constant has a bit of an outdated name at this point given that we use that word so much now. Maybe `INDENTATION_SPACES`?

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -122,64 +125,100 @@ export function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
+function getIndentations(state: State, chunk: Chunk): IndentationResult {
+  const { base, special, align } = deriveNextIndentationState(
+    chunk.indentation,
+    state.indentationState,
+  );
+
+  // AlignIndentation, used for EXISTS, COUNT, COLLECT
+  // Pushes base groups alignment to list to be used later
+  // for closing bracket
+  if (chunk.indentation.align === AlignIndentationOptions.Add) {
```
tomasnyberg, line 148: nit: was it possible to move this one to below the `AlignIndentationOptions.Remove` one? I think we discussed this in person

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -122,64 +125,100 @@ export function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
   return {
-    finalIndentation: finalIndent,
-    indentationState: {
-      base: nextBaseIndent,
-      special: nextSpecialIndent,
-      align: align,
-    },
+    finalIndentation: state.indentationState.base,
+    indentationState: { base, special, align },
   };
```
tomasnyberg, line 222: love the refactoring of this method in general; the "case x:" idea is a lot more clear than the long if / else from before. Good that you've described which cypher constructs use the specific types of alignment as well.

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -198,15 +237,14 @@ function getNeighbourState(curr: State, choice: Choice, split: Split): State {
...
+    choice.left,
+  );
 
   const actualColumn = curr.column === 0 ? finalIndentation : curr.column;
   const splitLength = !isBreak ? split.splitType.length : 0;
-  const leftLength =
-    choice.left.type === 'COMMENT' || choice.left.type === 'REGULAR'
-      ? choice.left.text.length
-      : 0;
-  const thisWordEnd = actualColumn + leftLength + splitLength;
```
tomasnyberg, line 209: nice! I guess this was left over from when we had groups as chunks?

tomasnyberg (commented):
Very nice 💯 ! IMO this resolves all the tech debt we had related to indentation. Solid description as well, the context might be important in the future.

nit: can the title be `Fixes indentation tech debt for formatter` to make it clear that it's the indentation tech debt specifically?

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: Agree, changing to `INDENTATION_SPACES` does reflect more its intent

packages/language-support/src/formatting/formatting.ts
```diff
@@ -356,9 +375,11 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
         text,
         groupsStarting: this.startGroupCounter,
         groupsEnding: 0,
-        modifyIndentation: 0,
-        specialIndentation: 0,
-        alignIndentation: AlignIndentationOptions.Maintain,
+        indentation: {
```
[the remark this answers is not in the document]
simonthuresson: Yepp, thought about that as well. Will fix!

packages/language-support/src/formatting/formatting.ts
simonthuresson: I like that error message better. Will change to that.

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -17,12 +18,12 @@ https://github.com/neo4j/cypher-language-support.`.trim();
 const INDENTATION = 2;
 const showGroups = false;
 
-export interface Split {
+interface Split {
```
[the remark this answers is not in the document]
simonthuresson: I will add it to description. Thought it was too small to being mentioned, but i'll add a sentence

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: Fixed!

simonthuresson:
> nit: can the title be `Fixes indentation tech debt for formatter` to make it clear that it's the indentation tech debt specifically?

That was my first thought, however it does not only cover tech debt about indentation I feel like? It is more about tech debt in general. But I do agree that the indentation part is the main focus...

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: I moved the `AlignIndentationOptions.Remove` up to above `AlignIndentationOptions.Add` instead. The reason being that we have to do the Add before a return with no indentation. With this change we follow the pattern of returning as early as possible while retaining the logic of adding alignIndentation when needed.

ncordon (approved):

PR #411, opened 2025-03-14
Improves EXISTS, COLLECT and COUNT nested with CASE

simonthuresson:
## Description
This PR improves cases where you have `EXISTS`, `COLLECT` and/or `COUNT` nested together with `CASE` in different arregnments. The issue was initially identified when introducing V2, and the following example was provided:
```
MATCH (n)
RETURN
  CASE
    WHEN (n)--() THEN
      CASE
        WHEN EXISTS {
        MATCH (person)-[:HAS_DOG]->(dog:Dog)
        WHERE person.name = 'Chris'
        WITH dog
        WHERE dog.name = 'Ozzy'
    } THEN 'Relationship'
        WHEN (n {prop: 42}) THEN 'Relationship'
      END
  END
```

### Issue
The problem arises because the closing bracket for EXISTS aligns incorrectly. The root cause is that the closing bracket attempts to align with the first active group, which, in this case, is the first WHEN clause. However, it should instead align with the second WHEN clause.

### Solution
To address this, we redesigned how CASE works. Previously, CASE relied on mustBreakBefore(), but we replaced this with breakLine instead. This change ensures that each WHEN clause starts with zero active groups, preventing misalignment.

With this new approach, EXISTS no longer aligns incorrectly with the first WHEN but instead follows the expected formatting behavior. The corrected formatting for the above example is now:
```
MATCH (n)
RETURN
  CASE
    WHEN (n)--() THEN
      CASE
        WHEN EXISTS {
        MATCH (person)-[:HAS_DOG]->(dog:Dog)
        WHERE person.name = 'Chris'
        WITH dog
        WHERE dog.name = 'Ozzy'
        } THEN 'Relationship'
        WHEN (n {prop: 42}) THEN 'Relationship'
      END
  END
```

Now the closing brackets align properly. But inside  `EXISTS` we dont get the indentation we want. This is because in the logic for finding correct indentation, the first check is for any `CASE` alignment which it matches. To address this, we simply add one `INDENTATION_SPACES` if inside `CASE` alignment and `EXISTS` alignment is present, yielding this result:

```   
MATCH (n)
RETURN
  CASE
    WHEN (n)--() THEN
      CASE
        WHEN EXISTS {
          MATCH (person)-[:HAS_DOG]->(dog:Dog)
          WHERE person.name = 'Chris'
          WITH dog
          WHERE dog.name = 'Ozzy'
        } THEN 'Relationship'
        WHEN (n {prop: 42}) THEN 'Relationship'
      END
  END
```

## Not included
This PR does not consider cases where you have a `CASE` inside a `EXISTS` expression, this likely requires major re work  of how `CASE` works. 

## Refactoring
This PR does includes some refactorization of the two function `getIndentations` (now `getFinalIndentation`) and `deriveNextIndentationState`,  which was coupled before somewhat unnecessary.

## Tests
 - Four tests added which has very complex nesting of `EXISTS` inside `CASE`


packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -156,38 +156,32 @@ function getIndentations(state: State, chunk: Chunk): IndentationResult {
     };
   }
 
-  // Case 1: Hard-break comments align with base group or base indentation
-  if (chunk.type === 'COMMENT' && chunk.breakBefore) {
-    const baseGroup = state.activeGroups[0];
-    const finalIndent = baseGroup ? baseGroup.align : base;
+  // Active groups, prioritize lining up on these if break
+  // Works because after breakLine no groups are active
```
tomasnyberg, line 160: i feel like this comment is hard to understand outside the context of this PR. I get that it relates to the fact that since we do breakLine() in case this will be accurate but it's hard to understand at a glance

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -156,38 +156,32 @@ function getIndentations(state: State, chunk: Chunk): IndentationResult {
...
   }
 
-  // Case 2: Special indentation, used with CASE
-  if (state.indentationState.special !== 0) {
-    const finalIndent =
-      state.activeGroups.length > 1
-        ? state.activeGroups.at(-1).align
-        : state.indentationState.special;
-
+  // Hard-break comments or chunk must break before
```
tomasnyberg, line 168: why change this comment? it's also not `or must break before` (it's &&) so it's a bit misleading.

suggestion : `Case 1: Hard-break comments align with base group or base indentation`

also I think we should keep the Case x: ... thing

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -156,38 +156,32 @@ function getIndentations(state: State, chunk: Chunk): IndentationResult {
...
     };
   }
 
-  // Case 3: Currently for EXISTS, COLLECT and COUNT
-  if (state.indentationState.align.length > 0) {
-    // base case
-    let finalIndent =
-      align.at(-1) + INDENTATION_SPACES + state.indentationState.base;
+  // Special indentation, used with CASE
+  if (state.indentationState.special !== 0) {
```
tomasnyberg, line 177: Does the order of all these cases have to change in order to make the logic work? If not, then I would prefer keeping them as is. If it has to change for it to work it's fine obviously

packages/language-support/src/formatting/formattingSolutionSearch.ts
tomasnyberg: also I don't get what the "if break" part means

packages/language-support/src/tests/formatting/edgecases.test.ts
```diff
@@ -274,8 +274,8 @@ AND TRUE AND TRUE AND TRUE AND TRUE AND TRUE AND TRUE AND TRUE THEN "(FK)" ELSE
     }) as columns`;
     const expected = `WITH s, t.name AS tableName, collect({name: c.name, pk:
   CASE (NOT pk IS NULL AND $printKeyInfo)
-    WHEN true AND true AND true AND true AND
-         true AND true AND true AND true AND true THEN "(PK)"
+    WHEN true AND true AND true AND true AND true AND true AND true AND true AND
```
tomasnyberg, line 277: any reason why this one is suddenly behaving differently? This verison is probably better I just don't see why it would change

packages/language-support/src/tests/formatting/edgecases.test.ts
```diff
@@ -669,4 +669,197 @@ RETURN n`;
...
+                 toString(interestList)
+        ELSE 'Active smartphone user with few interests: ' +
+        toString(interestList)
+      END
+    ELSE
+      CASE
+        WHEN NOT EXISTS {
+          MATCH (u)-[:OWNS]->(:Device {type: 'Smartphone'})
+        } AND likeCount <= 10
+        THEN 'Less active user without a smartphone, interests: ' +
```
tomasnyberg, line 831: should these not align  such that they come after the THEN? Or did we decide against that

packages/language-support/src/tests/formatting/edgecases.test.ts
```diff
@@ -669,4 +669,197 @@ RETURN n`;
...
+WITH u, count((u)-[:LIKES]->()) AS likeCount,
+     collect(DISTINCT u.interests) AS interestList
+RETURN
+  CASE
+    WHEN EXISTS {
+      MATCH (u)-[:OWNS]->(:Device {type: 'Smartphone'})
+    } AND likeCount > 10 THEN
+      CASE p.name
+        WHEN size(interestList)
+             > 3 THEN 'Active smartphone user with diverse interests: ' +
```
tomasnyberg, line 821: why does this one split in the middle of the > expression?

tomasnyberg (commented):
I like the end result and the tests are good 👍  Not sure about the changes to the getIndentation logic though but we can discuss

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: Yes this comment could need improvements. If feel like the content is correct though, roughly what is needed to understand the if statement.

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: I changed the comment because it does not have the same logic anymore. I agree though it should not be "or", it does not reflect the `&&`. 

Yes I might introduce Case x:  again, it felt less needed now though with how the logic is structured.

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: No they have actually not changed. What happened though is I moved out the logic for active groups to have its own if/return statement. I think this pushed everything down, which is probably what you see.

packages/language-support/src/tests/formatting/edgecases.test.ts
simonthuresson: No I dont really know, my guess is one of those places where the memoization because slightly different because of something unrelated as we have experienced before. 
Agree, this is better.

packages/language-support/src/tests/formatting/edgecases.test.ts
simonthuresson: Its because as we discussed we have no visitor for comparator operators. So you have group around `size(..)` and one starting with `THEN`, so the cheapest place to break is around the comparator.

packages/language-support/src/tests/formatting/edgecases.test.ts
simonthuresson: Im not entirely sure what you are referring to. But we decided that after a closing bracket it should align under it. The reason being that it should be clear that we are still part of the statement. Such as for `WHERE ...`. What also happens here is that because we start groups around `THEN` its cheaper to break before than inside the `THEN` statement.

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -116,100 +111,82 @@ function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
-  // Case 1: Hard-break comments align with base group or base indentation
+  // Case 3: If chunk is comment and has propery breakBefore
   if (chunk.type === 'COMMENT' && chunk.breakBefore) {
-    const baseGroup = state.activeGroups[0];
-    const finalIndent = baseGroup ? baseGroup.align : base;
-    return {
-      finalIndentation: finalIndent,
-      indentationState: { base, special, align },
-    };
+    return nextIndentationState.base;
```
simonthuresson, line 160: Anything else needed to be added to this comment? Dont know what else to write

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -116,100 +111,82 @@ function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
-    return {
-      finalIndentation: 0,
-      indentationState: { base, special, align },
-    };
+  // Case 2: This happens if align.pop() was done in deriveNextIndentationState
+  // Also means we are on a closing brackets and should align
+  // on what was the last element in align, however because we popped it before
+  // we need to fetch it from the state where it still exists
+  if (state.indentationState.align.length > nextIndentationState.align.length) {
+    return state.indentationState.align.at(-1);
```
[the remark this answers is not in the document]
simonthuresson: No this is actually not an alternative to starting groups at `{`. We still have to do that. While this code is not ideal, it was the best way I could come up with to fully separate `getIndentation` and `deriveNextIndentationStep`. Because we actually need a value from getIndentation (the align.pop()). This is a somewhat hacky solution to get it.

packages/language-support/src/formatting/formattingSolutionSearch.ts
simonthuresson: Improved the comment. Included additional information about why we use nextIndentationState as well.

PR #412, opened 2025-03-17
Adds (partial) formatting of syntactically incorrect queries

tomasnyberg:
## Description
As discussed in [this slack thread](https://neo4j.slack.com/archives/C05R69VT7HU/p1741787727103149) formatting, or at least partial formatting, of syntactically incorrect queries would be very helpful. Especially if the syntax error is a very minor one, such as using `.` instead of `:`. This PR makes the formatter more lenient toward syntactically incorrect queries, and produces a "best-effort" result if possible. 

We originally strayed away from doing this since the assumption was that most formatters do not support formatting of syntactically incorrect queries, but that assumption turned out to be flawed. Although they do not always produce the best output if there are syntax errors present, many formatters make an attempt to format regardless of what is present in the document.

### Examples (taken from the tests included in this PR):

Unclosed string 
```
match (n:Person {name:'Alice}) return n
->
MATCH (n:Person {name:'Alice})
RETURN n
```
Dot instead of colon in the node properties
```
call apoc.load.json(url) yield value as v
merge (a:Article {v.article_number})
on create set a += v {.content_text, .published_date, .title, .url }
merge (s:Source {name: v.source})
merge (a)-[:PUBLISHED_IN]->(s)
with a, v where trim(v.authors) <> '' 
unwind split(v.authors,',') as name
merge (author:Author {name:name})
merge (a)-[:WRITTEN_BY]->(author)

-> 

CALL apoc.load.json(url) YIELD value AS v
MERGE (a:Article {v.article_number})
  ON CREATE SET a += v {.content_text, .published_date, .title, .url}
MERGE (s:Source {name: v.source})
MERGE (a)-[:PUBLISHED_IN]->(s)
WITH a, v
WHERE trim(v.authors) <> ''
UNWIND split(v.authors, ',') AS name
MERGE (author:Author {name: name})
MERGE (a)-[:WRITTEN_BY]->(author)
```

### Cases where we give up
Since syntactically incorrect queries can consist of literally anything, there will be cases where the parser just gives up. These cases occur mostly when there are multiple incorrect tokens in a row, like for instance this query 

```
MATCH (n:Person)
CALL {
  WITH n
  MATCH (m:Movie)
  syntax error inside subquery
  RETURN m
}
RETURN n
```
The parser gets completely lost after the `syntax error ...` part and so the formatter cannot produce any output that is useful. We then instead throw this error: `Unable to format query due to syntax error near } at line 7`

The way I verify this is by counting the amount of non-whitespace errors in the query before and after formatting it. If they differ, we either throw a syntax error like above, or an internal formatting error if there was not a syntax error in the query.

### Code changes

- Since we can now no longer rely on any token actually being where we expect them to be, this PR now makes every .visit() call the .visitIfNotNull() method instead (see the `_visit()` method, which replaces visitIfNotNull)
- The visitor now keeps track of the last token it visited, and when it encounters an erroneous token, it will be the first token it recognizes. We then add all the text in between the last token and the erroneous token since they are not part of the parse tree.
- If the parser encounters an unrecoverable error, we save all text that comes after that point (in the `unParseable` variable) and do not create any new chunks for tokens that come after that point. Once we are done visiting, we instead add all of the trailing (erroneous) text to the result. Everything that comes before that point is formatted as per usual.


packages/language-support/src/formatting/formatting.ts
```diff
@@ -321,6 +365,9 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     if (ctx instanceof TerminalNode) {
       return ctx;
     }
+    if (ctx.getChildCount() === 0) {
+      return null;
```
simonthuresson, line 369: Why null?

packages/language-support/src/formatting/formatting.ts
```diff
@@ -432,27 +487,62 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
   };
 
-  visitRawIfNotNull = (ctx: TerminalNode, options?: RawTerminalOptions) => {
+  _visitTerminalRaw = (ctx: TerminalNode, options?: RawTerminalOptions) => {
     if (ctx) {
       this.visitTerminalRaw(ctx, options);
     }
   };
 
+  // Error nodes are attacahed to the first token that the parser recognizes again.
```
simonthuresson, line 502: Nit: typo

tomasnyberg: nice catch, fixing

packages/language-support/src/formatting/formatting.ts
tomasnyberg: the idea is to explicitly return null (rather than undefined) to show that there is supposed to be something there, but due to an exceptional case (in this case there not being a node there because of a syntax error) there isn't.

packages/language-support/src/formatting/formatting.ts
```diff
@@ -69,16 +75,20 @@ import {
...
   initialIndentation,
   isComment,
   RegularChunk,
+  SyntaxErrorChunk,
   wantsToBeConcatenated,
   wantsToBeUpperCase,
 } from './formattingHelpers';
 import { buffersToFormattedString } from './formattingSolutionSearch';
 
+const MISSING = '<missing';
```
simonthuresson, line 90: Is there any reasons for using this string? Or is just an arbitrary value?

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -385,6 +380,9 @@ function decisionsToFormatted(decisions: Decision[]): FinalResult {
     buffer.push(decision.chosenSplit.splitType);
   });
   let result = buffer.join('').trimEnd();
+  if (decisions.at(0).left.type === 'SYNTAX_ERROR') {
+    result = result.trimStart();
```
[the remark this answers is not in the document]
tomasnyberg: it's to avoid weird behavior with joining the chunklists together with \n. I added a comment

packages/language-support/src/formatting/formatting.ts
```diff
@@ -140,11 +181,14 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     if (indices.length < 2) {
       return;
     }
+    if (
+      this.currentBuffer()[indices[0]].type !== 'REGULAR' ||
+      this.currentBuffer()[indices[1]].type !== 'REGULAR'
+    ) {
+      return;
```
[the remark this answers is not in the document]
tomasnyberg: It is but I'm not sure if it becomes cleaner since more logic has to happen in the for loop in that case. we'd have to check essentially the same thing but in the for loop

packages/language-support/src/formatting/formatting.ts
```diff
@@ -1537,15 +1632,20 @@ export function formatQuery(
...
+  const visitor = new TreePrintVisitor(
+    tokens,
+    unParseable,
+    firstUnParseableToken,
+  );
 
   tokens.fill();
 
-  if (cursorPosition === undefined) return visitor.format(tree);
+  if (cursorPosition === undefined) return visitor.format(tree, query);
```
simonthuresson, line 1645: Move everything inside the constructor?

packages/language-support/src/formatting/formatting.ts
simonthuresson: There is no constant for this to be used?

simonthuresson (commented):
Nice! 

Does it make sense to add description about this in our README? I think it tangents on being needed. Definitely not a must, but could help.

packages/language-support/src/formatting/formatting.ts
tomasnyberg: I doubt it since the full string is e.g. <missing )> (depending on what character it is).

packages/language-support/src/formatting/formatting.ts
tomasnyberg: yes sounds good

tomasnyberg:
> Does it make sense to add description about this in our README?
Hmm not sure, maybe. It's not really part of the architecture per se but it might be hard to see the bigger picture just through the code. The description in the PR kind of serves the same purpose

packages/language-support/src/formatting/formatting.ts
tomasnyberg: Yeah it's not super obvious. I added a comment 👍

packages/language-support/src/tests/formatting/syntaxerror.test.ts
```diff
@@ -0,0 +1,177 @@
+import { formatQuery } from '../../formatting/formatting';
+import { verifyFormatting } from './testutil';
+
+describe('formatting despite syntax errors', () => {
```
ncordon, line 4: Could we have an example of a query that has recoverable failures in two different parts?

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -14,19 +13,16 @@ import CypherCmdParser, {
...
-    _r: Recognizer<CommonToken>,
-    offendingSymbol: T,
-    line: number,
-    column: number,
-  ) {
-    throw new Error(
-      `Could not format due to syntax error at line ${line}:${column} near "${offendingSymbol?.text}"`,
-    );
-  }
+  syntaxError() {}
```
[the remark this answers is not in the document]
tomasnyberg: Yes, not sure why I've kept it, nice catch. I've removed it now

packages/language-support/src/tests/formatting/edgecases.test.ts
```diff
@@ -22,13 +21,6 @@ RETURN 3;`;
     verifyFormatting(query, expected);
   });
 
-  test('syntax error', () => {
-    const query = 'MATCH (n) RETRUN n.prop,';
-    expect(() => formatQuery(query)).toThrowError(
-      `Could not format due to syntax error at line 1:10 near "RETRUN"`,
-    );
```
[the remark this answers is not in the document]
tomasnyberg: It no longer throws for syntax errors unless it's unrecoverable so I don't think so? I did add this test for that case:

```
  test('query that we are currently unable to handle should throw.', () => {
    const query = `MATCH (n:Person)
CALL {
  WITH n
  MATCH (m:Movie)
  syntax error inside subquery
  RETURN m
}
RETURN n`;
    expect(() => formatQuery(query)).toThrowError(
      `Unable to format query due to syntax error near } at line 7`,
    );
  });
```

packages/language-support/src/formatting/formatting.ts
```diff
@@ -97,17 +109,52 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   groupStack: number[] = [];
   startGroupCounter = 0;
   groupsToEndOnBreak: number[] = [];
-
-  constructor(private tokenStream: CommonTokenStream) {
+  previousTokenIndex: number = -1;
+  unParseable: string = '';
+  unParseableStart: number = 1e9;
```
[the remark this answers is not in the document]
tomasnyberg: I've changed it to an undefined check instead, let me know if it's still confusing

packages/language-support/src/tests/formatting/syntaxerror.test.ts
tomasnyberg: Yes, good idea. I've added three now

packages/language-support/src/formatting/formatting.ts
```diff
@@ -880,16 +980,19 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
       if (i == n - 1) {
         this.avoidSpaceBetween();
       }
     }
-    this.visit(ctx.labelExpression1());
+    this._visit(ctx.labelExpression1());
   };
 
   visitTerminal = (node: TerminalNode) => {
+    if (node.getText().startsWith(MISSING)) {
```
OskarDamkjaer, line 993: This feels a little brittle, are there not nodes that could start with this text? Thinking identifiers and the like?

packages/language-support/src/formatting/formatting.ts
```diff
@@ -112,17 +124,52 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   groupStack: number[] = [];
   startGroupCounter = 0;
   groupsToEndOnBreak: number[] = [];
-
-  constructor(private tokenStream: CommonTokenStream) {
+  previousTokenIndex: number = -1;
+  unParseable: string = '';
+  unParseableStart: number = 1e9;
```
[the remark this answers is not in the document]
tomasnyberg: I've changed it to undefined now 👍

tomasnyberg:
(responding here since I can't respond to the suggested change for some reason?) 

> If you want to be less repetitive you could do something like

```
if (
      indices
        .slice(0, 2)
        .some((idx) => this.currentBuffer()[idx].type !== 'REGULAR')
    ) {
      return;
```

Personally I prefer the explicitness of the previous version, WET and so on. I'm happy to change it/add a comment if you think it's unclear though

packages/language-support/src/formatting/formatting.ts
```diff
@@ -112,17 +124,52 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    // whitespace changes. If we did so even when there wasn't a syntax error,
+    // it should raise an internal error. If there is a syntax error, we do not guarantee
+    // we can format the query.
+    if (originalNonWhitespaceCount !== formattedNonWhitespaceCount) {
+      if (this.unParseable && this.firstUnParseableToken) {
+        throw new Error(
+          `Unable to format query due to syntax error near ${this.firstUnParseableToken?.text} at line ${this.firstUnParseableToken?.line}`,
+        );
+      }
+      throw new Error(errorMessage);
```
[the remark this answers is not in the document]
tomasnyberg: yes, good idea. I've changed it to that

OskarDamkjaer:
@tomasnyberg yes that's alright, I mostly wanted to show how it'd look without a loop

packages/language-support/src/formatting/formatting.ts
tomasnyberg: I don't think a single token can start with this text? I tried e.g. this as well

```
MATCH (n:`<missing`)
return n
```

but the token then contains the backticks.  I wasn't able to construct another case where <missing would be valid syntax and the same token. 

The reason I check for string equality instead of something else is that I couldn't find any property in the node that says it's missing (I don't think there is one?)

OskarDamkjaer: Could we add a test for something like this: `CREATE ({`<missing`: "<missing"});` (and possibly other rules for identifiers/strings) so that if we ever change the AST to not include the backticks we catch the issue in tests?

tomasnyberg: This one `CREATE ({<missing: "<missing"});` is a syntax error because of the first `<` but I can add one without it.

> if we ever change the AST to not include the backticks

Wouldn't the cypher grammar have to change how it handles strings (which feels veery unlikely) for that to happen? Or am I missing something? Regardless, I've added a test for it.

PR #415, opened 2025-03-18
Adds support for common commands to formatter

tomasnyberg:
## Description
Commands are administrative queries that you can run against a neo4j database, such as 
```
CREATE CONSTRAINT `constraintName` IF NOT EXISTS
FOR (n: nodeType)
REQUIRE (n.id) IS UNIQUE;
```
or a plethora of other ones, such as `DROP INDEX`, `SHOW TRANSACTIONS` etc. etc.

As of yet, we have had very limited support for these, beyond capitalizing the keywords. The main reason for this is that commands have grammar that is essentially completely separated from the rest of the grammar (including e.g. node and relationship patterns). This PR adds support for _most_ of the common commands.

### Scoping / context
Most queries are not commands. In fact, out of 1174529 queries, 11749 of them (1%) were commands or contained at least one command. Further, out of those 11749 commands, 7811 of them contained "CREATE CONSTRAINT" (tested with a simple grep so might not be 100% accurate). Another 1338 of them contained "CREATE INDEX". 

Given the numbers above, we have decided that we don't necessarily care about supporting 100% of commands (yet) since they are unlikely to ever reach our formatter (a lot of commands also don't need formatting, e.g. `SHOW TRANSACTIONS`). At some point, when the formatter is in a more stable state, it might make sense to go for full support, but it is not prioritized right now (we are prioritizing V3 over this).

Instead, this PR focuses on the most commonly used ones, such as `CREATE CONSTRAINT`and `CREATE INDEX`.

### Unit tests
17 new unit tests that test all the `CREATE CONSTRAINT`and `CREATE INDEX` grammar cases plus a few other things

### ✨Vibe check✨ 
I spent roughly 25 minutes clicking through all the commands from the sample queries, and evaluated (based on vibes) whether the formatting was reasonable. All the output I saw was reasonable (though I am probably missing a few rare commands).

### Verification check
I ran the verification check on the 11749 commands, and all but 6 passed. The ones that didn't pass were all `SHOW TRANSACTIONS`, which is valid cypher (I could run it in query) but does not actually follow the grammar since it is missing some required parts. This leads to an internal error in `getBottomChild` because of the missing node. The syntax error Pr (#412) will take care of this however so I have not fixed it here.

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -577,6 +577,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   visitCommandOptions = (ctx: CommandOptionsContext) => {
     this.breakLine();
     this.visit(ctx.OPTIONS());
+    this.avoidBreakBetween();
     this.visit(ctx.mapOrParameter());
   };
 
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -459,6 +459,11 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     }
   };
 
+  breakAndVisitChildren = (ctx: ParserRuleContext) => {
+    this.breakLine();
+    this.visitChildren(ctx);
+  };
+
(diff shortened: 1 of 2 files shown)


OskarDamkjaer (approved):
Very modern with vibe based validation

PR #416, opened 2025-03-19
Adds integrity check of formatter as a CI/CD step

tomasnyberg:
## Description
As discussed in three different syncs (XD) and in various slack threads: 

- https://neo4j.slack.com/archives/C089SEWEJ8Z/p1741778296349149
- https://neo4j.slack.com/archives/C089SEWEJ8Z/p1741170196254589

We would like to add a CI / CD step that automatically runs the "verify formatting" check that we have been manually running with local files. This check essentially just means running the following integrity checks on a large amount of sample queries (real queries from Aura that we got from the product-analytics people):

1. The AST should not change as a result of formatting
1. Formatting should be idempotent
1. Formatting should not change the amount of non-whitespace characters in a query

This integrity check should happen on any change to the src/formatting folder (see the new file `formatting-integrity-check.yaml`)

### Why has this not been introduced earlier?

1. The sample queries were not sufficiently anonymized; they still contained node and relationship names
3. The sample queries contained lots of duplicates, meaning any attempt would waste a lot of compute looking at queries that are identical in terms of their AST (and thus not very interesting to verify)
4. Storing a large blob of test data directly in git seemed unreasonable, and we did not want to add complexity to the project by e.g. having something on s3.

### Why is it being introduced now; how have we solved / changed our opinions on these issues?

1. By running an "anonymizing visitor" on the queries I was able to remove any custom node/relationship names
2. Through the same visitor as above I could replace all of them with `aaaaa` and deduplicate queries by matching strings
3. We came to the conclusion that storing a reasonably sized (< 10 mb) file should not pose any significant problems with things such as cost. 
   - s3 would of course be cheaper, but for <  10 MB the costs are negligible anyways; 1 GB on Git LFS costs $0.07 per month, with an additional $0.0875 per GiB transfer charge. If we assume that those 10 mb are transferred 1000 times each month (I'm assuming github charges for the data it needs to transfer for its action runners?), they will cost us an additional 0.01GB*1000*$0.0875/GB = $0.875 per month.
       - Note that this PR does not use Git LFS, I just couldn't find a reference for what GH charges for storing regular files so I used the Git LFS numbers for reference. 

### Testing
- I introduced a meaningless change in the formatting visitor to break the tests, and it failed with a nice output: https://github.com/neo4j/cypher-language-support/actions/runs/13951159540/job/39050765538
    - (see the commit ["introduce dumb change to check that it fails"](https://github.com/neo4j/cypher-language-support/pull/416/commits/58f2aa6936aa883561cade8e273eaddb595d8256)
- I introduced a meaningless change that does not break the formatter and it passed nicely: https://github.com/neo4j/cypher-language-support/actions/runs/13951484722/job/39051899077
    - see the commit ["add meaningless change to verify happy path"](https://github.com/neo4j/cypher-language-support/pull/416/commits/15eada081f46cefe1cf7ecea9dbd4d27c0c5d126)
- I removed both of those and the verification run did not happen


package.json
```diff
@@ -21,6 +21,7 @@
     "watch": "turbowatch ./turbowatch.ts",
     "test": "turbo run test",
     "test:e2e": "turbo run test:e2e",
+    "test:formattingIntegrity": "ts-node ./packages/language-support/src/tests/formatting/verification/verificationCheck.ts",
```
tomasnyberg, line 24: I'm not sure if there's a more idiomatic way to run a file as a script? Or maybe it shouldn't be run as a script at all, maybe there's a better way? Happy to take any suggestions here.

packages/language-support/src/tests/formatting/verification/verificationCheck.ts
```diff
@@ -0,0 +1,53 @@
...
+
+function throwError(message: string, query: string, formatted: string): never {
+  throw new Error(`${message},
+---------   QUERY BEFORE START  ------------
+${query}
+---------   QUERY BEFORE END    ----------
+
+---------   QUERY FORMATTED START  ------------
+${formatted}
+---------   QUERY FORMATTED END    ----------
```
tomasnyberg, line 14: This leaves some pretty nice output which should make it easy to debug, see this one for instance:

https://github.com/neo4j/cypher-language-support/actions/runs/13951159540/job/39050765538

.github/workflows/ciformatting.yaml
```diff
@@ -0,0 +1,31 @@
...
+    branches:
+      - main
+    paths:
+      - 'packages/language-support/src/formatting/**'
+
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'packages/language-support/src/formatting/**'
```
OskarDamkjaer, line 12: Do we need both PRs and pulls? I think only pull requests would be good enough. Perhaps we'd also want it as part of the release checks?

.github/workflows/ciformatting.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  formatting-integrity-check:
```
OskarDamkjaer, line 15: Could we change the file name to match the job name instead of `ciformatting`

package.json
OskarDamkjaer: In other cases, I've made the script build js and then execute it. We've also needed to use `cross-env` for some cases (see the `gen-parser` script, although being on mac I don't really know why it's needed 😅 )

OskarDamkjaer: For the record I think it's fine to leave as is here

OskarDamkjaer (approved):
I think it looks good (with a few small tweaks). Let's see what @ncordon thinks 😄

.github/workflows/ciformatting.yaml
tomasnyberg: Any particular reason why you think it shouldn't run on both PRs and commits? My reasoning was it might help catch cases where a check on a PR is outdated, and I don't think it's too expensive. it's also faster than e2e tests

OskarDamkjaer: I just think it's a bit wasteful, but feel free to leave it in

tomasnyberg: @ncordon  What's your opinion on this?

package.json
ncordon: `cross-env` is only to use environment variables (we cannot set them for builds on Windows otherwise)

ncordon (approved):
It looks good to me but just so you know the file is not stored with lfs, which you could and I don't know if that was the intention given the description of the PR?

tomasnyberg:
> It looks good to me but just so you know the file is not stored with lfs, which you could and I don't know if that was the intention given the description of the PR?

Yeah I couldn't find what the exact pricing was for regular objects in a git repo so I used the LFS numbers as a reference. The assumption I'm making is that the price is comparable

package.json
OskarDamkjaer: If we move to `node22` we can run it without ts-node as well with `--experimental-strip-types`

PR #417, opened 2025-03-20
Add `:style` cmd to command parser

OskarDamkjaer:
This is to let users show a frame to set/change the style of their graph


daveajrussell:
observation - with this change we get both `:style` and `:style reset` in the autocompletions.

do we want this? or do we want it to behave more like the `:server` command, where we only get the one `:server` autocompletion.. and _then_ we get argument autocompletions for `connect` and `disconnect`

I wonder if we follow the same pattern, so we just get `:style` in the autocompletions - which can be used to launch the style frame, and optionally we can pass an argument to it like `reset`, or any others in the future

https://github.com/neo4j/cypher-language-support/pull/293 for example

if we change `CypherCmdParser.g4` to something like

```
styleCmd: styleCompletionRule styleArgs?;

styleArgs: RESET;

styleCompletionRule: STYLE;
```

then in `parserWrapper.ts` change to something like

```
      const styleCmd = consoleCmd.styleCmd();
      const styleArgs = styleCmd?.styleArgs();
      if (styleCmd) {
        if (styleArgs) {
          const reset = styleArgs.RESET();
          if (reset) {
            return { type: 'style', operation: 'reset', start, stop };
          }
        }

        return {
          type: 'style',
          start,
          stop,
        };
      }
```

then there's a few other additions in `syntaxColouring.ts`, and `completionCoreErrors.ts` and we should get the single `:style` autocompletion, with the argument autocompletions separate from the main command;

Screenshot 2025-03-20 at 08 50 27

Screenshot 2025-03-20 at 08 50 32

daveajrussell (approved):

OskarDamkjaer:
@daveajrussell I agree it's nicer to do the completions in two steps in this case and I think that is what the completion engine would do by default. @ncordon extended it in [this pr](https://github.com/neo4j/cypher-language-support/pull/59/files). The idea is I think to avoid the "need" to do one completion and then another completion with only one option 

CleanShot 2025-03-20 at 10 08 26@2x

The change you outlined is effectively working around an extension to the completion engine we've made ourselves, so I think it makes more sense to remove our extensions if we don't want it.

PR #418, opened 2025-03-20
Improves CASE inside CALL clause

simonthuresson:
### Description
During #411 I noticed that `CASE` statements inside a `CALL` clause does not get the base indentation unlike `EXISTS` does. This PR improves `CASE` expression inside a `CALL` clause by adding correct base indentation. 

### Tests
 - One test added
 - All tests passes


tomasnyberg (approved):
lgtm

PR #420, opened 2025-03-21
IntelliJ plugin misc fixes

jamthief:
- Add language description
- Add file template description
- Make supported future versions open ended

### DIFF
--- a/editor-plugin/intellij/build.gradle.kts
+++ b/editor-plugin/intellij/build.gradle.kts
@@ -35,6 +35,15 @@ tasks {
         }
     }
 
+    runIde {
+        debugOptions {
+           enabled = false
+           port = 8000
+           server = true
+           suspend = true
+       }
+    }
+
     prepareSandbox {
         doFirst {
             exec {
@@ -45,11 +54,6 @@ tasks {
             include("*.js")
             into("cypher-lsp-support")
         }
-    }
-
-    patchPluginXml {
-        sinceBuild.set("242")
-        untilBuild.set("242.*")
     }
 
     signPlugin {
--- a/editor-plugin/intellij/src/main/java/org/neo4j/intellij/lsp/language/CypherLanguage.java
+++ b/editor-plugin/intellij/src/main/java/org/neo4j/intellij/lsp/language/CypherLanguage.java
@@ -1,11 +1,17 @@
 package org.neo4j.intellij.lsp.language;
 
 import com.intellij.lang.Language;
+import org.jetbrains.annotations.NotNull;
 
 public class CypherLanguage extends Language {
-
     public static final CypherLanguage INSTANCE = new CypherLanguage();
+
     private CypherLanguage() {
         super("Cypher");
+    }
+
+    @Override
+    public @NotNull String getDisplayName() {
+        return "Cypher";
     }
 }
--- a/editor-plugin/intellij/src/main/resources/META-INF/plugin.xml
+++ b/editor-plugin/intellij/src/main/resources/META-INF/plugin.xml
@@ -1,6 +1,9 @@
 <idea-plugin>
     <id>org.neo4j.intellij.lsp.cypher-lsp-support</id>
 
+    <!-- see https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html -->
+    <idea-version since-build="242.22855.74"/>
+
     <name>Cypher Lsp Support</name>
 
     <vendor url="https://www.neo4j.com">Neo4j Inc.</vendor>
@@ -17,11 +20,13 @@
     <extensions defaultExtensionNs="com.intellij">
         <platform.lsp.serverSupportProvider
                 implementation="org.neo4j.intellij.lsp.cypherlspsupport.CypherLspServerSupportProvider"/>
+
         <fileType
                 name="Cypher File"
-                implementationClass="org.neo4j.intellij.lsp.language.CypherFileType"
-                fieldName="INSTANCE"
                 language="Cypher"
-                extensions="cypher"/>
+                implementationClass="org.neo4j.intellij.lsp.language.CypherFileType"
+                extensions="cypher"
+                fieldName="INSTANCE"
+        />
     </extensions>
 </idea-plugin>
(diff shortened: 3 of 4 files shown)


turbo.json
```diff
@@ -6,6 +6,15 @@
...
     },
+    "build:intellij": {
+      "dependsOn": ["build"],
+      "outputs": ["packages/intellij-extension/build/**"]
+    },
+    "run:intellij": {
+      "cache": false,
+      "persistent": true,
+      "dependsOn": ["build:intellij"]
+    },
```
jamthief, line 17: Can't seem to get `run:intellij` to force a rebuild of the plugin when the language-server files have changed. Is turbo cacheing the build output somewhere?

jamthief:
The E2E check looks flaky, but I can't find a way to re-run it. Maybe I'm missing the permission?

ncordon (changes requested):
I'm not sure why, I get this when I do `npm run build`:

```
2025-04-07 13:17:45,255 [   9534]   WARN - #c.i.k.a.KubernetesApiProvider - Cannot run program "kubectl": error=2, No such file or directory
com.intellij.execution.process.ProcessNotCreatedException: Cannot run program "kubectl": error=2, No such file or directory
        at com.intellij.execution.configurations.GeneralCommandLine.createProcess(GeneralCommandLine.java:384)
        at com.intellij.execution.process.OSProcessHandler.startProcess(OSProcessHandler.java:85)
```

What's Kubernetes used in the build exactly? Is it that sandbox we've added? It works after installing `kubectl` but not sure why it wasn't before.

In main executing the build of the plugin was working fine for me.

packages/intellij-extension/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "neo4j-for-intellij",
+  "version": "1.0.0",
+  "description": "Highlighting, completions and more for Neo4j Cypher in IntelliJ",
+  "publisher": "neo4j-extensions",
+  "author": "Neo4j Inc.",
+  "license": "Apache-2.0",
+  "preview": true,
+  "scripts": {
+    "build": "./gradlew -q buildPlugin",
```
ncordon, line 10: I'd prefer to have this as:

```
if [[ \"$BUILD_INTELLIJ\" == \"true\" ]]; then ./gradlew -q buildPlugin; else echo 'Skipping Intellij build, set BUILD_INTELLIJ=true if that is not the desired behaviour'; fi
```

So this does not get executed for the people that work currently on the repo unless they specifically have that env variable set. The reason is we have 4 / 5 people that contribute regularly right now from the company, some of them using OSX, some using Windows, so I don't want to force them to install Kubernetes locally.

ncordon: @anderson4j can you try this works for you?

.gitignore
```diff
@@ -17,6 +17,49 @@ generated
...
+/.nb-gradle/
+
+### VS Code ###
+.vscode/
+
+### Mac OS ###
+.DS_Store
+
+### Javascript files
+*.js
```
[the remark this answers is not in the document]
jamthief: Sorry, this was all unnecessarily copied from the inner .gitignore. I'll scale it back!

packages/intellij-extension/package.json
jamthief: @ncordon I think I get your gist - although it needs a single equals sign to work: `[[ \"$BUILD_INTELLIJ\" = \"true\" ]]` - unfortunately though, this gets cached by turbo. So running it with a different value will just do the same thing over and over, even if you change the environment variable.

However, in principle, I want to argue against doing this anyway. My reasoning is as follows:

- The VS code extension and the IntelliJ plugin are both at identical positions in the build hierarchy: they are dependent on the language server build, and they are leaf node outputs. Running a generalised `build` at the top level should be the equivalent of something like `make all`; it should _attempt_ to build everything by default.
- If a developer has a workflow involving one or the other plugin, they should use the specialised build target
- If running the whole build doesn't work in CI, this PR is incomplete
- If we don't attempt to build it regularly, and fail the build when it also fails, then we won't know when we've broken it, and the plugin will become stale

jamthief:
> I'm not sure why, I get this when I do `npm run build`:
> 
> ```
> 2025-04-07 13:17:45,255 [   9534]   WARN - #c.i.k.a.KubernetesApiProvider - Cannot run program "kubectl": error=2, No such file or directory
> com.intellij.execution.process.ProcessNotCreatedException: Cannot run program "kubectl": error=2, No such file or directory
>         at com.intellij.execution.configurations.GeneralCommandLine.createProcess(GeneralCommandLine.java:384)
>         at com.intellij.execution.process.OSProcessHandler.startProcess(OSProcessHandler.java:85)
> ```
> 
> What's Kubernetes used in the build exactly? Is it that sandbox we've added? It works after installing `kubectl` but not sure why it wasn't before.
> 
> In main executing the build of the plugin was working fine for me.

It's pretty gross, but also harmless. The build still passes. I see it too when I forcibly rename the `kubectl` binary to something else. Did you have kubectl installed before? I wonder if it thinks you have kubernetes support and so tries to run it.

jamthief:
@ncordon

> I'm not sure why, I get this when I do `npm run build`:
> 
> ```
> 2025-04-07 13:17:45,255 [   9534]   WARN - #c.i.k.a.KubernetesApiProvider - Cannot run program "kubectl": error=2, No such file or directory
> com.intellij.execution.process.ProcessNotCreatedException: Cannot run program "kubectl": error=2, No such file or directory
>         at com.intellij.execution.configurations.GeneralCommandLine.createProcess(GeneralCommandLine.java:384)
>         at com.intellij.execution.process.OSProcessHandler.startProcess(OSProcessHandler.java:85)
> ```
> 
> What's Kubernetes used in the build exactly? Is it that sandbox we've added? It works after installing `kubectl` but not sure why it wasn't before.
> 
> In main executing the build of the plugin was working fine for me.

It's pretty gross, but also harmless. The build still passes. I see it too when I forcibly rename the `kubectl` binary to something else. Did you have kubectl installed before? I wonder if it thinks you have kubernetes support and so tries to run it.

PR #421, opened 2025-03-21
Improves formatting of long patterns (split before nodes, not relations)

tomasnyberg:
## Description
Credit to @arnefischereit for suggesting this :)!

After our presentation during the cypher iteration demo this morning, we got the feedback that long patterns should generally prefer to split before nodes, not before relationships.

So this query (which we showcased during the demo), for instance, should not be formatted like this:
```
MATCH (a:person {name: 'alice', age: 30})-[r:friend_of]->
      (b:person {name: 'bob'})-[s:colleague_of]->(c:person {name: 'carol'})-
      [t:partner_of]->(d:person {name: 'david'})-[u:mentor_and_friend_of]->
      (e:person {name: 'eve'})
RETURN a
```
but rather like this:

```
MATCH (a:person {name: 'alice', age: 30})-[r:friend_of]->
      (b:person {name: 'bob'})-[s:colleague_of]->
      (c:person {name: 'carol'})-[t:partner_of]->
      (d:person {name: 'david'})-[u:mentor_and_friend_of]->
      (e:person {name: 'eve'})
RETURN a
```

We've gone through a few examples of other long graph patterns and this approach seems way cleaner. 

This PR adds a visitor for `patternElement` to group [node, relationship, QPP] together, which will result in the solution search always preferring to split before a node rather than a relationship if possible.

### Warning: messy visitor
The grammar clause for patternElements looks like this:

```
patternElement
   : (nodePattern (relationshipPattern quantifier? nodePattern)* | parenthesizedPath)+
   ;
```
which is problematic, because we do not know where the quantifiers appear. Because of this, we have to manually traverse the children and visit them based on their types. The fact that we want the grouping as described above makes it even more annoying, leading to some quite convoluted code.

#### Typecasts
The code does, regrettably, include a few typecasts. The motivation behind them is that where they are used, the code looks like this:

```
if(typeCheck(getX())) {
  typedcall(getX() as Type) // Type error without the cast
}
```
The only reason typescript can't figure out that this is fine is because it can't be certain that `getType` is a pure function. IMO since the typecheck is right above it it's fine.

## Testing
- Four new tests added that are similar to the one above

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -123,7 +123,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   cursorPos = 0;
   groupID = 0;
   groupStack: number[] = [];
-  startGroupCounter: number = 0;
+  startGroupCounter = 0;
   groupsToEndOnBreak: number[] = [];
   previousTokenIndex: number = -1;
   unParseable: string = '';
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -1270,7 +1270,6 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
             ctx.getChild(i) as RelationshipPatternContext,
           );
           i++;
-          this.endGroup(nodeRelPatternGrp);
 
           // Optionally, visit a quantifier if present.
           if (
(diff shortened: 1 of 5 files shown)


packages/language-support/src/formatting/formatting.ts
```diff
@@ -1239,6 +1240,69 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+  // and the grammar is very unhelpful...
+  _processNodeRelSequence = (
+    ctx: PatternElementContext,
+    startIndex: number,
+    n: number,
+  ): number => {
+    let i = startIndex;
+    let nodeRelPatternGrp = this.startGroup();
+
+    this.visitNodePattern(ctx.getChild(i) as NodePatternContext);
```
tomasnyberg, line 1277: There's a few typecasts here because getting rid of them made the code a lot more messy; see the description. If you find a great way to refactor this I'm happy to change it though

PR #422, opened 2025-03-24
Fixes parameters not being available on disconnect in VSCode

ncordon:
When disconnecting from the VSCode extension, we want the parameters to still be available for completions, because it is a bit weird they are present in the side pane but cannot be used.

### DIFF
--- a/packages/language-support/src/autocompletion/completionCoreCompletions.ts
+++ b/packages/language-support/src/autocompletion/completionCoreCompletions.ts
@@ -352,7 +352,9 @@ const parameterCompletions = (
         // We need to have the insert text without the starting $ in VSCode,
         // otherwise when we have 'RETURN $' and we get offered $param
         // we would complete RETURN $$param, which is not what we want
-        maybeInsertText = suffix ? { insertText: paramName } : {};
+        maybeInsertText = suffix
+          ? { insertText: param.slice(1) }
+          : maybeInsertText;
       }
 
       return {
--- a/packages/schema-poller/src/schemaPoller.ts
+++ b/packages/schema-poller/src/schemaPoller.ts
@@ -140,7 +140,10 @@ export class Neo4jSchemaPoller {
     this.connection?.dispose();
     this.metadata?.stopBackgroundPolling();
     this.connection = undefined;
-    this.metadata = undefined;
+    if (this.metadata) {
+      const parameters = this.metadata.dbSchema.parameters;
+      this.metadata.dbSchema = { parameters: parameters };
+    }
     this.driver = undefined;
     clearTimeout(this.reconnectionTimeout);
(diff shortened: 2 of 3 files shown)


packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -352,7 +352,9 @@ const parameterCompletions = (
         // We need to have the insert text without the starting $ in VSCode,
         // otherwise when we have 'RETURN $' and we get offered $param
         // we would complete RETURN $$param, which is not what we want
-        maybeInsertText = suffix ? { insertText: paramName } : {};
+        maybeInsertText = suffix
+          ? { insertText: param.slice(1) }
+          : maybeInsertText;
```
ncordon, line 357: This was not backticking the parameters with a space for example, because we were using the param name directly, not the backticked one.

`param` contains the backticked one, if it applies, so we just need to remove the initial `$`

packages/schema-poller/src/schemaPoller.ts
```diff
@@ -140,7 +140,10 @@ export class Neo4jSchemaPoller {
     this.connection?.dispose();
     this.metadata?.stopBackgroundPolling();
     this.connection = undefined;
-    this.metadata = undefined;
+    if (this.metadata) {
+      const parameters = this.metadata.dbSchema.parameters;
+      this.metadata.dbSchema = { parameters: parameters };
+    }
```
ncordon, line 146: This fixes the parameters not being available in VSCode

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -348,11 +349,16 @@ const parameterCompletions = (
...
+        const hasDollar =
+          previousToken.type === CypherLexer.DOLLAR ||
+          tokens.at(previousToken.tokenIndex - 1).type === CypherLexer.DOLLAR;
+        // If the $ symbol is already there, we need to have the insert
+        // text without the starting $ in VSCode, otherwise when we have
+        // 'RETURN $' and we get offered $param we would complete
+        // RETURN $$param, which is not what we want
+        maybeInsertText = hasDollar
+          ? { insertText: param.slice(1) }
+          : maybeInsertText;
```
ncordon, line 361: Readjusts this part to fix the completions on the VSCode side. If we have started the dollar part, we need the insert text to come without the `$`, otherwise we would end up with double dollars.

ncordon:
> But the solution doesnt quite feel right to me, because though it works as I would expect when you disconnect - it doesnt work if you start up with no connection.

Nice spot! This should be fixed now

anderson4j (approved):
Nice work!

PR #423, opened 2025-03-24
Adds parameter deletion ability in VSCode

ncordon:
This adds both a command and a context menu right click option to delete parameters individually.

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -99,6 +99,11 @@
         "title": "Clear all parameters",
         "category": "Neo4j",
         "icon": "$(trash)"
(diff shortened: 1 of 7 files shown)


anderson4j (approved):
LGTM

anderson4j:
Actually, one thing, don't forget to add changeset

ncordon:
> Actually, one thing, don't forget to add changeset - unless maybe this comes under something else like "adding parameter panel" in the changelog?

I think that changelog is enough for users to understand the whole feature has been shipped, as long as we don't ship a panel that only has partial support to work with the parameters (addition but not deletion)

PR #431, opened 2025-03-26
Adds editing of params

anderson4j:
WIP

### DIFF
--- a/packages/vscode-extension/src/commandHandlers/params.ts
+++ b/packages/vscode-extension/src/commandHandlers/params.ts
@@ -15,6 +15,7 @@ import { getSchemaPoller } from '../contextService';
 import {
   clearParameters,
   deleteParameter,
+  getParameter,
   setParameter,
 } from '../parameterService';
 import {
@@ -22,9 +23,13 @@ import {
   parametersTreeDataProvider,
 } from '../treeviews/parametersTreeDataProvider';
 
-export async function addParameter(): Promise<void> {
+export async function isConnected(): Promise<boolean> {
   const schemaPoller = getSchemaPoller();
-  const connected = await schemaPoller.connection?.healthcheck();
+  return schemaPoller.connection?.healthcheck();
+}
+
+export async function addParameter(): Promise<void> {
+  const connected = await isConnected();
 
   if (!connected) {
     await window.showErrorMessage(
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/src/commandHandlers/params.ts
```diff
@@ -55,6 +60,27 @@ export async function addParameter(): Promise<void> {
   await evaluateParam(paramName, paramValue);
 }
 
+export async function modifyParameter(paramItem: ParameterItem): Promise<void> {
```
ncordon, line 63: It'd be good to call this `editParameter` to match the name of the command

packages/vscode-extension/tests/specs/webviews/params.spec.ts
```diff
@@ -46,6 +46,10 @@ suite('Params panel testing', () => {
     );
   }
 
+  async function forceModifyParam(key: string, value: string) {
+    await forceAddParam(key, value);
```
ncordon, line 50: Add a comment on top to explain why we are calling add param

packages/vscode-extension/src/commandHandlers/params.ts
```diff
@@ -55,6 +60,27 @@ export async function addParameter(): Promise<void> {
...
+  if (!existingParam) {
+    return;
+  }
+  const paramValue = await window.showInputBox({
+    prompt: 'Parameter value',
+    value: existingParam.evaluatedStatement,
+    ignoreFocusOut: true,
+  });
+
+  await evaluateParam(paramItem.id, paramValue);
```
ncordon, line 81: Similar to the add parameter functionality, we should add the check:

```ts
  if (!paramValue) {
    await window.showErrorMessage('Parameter value cannot be empty.');
    return;
  }
```

before the `evaluateParam` call.

ncordon: I also realized the `if (!paramName)` and `if(!paramValue)` checks in the `addParameter()` method are missing the early return, so fix that in the PR 🙏.

Probably that's why we were saying `wrong format for the parameter` messages pop up when it didn't make sense.

packages/vscode-extension/tests/specs/webviews/params.spec.ts
```diff
@@ -93,6 +97,34 @@ suite('Params panel testing', () => {
     });
   });
 
+  test.only('Should correctly modify cypher parameters', async function () {
```
ncordon, line 100: Don't forget to remove the `test.only` here please

ncordon (commented):
I left some feedback.

Please write a proper PR title and description and make sure to refetch the branch, I've added myself as co-author of the commit.

ncordon (approved):

PR #437, opened 2025-03-26
Adds check to formatter that ensures correct group starts/ends

tomasnyberg:
## Description
There are a few queries that result in groups being started but never ended, or vice versa. This should never happen if all visit methods are correct, though because of some complicated cases (pun intended) it still happens.

This PR fixes all of those cases, and additionally adds adds two checks:
1. We should never end a group if there is not a group active
2. There should not be any active groups left when we are done formatting

If either of these are true, **the formatter suffers an internal error**.

### Testing
- No new tests added, but all tests pass without triggering the above checks.

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -152,7 +152,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   format = () => {
     this._visit(this.root);
     // There may be some trailling groups that are left to achieve indentation
-    this.removeAllActiveGroups();
+    this.removeAllPendingGroups();
     const result = buffersToFormattedString(this.buffers);
     this.cursorPos += result.cursorPos;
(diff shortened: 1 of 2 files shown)


ncordon (commented):
Are we doing this because of a bug? It would be nice to add a test of something that broke before and doesn't now?

tomasnyberg:
> Are we doing this because of a bug?

Not really, it's more of a tech debt thing. Think of it like removing unnecessary lines of code that have no effect on the actual result.

The reason we did this now is because getting the V3 line break behavior is quite difficult if there are incorrect groups. Imagine a group decides to split between all children but then never ends for instance. It's not noticeable in V2 since groups just steer the cost function.

> It would be nice to add a test of something that broke before and doesn't now?

I haven't added any new tests, but there were about 20 or so tests that failed due to the check before the fixes in this PR. So there's no new tests but it does "fix" them

ncordon (approved):

PR #445, opened 2025-03-28
Adjusts job to update grammar and semantic analysis

ncordon:
Bear in mind the database is cut off 

This PR contains changes to:

* Fix the cron job expression we were using, which was executing everyday from the 1st day on the month to the 7th and on Fridays. We wanted something that was executing only the first Friday of every month, but that does not seem possible to get with cron, so we have settled for executing the 4th day of the month and the 7th day of the month. 

In the worst case scenario the 1st day of the month is a Friday so we would get the update on the Sunday (3rd day), leaving us with a whole week to do the changes. If the 1st day of the month was a Saturday, then we would get this update on the first Friday, the 6th.

* Uses a PAT token to make the PRs execute the CI when they are opened, rather than the default `GITHUB_TOKEN`

### DIFF
--- a/.github/workflows/fetch-dev-artifacts.yaml
+++ b/.github/workflows/fetch-dev-artifacts.yaml
@@ -6,7 +6,7 @@ permissions:
 
 on:
   schedule:
-    - cron: '10 4 1-7 * 5'
+    - cron: '10 4 4,7 * *'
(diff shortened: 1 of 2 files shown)


anderson4j (approved):
LGTM

PR #447, opened 2025-03-31
adding support for :play command

nnecla:
This pr is to add support for the `:play` command, which is used to play custom guides within the query application.


anderson4j (approved):
I think we should add :play to the "parses simple commands without args"- and "completes basic console cmds on :"-tests too.

other than that LGTM, so I'll do a preemptive approve

PR #450, opened 2025-04-01
Modifies fetch-dev-artifacts to only fetch on the first friday

anderson4j:
This way the workflow will be scheduled to run on all fridays, but checking if it's the first friday of the month before running any fetching

### DIFF
--- a/.github/workflows/fetch-dev-artifacts.yaml
+++ b/.github/workflows/fetch-dev-artifacts.yaml
@@ -20,7 +20,7 @@ jobs:
         id: check_skip
         run: |
           DAY=$(date +'%d')
-          if [ "$DAY" -ge 8 ] && [ "${{ github.event_name }}" == "schedule" ]; then
+          if [[ "$DAY" -ge 8 ]] && [[ "${{ github.event_name }}" == "schedule" ]]; then
             echo "Skipping fetch"
             echo "skip=true" >> $GITHUB_OUTPUT
           else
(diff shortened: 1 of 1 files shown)


.github/workflows/fetch-dev-artifacts.yaml
```diff
@@ -6,14 +6,33 @@ permissions:
...
+  check-day-of-month:
+    runs-on: ubuntu-latest
+    outputs:
+      skip: ${{ steps.check_skip.outputs.skip }}
+    steps:
+      - name: Check date if run on schedule
+        id: check_skip
+        run: |
+          DAY=$(date +'%d')
+          if [ "$DAY" -ge 8 ] && [ "${{ github.event_name }}" == "schedule" ]; then
```
anderson4j, line 23: I added the event_name == "schedule" check before setting skip=true so we could still run it manually any day

anderson4j:
This way the history of runs will contain a run every friday that skips everything but date check (except first friday), so job execution history will get 2 extra entries per month compared to current way. 

What we gain is that only 1 PR is created exactly a week before the release we want it done for. Also if you look at each run of the job, it's clear when the fetching is skipped, so for me the history part is not too bad.

.github/workflows/fetch-dev-artifacts.yaml
ncordon, line 23: Normally the preferred syntax in bash is double square brackets `[[` and `]]`.

`$(date + '%d')` seems to be padding with an initial 0. Are you sure this works as intended?

anderson4j: Yep, I tested it on my local repo

ncordon (approved):

PR #451, opened 2025-04-01
Prepares VSCode release

ncordon:
* Gets out of the preview mode (no more red preview banner on the VSCode marketplace)
* Reviews the VSCode extension `README`
* Corrects a few small bugs in the VSCode extension `package.json` (we could not use the embedded menus to create a cypher file from a selection in Java for example) or we were showing `Neo4j: Neo4j Run Cypher statements` in the command palette (duplicated `Neo4j`)

### DIFF
--- a/packages/vscode-extension/README.md
+++ b/packages/vscode-extension/README.md
@@ -85,17 +85,17 @@ Clicking on an instance will automatically connect to it. You can also manage a 
 ![demo-connect](https://github.com/neo4j/cypher-language-support/blob/main/packages/vscode-extension/resources/images/demo-connect.png?raw=true)
 ![demo-disconnect](https://github.com/neo4j/cypher-language-support/blob/main/packages/vscode-extension/resources/images/demo-disconnect.png?raw=true)
 
-Once a connection is added, and a connection to your Neo4j instance is established, you will be able to use database aware features of the extension.
+Once a connection is added, and a connection to your Neo4j instance is established, you will be able to use database aware features of the extension. You can change the database you want to run queries against by clicking or right clicking (`Switch to database`) on it.
 
 You can store connections to different Neo4j instances.
 
-## Managing parameters
+## Managing parameters
 
 To add a new parameter, you can use the `Neo4j: Add parameter` command from the Command Palette or use the `+` on the Neo4j parameters pane:
 
 ![demo-add-parameter](https://github.com/neo4j/cypher-language-support/blob/main/packages/vscode-extension/resources/images/demo-add-parameter.png?raw=true)
 
-You'll be prompted for the name and value you want to use for that parameter. The type will be automatically recognized from the value you try and set. Some examples:
+You will be prompted for the name and value you want to use for that parameter. The type will be automatically recognized from the value. Some examples:
 
 ```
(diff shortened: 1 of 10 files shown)


packages/vscode-extension/README.md
```diff
@@ -65,24 +74,56 @@ const d = '/*cypher*/ MATCH (n) RETURN n';
...
 
 ![demo-connect](https://github.com/neo4j/cypher-language-support/blob/main/packages/vscode-extension/resources/images/demo-connect.png?raw=true)
 ![demo-disconnect](https://github.com/neo4j/cypher-language-support/blob/main/packages/vscode-extension/resources/images/demo-disconnect.png?raw=true)
 
-Once a connection is added, and a connection to your Neo4j instance is established, you will be able to use database aware features of the extension.
+Once a connection is added, and a connection to your Neo4j instance is established, you will be able to use database aware features of the extension. You can change the database you want to run queries against by clicking or right clicking (`Switch to database`) on it.
+
+You can store connections to different Neo4j instances.
+
+## Managing parameters
```
anderson4j, line 92: I think it would be nice to mention editing/deleting parameters here too

packages/vscode-extension/README.md
```diff
@@ -65,24 +74,56 @@ const d = '/*cypher*/ MATCH (n) RETURN n';
...
+As a rule of thumb, any `expression` that can appear in a `RETURN expression` can be set as value.
+
+Parameters can be accessed from the code prepending them with a dollar sign `$`.
+
+## Executing a Cypher file
+
+Once you've written your desired query (for example `RETURN $a, $b`) you can execute it by either of:
+* Using the `Neo4j: Run cypher statements` command from the Command Palette.
+* Right clicking inside the file you want to run and clicking on the `Neo4j: Run cypher statements` item.
+* Using the shortcut `Ctrl+Command+Space` (where `Command` is either the `⌘` key or the Windows key depending on your keyboard).
```
anderson4j, line 116: Time for my expertise to show!

anderson4j, line 116:
```suggestion
* Using the shortcut `Ctrl+Command+Space` (where `Command` is either the `⌘` key or the alt key depending on your keyboard).
```

anderson4j (changes requested):
Nice improvements to the readme :) Had some suggestions where I think we could be even more clear, and one minor error that needs fixing.

packages/vscode-extension/README.md
```diff
@@ -65,24 +74,56 @@ const d = '/*cypher*/ MATCH (n) RETURN n';
...
+You will be prompted for the name and value you want to use for that parameter. The type will be automatically recognized from the value. Some examples:
+
+```
+"something" // would set a string parameter
+1234 // would set an Integer parameter
+datetime() // would set a datetime parameter
+[1,2,3] // would set a List parameter
+```
+
+As a rule of thumb, any `expression` that can appear in a `RETURN expression` can be set as value.
```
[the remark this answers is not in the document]
ncordon: We saw how you can have a node in a return but not as a parameter

anderson4j (approved):
Good stuff

PR #452, opened 2025-04-01
triggering a param add pop-up if running a cypher with missing params

nnecla:
This pr is to trigger the parameter add pop-up when a user wants to run a query with missing parameters. With these changes:
- Passing a defaultName to the addParameter and adding param name to the `prompt`
- Changing parameter error messages from `await` to `void` to prevent the wait for user to manually close the error message before opening the add pop-up for the following missing parameter
- Adding a function `extractParameters` which gets an array of strings(queries) and returns distinct paramNames

### DIFF
--- a/packages/vscode-extension/src/cypherRunner.ts
+++ b/packages/vscode-extension/src/cypherRunner.ts
@@ -1,7 +1,31 @@
 import { parseStatementsStrs } from '@neo4j-cypher/language-support';
 import { Uri } from 'vscode';
+import { addParameter } from './commandHandlers/params';
 import { Connection } from './connectionService';
+import { getDeserializedParams } from './parameterService';
 import ResultWindow from './webviews/resultWindow';
+
+export function extractParameters(strings: string[]): string[] {
+  const paramSet = new Set<string>();
+
+  strings.forEach((str) => {
+    const matches = str.match(/\$`([^`]+)`|\$(\w+)/g);
+    if (matches) {
+      matches.forEach((match) => {
+        // Remove the '$'
+        const param = match.slice(1);
+        if (param.startsWith('`') && param.endsWith('`')) {
+          // Remove the backticks
+          paramSet.add(param.slice(1, -1));
+        } else {
+          paramSet.add(param);
+        }
+      });
+    }
+  });
+
+  return Array.from(paramSet);
+}
 
 export default class CypherRunner {
   private results: Map<string, ResultWindow> = new Map();
@@ -11,6 +35,14 @@ export default class CypherRunner {
   async run(connection: Connection, uri: Uri, input: string) {
     const statements = parseStatementsStrs(input);
     const filePath = uri.toString();
+    const parameters = getDeserializedParams();
+    const statementParams = extractParameters(statements);
+
+    for (const param of statementParams) {
+      if (parameters[param] === undefined) {
+        await addParameter(param);
+      }
+    }
 
     if (this.results.has(filePath)) {
       const resultWindow = this.results.get(filePath);
--- a/packages/vscode-extension/tests/specs/webviews/params.spec.ts
+++ b/packages/vscode-extension/tests/specs/webviews/params.spec.ts
@@ -1,6 +1,7 @@
 import { browser } from '@wdio/globals';
 import { before } from 'mocha';
 import { Workbench } from 'wdio-vscode-service';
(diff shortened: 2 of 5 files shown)


packages/vscode-extension/tests/specs/webviews/params.spec.ts
```diff
@@ -187,11 +202,45 @@ suite('Params panel testing', () => {
     // to execute the file we need to be on the user database
     await forceSwitchDatabase('neo4j');
     await executeFile(workbench, 'params.cypher');
+
+    // to close the automatic param add pop-ups
+    for (let i = 0; i < 4; i++) {
+      await browser.keys([Key.Escape]);
+      await browser.pause(1000);
```
anderson4j, line 209: Felt like we had something to handle similar situations in another test, and I found it! To not have to wait an arbitrary time, we should be able to wait until the notification that parameter setting failed by instead doing:
      await waitUntilNotification(browser, 'Parameter value cannot be empty.');
      
 (Looks like you have the same pause several times later, but I wont add a comment on each one)

anderson4j (changes requested):
Good stuff! I think I found an improvement to speed up tests (or for a computer from the 1900s maybe make it less flakey), but other than that I think it's good to go

packages/vscode-extension/src/cypherRunner.ts
```diff
@@ -1,8 +1,32 @@
...
+          paramSet.add(param.slice(1, -1));
+        } else {
+          paramSet.add(param);
+        }
+      });
+    }
+  });
+
+  return Array.from(paramSet);
+}
```
ncordon, line 28: I think it'd make more sense to have a small method in the language support that returns the parameters using the parsing?

We have a data type:

```
export interface ParsedStatement {
  command: ParsedCommand;
  parser: CypherParser;
  tokens: Token[];
  // A statement needs to be parsed with the .statements() rule because
  syntaxErrors: SyntaxDiagnostic[];
  cypherVersionError: SyntaxDiagnostic | undefined;
  stopNode: ParserRuleContext;
  collectedLabelOrRelTypes: LabelOrRelType[];
  collectedVariables: string[];
  collectedFunctions: ParsedFunction[];
  collectedProcedures: ParsedProcedure[];
  cypherVersion?: CypherVersion;
}
```

and we use visitors to collect variables / functions / procedures used in the query, so worth looking into that.

If we add something like `parameters` that would make showing errors for those parameters that are not found in the schema easier.

nnecla: Makes sense! I will continue with this pr after I merge #456 and try to use `collectedParameters` instead of this method.

packages/language-support/src/index.ts
```diff
@@ -7,7 +7,12 @@ export { _internalFeatureFlags } from './featureFlags';
 export { formatQuery } from './formatting/formatting';
 export { antlrUtils } from './helpers';
 export { CypherTokenType, lexerSymbols } from './lexerSymbols';
-export { parse, parserWrapper, parseStatementsStrs } from './parserWrapper';
+export {
+  createParsingResult,
```
ncordon, line 11: Maybe it'd be better to have a `parseParameters` method in the language support rather than exposing this internal? Then we can move the `getClearParamName` method there?

nnecla: Changed it to have a parseParameters which returns an array of strings(cleared param names)

ncordon (approved):

ncordon (commented):
Really noice!

packages/vscode-extension/tests/specs/webviews/params.spec.ts
```diff
@@ -14,6 +15,14 @@ suite('Params panel testing', () => {
     workbench = await browser.getWorkbench();
   });
 
+  async function escapeModal(count: number) {
+    for (let i = 0; i < count; i++) {
+      await browser.pause(500);
+      await browser.keys([Key.Escape]);
+      await waitUntilNotification(browser, 'Parameter value cannot be empty.');
+    }
+  }
```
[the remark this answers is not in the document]
nnecla: Yes, actually, it was flaky. So, I'd like to keep these just in case, so that the browser wouldn't click on the `Escape` button before the parameter add pop-up opens :/

anderson4j (approved):

PR #453, opened 2025-04-02
Adds debouncing to parameter validation

anderson4j:
Uses a timeout to prevent parameter validation from running until the user has not changed input for 200 ms.

Previously every keystroke would cause validation to run and change the validation message, but now if you type quite quickly it will wait for a stop in typing.

### DIFF
--- a/packages/vscode-extension/src/commandHandlers/params.ts
+++ b/packages/vscode-extension/src/commandHandlers/params.ts
@@ -1,5 +1,4 @@
 import { DbSchema, lintCypherQuery } from '@neo4j-cypher/language-support';
-import { DEBOUNCE_TIME } from '@neo4j-cypher/react-codemirror';
 import {
   cypherDataToString,
   CypherDataType,
@@ -9,7 +8,6 @@ import {
   Neo4jType,
   serializeTypeAnnotations,
 } from '@neo4j-cypher/schema-poller';
-import { debounce } from 'lodash';
 import { Neo4jError } from 'neo4j-driver';
 import { window } from 'vscode';
 import { DiagnosticSeverity } from 'vscode-languageclient';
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/commandHandlers/params.ts
```diff
@@ -1,4 +1,5 @@
 import { DbSchema, lintCypherQuery } from '@neo4j-cypher/language-support';
+import { DEBOUNCE_TIME } from '@neo4j-cypher/react-codemirror/src/constants';
```
ncordon, line 2: We shouldn't be importing from `src` folders. And it's not worth it to add the entire `react-codemirror` package as a dependency so just duplicate the variable

packages/vscode-extension/src/commandHandlers/params.ts
```diff
@@ -66,12 +68,21 @@ export async function addParameter(): Promise<void> {
...
     prompt: 'Parameter value',
     placeHolder:
       'The value for your parameter (anything you could evaluate in a RETURN), for example: 1234, "some string", datetime(), {a: 1, b: "some string"}',
     ignoreFocusOut: true,
-    validateInput: (paramValue) => validateParamInput(paramValue, dbSchema),
+    validateInput: (paramValue) => {
+      return new Promise<string>((resolve) => {
+        clearTimeout(timeout);
+        timeout = setTimeout(() => {
+          const validationResult = validateParamInput(paramValue, dbSchema); //debouncedValidate.flush();
```
ncordon, line 81: Cleanup the `//debouncedValidate.flush();` please

packages/vscode-extension/src/commandHandlers/params.ts
```diff
@@ -44,6 +44,8 @@ export function validateParamInput(
   return undefined;
 }
 
+const DEBOUNCE_TIME = 200;
```
ncordon, line 47: There's no need for this constant, it's only used in that timeout. Also it's not a debounce time anymore

ncordon (approved):

PR #457, opened 2025-04-03
Fixes backticking of params

anderson4j:
Before we used the same backticking condition as the one for labels/reltypes/propertyKeys, which cant start with a number. Params can though, so we need a different condition.
Ex.
```
param "100param": RETURN $ ->
before: RETURN $`100param`
now: RETURN $100param
```


packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -41,20 +41,27 @@ const uniq = <T>(arr: T[]) => Array.from(new Set(arr));
...
-export function backtickIfNeeded(e: string): string | undefined {
-  if (e == null || e == '') {
-    return undefined;
-  } else if (/[^\p{L}\p{N}_]/u.test(e) || /[^\p{L}_]/u.test(e[0])) {
-    return `\`${e}\``;
-  } else {
-    return undefined;
-  }
-}
+type BacktickVariant = 'label' | 'propertyKey' | 'relType' | 'dbName' | 'param';
```
ncordon, line 44: I don't see the need for having these many types. The only one that is different is `param`, so having `'param' | 'nonParam'` would be more concise

ncordon (commented):
This is missing a changeset

packages/language-support/src/autocompletion/completionCoreCompletions.ts
anderson4j: dbName also has a special rule (periods are allowed without backticks if it's not the first or last symbol), and I figured for readability it's nice to see what type of elements we use which regex for

I did notice however that the old method was not checking for a period in the end of the name, so will adjust that

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -41,20 +41,29 @@ const uniq = <T>(arr: T[]) => Array.from(new Set(arr));
...
+      variant === 'propertyKey' ||
+      variant === 'relType') &&
+    (/[^\p{L}\p{N}_]/u.test(e) || /[^\p{L}_]/u.test(e[0]))
+  ) {
+    return `\`${e}\``;
+  } else if (
+    variant === 'dbName' &&
+    (/[^\p{L}\p{N}_.]/u.test(e) ||
+      /[^\p{L}_]/u.test(e[0]) ||
+      /[^\p{L}\p{N}_]/u.test(e[e.length - 1]))
```
ncordon, line 63: The `_` is not valid in a database name.

The nicer Typescript way to get the last element is `e.at(-1)`

ncordon: If you are going to change this you should include tests for some db names:

```
some.database
database-
database.
100database
```

anderson4j: yeah I noticed the db-name thing while testing the last letter. Since this means that dbs cant have a `_` in the string, e will never contain `_`, so this check doesnt affect valid dbnames. And since it still works for both direct names and aliases (and I think dbName is a nice shared variant name) I figure we can avoid an extra case by keeping it as is

anderson4j: e2e was failing so I think maybe some small detail was broken at the same time, but ill adjust to the nicer "-1"-way of indexing when i get to the bottom of the e2e test

anderson4j: Didnt see your second comment on my last message(s), but yeah we dont have any tests for certain odd but possible endings like `-` and `.`, so I will add these

ncordon (approved):

PR #460, opened 2025-04-07
Upgrades formatter from V2 -> V3

tomasnyberg:
## Description
This PR implements a baseline for the V3 iteration of the formatter which, in short, makes our formatter work similarly to Prettier (the Javascript formatter).

### Sample formatting output, before and after.
```
// Before
MERGE (a:Author {name: 'J.K. Rowling'})
  ON CREATE SET
    a.birthYear = 1965, a.nationality = 'British',
    a.booksWritten = 7, a.netWorth = 1000000000,
    a.genre = 'Fantasy'
MERGE (b:Book {title: 'Harry Potter and the Sorcerers Stone'})
  ON CREATE SET
    b.publishedYear = 1997,  b.sales = 120000000,
    b.rating = 4.8,  b.genre = 'Fantasy'
MERGE (a)-[:WROTE]->(b)
RETURN a, b

// After
MERGE (a:Author {name: 'J.K. Rowling'})
  ON CREATE SET
    a.birthYear = 1965,
    a.nationality = 'British',
    a.booksWritten = 7,
    a.netWorth = 1000000000,
    a.genre = 'Fantasy'
MERGE (b:Book {title: 'Harry Potter and the Sorcerers Stone'})
  ON CREATE SET
    b.publishedYear = 1997,
    b.sales = 120000000,
    b.rating = 4.8,
    b.genre = 'Fantasy'
MERGE (a)-[:WROTE]->(b)
RETURN a, b
```

```
// Before
CREATE (a:Location {name: "DXe5KhL3"}), (b:Location {name: "v2BpdkOj"}),
       (c:Location {name: "Fi5CMJ9Y"}), (d:Location {name: "S31K3X1o"}),
       (a)-[:ROUTE_TO {distance: "zjisNPKv", duration: "ivAC2TGF"}]->(b),
       (b)-[:ROUTE_TO {distance: "Irogkqf1", duration: "QsCt67v1"}]->(c),
       (c)-[:ROUTE_TO {distance: "Y53yoQwn", duration: "X41tnMDd"}]->(d);

// After
CREATE
  (a:Location {name: "DXe5KhL3"}),
  (b:Location {name: "v2BpdkOj"}),
  (c:Location {name: "Fi5CMJ9Y"}),
  (d:Location {name: "S31K3X1o"}),
  (a)-[:ROUTE_TO {distance: "zjisNPKv", duration: "ivAC2TGF"}]->(b),
  (b)-[:ROUTE_TO {distance: "Irogkqf1", duration: "QsCt67v1"}]->(c),
  (c)-[:ROUTE_TO {distance: "Y53yoQwn", duration: "X41tnMDd"}]->(d);
```

```
// Before
CALL apoc.periodic.iterate("eZQB0P0q", "1p7EFkyE",
                           {batchSize: "v0Ap5F8F", parallel: "UUc75lVg"})
YIELD batches, total, timeTaken, committedOperations, failedOperations

// After
CALL
  apoc.periodic.iterate(
    "eZQB0P0q",
    "1p7EFkyE",
    {batchSize: "v0Ap5F8F", parallel: "UUc75lVg"}
  )
  YIELD batches, total, timeTaken, committedOperations, failedOperations
```
For a more comprehensive overview, see the diffs in the test files.

### TL;DR: The Prettier approach
The "Prettier" approach to formatting is very simple; upon encountering a group, we check whether or not it would fit on the rest of the line. If it does, we put it there as is, putting `' '` or `''` between all of its children. If it does not, however, we insert a newline in every place where we would otherwise insert the spacing strings. Note that this is essentially the approach described in the ["A Prettier printer"](https://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf) paper, though we do not use the same operators / terminology used in that paper.

This comes with some _drastic_ changes to the whole formatting approach, though not all changes will be included in this PR. We now no longer even need the whole solution search logic, since the formatter will ever only have one choice between splits to put between two chunks. **This makes the formatter run in linear time** (!!!), and also greatly simplifies the formatting logic.

### Why is the logic simpler now?
The Prettier approach does not allow for much nuance in terms of what formatting output is better. It doesn't optimize for something like amount of lines used, nor anything else really beyond trying to keep the line within the maximum column width (which it does not do a perfect job of either). But, if you are okay with the output that it is limited to, that is okay! In fact, as became evident in our survey on preferred cypher styling, most people even prefer this simpler output. To illustrate this, consider the following formatting outputs:

```
// Formatting output A
MATCH (p:Product)
RETURN p.name AS productName, p.price AS basePrices, p.price AS priceWithTasx,
       p.price AS basePrice, p.price AS priceWithTax
LIMIT 25;

// Formatting output B
MATCH (p:Product)
RETURN
  p.name AS productName,
  p.price AS basePrices,
  p.price AS priceWithTasx,
  p.price AS basePrice,
  p.price AS priceWithTax
LIMIT 25;
```
Output A is significantly harder to achieve; making perfect use of the available space when splitting between any two elements is allowed effectively requires some cost-based solution search like what we have right now. Output B, on the other hand, is comparatively easy to achieve if you have already grouped elements together; the return list simply needs to split after each comma. Since output B is the one that people overwhelmingly preferred, using this simpler approach is perfectly fine in our case.

The solution search is thus only needed if you want to achieve tricky formatting outputs where splits occur in the middle of lines and with fancy alignment logic. If you are okay with outputs that do not require that, as I argue we are in the case of formatting Cypher, we can get away with this far simpler approach.

### Why did we not just do this in the first place, rather than the solution-search logic?
We arguably should have. The reason we didn't was that a lot of industry-standard formatters, and in particular the ones we took inspiration from, take the optimization-based approach we used for V2. We figured the Prettier approach would be (1) harder to implement and (2) not be flexible enough to allow for the formatting outputs we actually wanted. In hindsight, implementing this approach from the start might have been easier.

It is of course also very easy to make claims after-the-fact. We experimented with this approach earlier on, but were not able to make it work as easily as we could with the V2 approach. Hence why we launched the full V2 iteration using that logic. The Prettier approach is conceptually simple, but also *very* tricky to implement correctly, especially the indentation parts. Leveraging all the (working!) code we had from V2 also made it _much_ easier to implement this than when we were doing it from scratch. In particular, having perfectly grouped chunks for almost all language constructs helped a ton.

## Technical changes
- Groups now have sizes, that are precomputed in the visitor once it has visited all nodes
  - They also include the text that the group is made up of for debugging purposes (very nice to have)
- Group starts are now attached to the chunk prior, rather than on the chunk where it starts
  - This is  primarily to enable the "split before the group" behavior
- Indentation in the formatting state is now just one number (In V2 it consists of three different fields)
  - Groups no longer have anything to do with indentation.
  - Indentation can be added and removed only through `addIndentation` / `removeIndentation` in the visitor.
  - Similarly to groups, chunks now keep track of where indentation starts and ends, through IndentationModifier objects
- The solution search now only ever has one possible split choice between two chunks
  - What choice it has is decided by the Prettier logic described above
    - Specifically, if a group `breaksAll`, all split choices will be `\n`, and spaces otherwise. An exception is made if `avoidBreakBetween()` or `mustBreakBetween()` is applied.
  - We actually don't even need a solution search anymore, see the next section
- Inline comments are now attached to the chunk they are closest to, and are inserted later.
  - Hard-break comments, or ones with a newline before them, stay in the chunkList like before however.
    - They also cause all groups they are part of to break

### Technical debt, or, things that are missing in this PR but will be fixed in upcoming ones.
This PR is large, but omits a *lot* of things that should also be introduced as part of these changes. Because working on local branches is very inconvenient and slows down development, and because we think it's easier to review this way, we have decided to open this PR now, and then make follow up PRs to it fixing these issues. 

(will check these off once they have been implemented in future PRs).
- [ ] We do not actually need the solution search anymore, since we now only ever have one choice between chunks. 
  - [ ] We can probably remove a lot of the types used right now, such as Decision 
- [ ] The "split before" logic that allows for splits before the return item list as in output B above is way too convoluted.
  - We probably want to rethink where groups are placed to simplify this
- [ ] We no longer need multiple chunkLists; everything should just be one long list of chunks
  - The algorithm is linear, so splitting between e.g. clauses doesn't actually bring any benefit anymore. It actually just makes things more complicated.
- [ ] CASE, EXISTS, COUNT (maybe some others?) that use breakLine() and end groups between chunk lists do not work fully yet (though close to it).
  - They work reasonably well but not perfectly due to having separate chunk lists. See the commented out tests in this PR.
- [ ] Lists and maps that are the only thing in e.g. a return clause do not split exactly like we want them to yet. Ideally the opening bracket / brace should stay on the original line, like this:

```
RETURN [
  "M6XNVJsO",
  "NcdW0tuB",
  "Pf6RIuP4",
  "6tKStKwl",
  "HfvahDu5",
  "gJoq3HnU",
  "g7LjxbGD"
]
```

- [ ] The way parenthesized expressions apply indentation is not great (they need to add indentation even though they are not allowed to split after `(`.
- [ ] Inline comments allow for things to move before them nicely, but hard-break comments stay exactly where they are without compromises. In some cases where hard-break comments break the natural flow of for instance `OR` expressions, it would be nice to have that there as well.
- [ ] There may be cases with inline comments (these ones: `/* */`) that are not handled perfectly.


simonthuresson:
Nice description

packages/language-support/src/formatting/formatting.ts
```diff
@@ -174,25 +196,107 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     return resultString;
   };
 
+  // These three are helpers for the fillInGroupSizes method to make it more manageable
```
simonthuresson, line 199: Move these helpers to formattingHelpers file?

packages/language-support/src/formatting/formatting.ts
```diff
@@ -232,13 +336,10 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
-      groupsEnding: prefix.groupsEnding + suffix.groupsEnding,
-      indentation: {
-        base: prefix.indentation.base + suffix.indentation.base,
-        special: prefix.indentation.special + suffix.indentation.special,
-        align: prefix.indentation.align,
-      },
+      groupsStarting: prefix.groupsStarting.concat(suffix.groupsStarting),
+      groupsEnding: prefix.groupsEnding.concat(suffix.groupsEnding),
+      indentation: prefix.indentation.concat(suffix.indentation),
+      comment: (prefix.comment || '') + (suffix.comment || '') || undefined,
```
simonthuresson, line 342: Consider using empty string instead

packages/language-support/src/formatting/formatting.ts
```diff
@@ -311,80 +408,60 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
 
   startGroup = (): number => {
-    this.startGroupCounter += 1;
-    this.groupStack.push(this.groupID);
+    const lastChunk = this.lastInCurrentBuffer();
+    const newGroup: Group = {
+      id: this.groupID,
+      dbgText: '',
+      size: 0,
+      align: 0, // Irrelevant here
```
simonthuresson, line 426: Align can definitely be removed right now

packages/language-support/src/formatting/formatting.ts
```diff
@@ -774,48 +863,46 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
-    this.endGroup(withClauseGrp);
     this._visit(ctx.whereClause());
   };
 
   visitMatchClause = (ctx: MatchClauseContext) => {
     this._visit(ctx.OPTIONAL());
     this._visit(ctx.MATCH());
-    this.avoidBreakBetween();
-    const matchClauseGrp = this.startGroupAlsoOnComment();
+    const indentId = this.addIndentation();
```
simonthuresson, line 881: Nit: We could improve the variable name on these. Similar to group names

packages/language-support/src/formatting/formatting.ts
```diff
@@ -861,12 +949,16 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    const indentId1 = this.addIndentation();
     this._visit(ctx.expression());
     const asGrp = this.startGroup();
+    this.avoidBreakBetween();
     this._visit(ctx.AS());
+    const indentId2 = this.addIndentation();
     this._visit(ctx.variable());
+    this.removeIndentation(indentId2);
+    this.removeIndentation(indentId1);
     this.endGroup(asGrp);
```
simonthuresson, line 962: It would be nice if removeIndentation and endGroup could take an arbitrary amount of id:s and end them all. It would remove one line here atleast

packages/language-support/src/formatting/formatting.ts
tomasnyberg: We still use it to calculate the state key so I figured we would lave it until then? Otherwise this PR also has to change the state key. And if it does that it should arguably remove it entirely.

packages/language-support/src/formatting/formatting.ts
tomasnyberg: Hmm not sure if I agree, I really like the explicitness of ending them like this. In my mind the indentations simply happen to end at the same place. As such, we do not logically want to end multiple indentations, two of them just happen to end right after one another. If it was the same method call **it would imply that the indentations are related to each other**, which they are not.

It also only happens in < 5 places, and so wouldn't really reduce LOC

packages/language-support/src/formatting/formatting.ts
tomasnyberg: Yeah these should all have names similar to the group ones, just haven't been bothered to change them just yet ^^. Trying to find a good way to let GPT do this for me...

packages/language-support/src/formatting/formatting.ts
```diff
@@ -1244,8 +1377,12 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     }
     const anonymousPatternGrp = this.startGroup();
     this._visit(ctx.anonymousPattern());
+    if (ctx.variable()) {
+      this.removeIndentation(indentId);
+    }
     this.endGroup(anonymousPatternGrp);
     this.endGroup(selectorAnonymousPatternGrp);
+    this.endGroup(wholePatternGrp);
```
simonthuresson, line 1396: It is a little confusing why this is not inside the if statement, even though it is technically not needed. Maybe move inside if statement. Or removeIndentation could have the same check as endGroup

simonthuresson: Here it would be really awesome as well if endGroup could take an arbitrary amount of arguments.

packages/language-support/src/formatting/formatting.ts
tomasnyberg: I think using ? makes the logic more clear, since it more explicitly shows that there is no comment attached to some chunks. An empty string feels less correct in that regard.

Of course, that makes this line a bit messy, but I think it's a price worth paying in this case.

Curious what @OskarDamkjaer and @ncordon think here.

packages/language-support/src/formatting/formatting.ts
tomasnyberg: Yeah good idea. Will do.

packages/language-support/src/formatting/formatting.ts
tomasnyberg: I'd make the same argument [as above](https://github.com/neo4j/cypher-language-support/pull/460#discussion_r2031048001) here. 

What do you guys think? @ncordon @OskarDamkjaer

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -61,18 +53,17 @@ export interface CommentChunk extends BaseChunk {
...
 export type Chunk = RegularChunk | CommentChunk | SyntaxErrorChunk;
 
-export const initialIndentation: ChunkIndentation = {
-  base: 0,
-  special: 0,
-  align: AlignIndentationOptions.Maintain,
-};
+export interface IndentationModifier {
+  id: number;
+  change: number;
```
simonthuresson, line 58: Should this perhaps be an Enum?

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -94,7 +101,7 @@ const noSpaceNoBreakSplit: Split[] = [{ splitType: '', cost: 0 }];
 const onlyBreakSplit: Split[] = [{ splitType: '\n', cost: 0 }];
 const onlyDoubleBreakSplit: Split[] = [{ splitType: '\n\n', cost: 0 }];
 
-function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
+export function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
```
simonthuresson, line 104: Move to helper?

packages/language-support/src/formatting/formattingHelpers.ts
tomasnyberg: Great idea, the values should definitely be restricted on a type level. I think an enum is a bit overkill though, but I've changed it to this:

```
export interface IndentationModifier {
  id: number;
  change: 1 | -1;
}
```

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -25,7 +29,14 @@ interface Choice {
   possibleSplitChoices: Split[];
 }
 
-interface Group {
+export interface Group {
+  id: number;
+  // Should this group break in the "Prettier" fashion, breaking between
+  // all of its children?
+  breaksAll?: boolean;
```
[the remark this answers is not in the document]
simonthuresson: Those were suggestions by Chat

packages/language-support/src/formatting/formattingSolutionSearch.ts
tomasnyberg: yes. moved it.

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -103,89 +110,32 @@ function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
-  //   - the base indentation.
-  if (state.indentationState.align.length > 0) {
-    return (
-      nextIndentationState.align.at(-1) +
-      INDENTATION_SPACES +
-      state.indentationState.base
-    );
+  isBreak: boolean,
+): [number, IndentationModifier[]] {
+  let newActive = [...state.activeIndents];
```
simonthuresson, line 118: newActive does not really describe what it is. Why are they "new"? `activeIndentations` perhaps?

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -103,89 +110,32 @@ function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
+  // The "indentation" number is the actual amount of indentation at the current state,
+  // but it should not be confused with the entire indentation state. The entire indentation state
+  // also includes the indentation modifiers, that keep track of when this number should change
+  let nextIndentation: number = state.indentation;
```
simonthuresson, line 122: Nit: `: number` is unecessary

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -103,89 +110,32 @@ function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
+  for (const indent of chunk.indentation) {
+    if (indent.change === 1 && isBreak) {
```
simonthuresson, line 124: Here the Enum as talked about earlier would make it easier to read

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -103,89 +110,32 @@ function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
+      newActive.push(indent);
+      nextIndentation += INDENTATION_SPACES;
+    }
+    if (indent.change === -1) {
+      // PERF: This is O(n) twice and doesn't neccesarily have to be. Might be worth optimizing if profiling
+      // shows that it is actually a perf sink.
+      const index = newActive.findIndex((i) => i.id === indent.id);
```
simonthuresson, line 131: Not obvoius what `i` is here

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -103,89 +110,32 @@ function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
+      if (index !== -1) {
+        newActive = newActive.filter((i) => i.id !== indent.id);
```
simonthuresson, line 133: Same as above here, what is `i`

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -103,89 +110,32 @@ function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
...
+        nextIndentation -= INDENTATION_SPACES;
+      }
+    }
   }
-
-  // Default case
-  return state.indentationState.base;
+  return [nextIndentation, newActive];
```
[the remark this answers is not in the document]
tomasnyberg: yes. I've changed the state to have an indentationState type as wel.

packages/language-support/src/formatting/formattingSolutionSearch.ts
tomasnyberg: updated now

packages/language-support/src/formatting/formattingSolutionSearch.ts
tomasnyberg: removed it 👍

packages/language-support/src/formatting/formattingSolutionSearch.ts
tomasnyberg: I think an enum is a bit too much but I've restricted it to have to be either 1 or -1

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -223,17 +166,35 @@ function getNeighbourState(curr: State, choice: Choice, split: Split): State {
       : thisWordEnd - splitLength;
   const overflowingCount = Math.max(0, endWithoutCommentAndSplit - MAX_COL);
 
-  for (let i = 0; i < choice.left.groupsStarting; i++) {
-    nextGroups.push({
+  for (let i = 0; i < choice.left.groupsStarting.length; i++) {
```
simonthuresson, line 169: Change to for (const group of choice.left.groupStarting) ... ?

Same for groups ending

tomasnyberg: yeah nice catch! It's an indexed loop right now because it used to be a number, but that obv doesn't make sense anymore since it's now a list of objects.

packages/language-support/src/formatting/formatting.ts
OskarDamkjaer: With an empty string it's hard to tell between nothing and an empty comment no?

packages/language-support/src/formatting/formattingSolutionSearch.ts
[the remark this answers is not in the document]
OskarDamkjaer: @tomasnyberg at the same time the comment describing the variable I think is clear, and it uses "should"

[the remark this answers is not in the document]
OskarDamkjaer: `breakGroup` or `shouldBreakGroup` would also work imo

OskarDamkjaer (approved):

packages/language-support/src/tests/formatting/linebreaks.test.ts
```diff
@@ -159,44 +134,52 @@ RETURN path`;
...
   });
 
   test('aligns large maps one further than the opening brace', () => {
     const query = `RETURN {looooooooooooooooooooooongkey:value, loooooooooooooooooooongkeeeyyyyyyyy:value2, looooooooooooooongkeeey:value3}`;
     const expected = `
-RETURN {looooooooooooooooooooooongkey: value,
-        loooooooooooooooooooongkeeeyyyyyyyy: value2,
-        looooooooooooooongkeeey: value3}`.trimStart();
+RETURN
+  {
```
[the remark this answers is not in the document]
tomasnyberg: yes!

packages/language-support/src/formatting/formattingSolutionSearch.ts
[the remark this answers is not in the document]
tomasnyberg: > @tomasnyberg at the same time the comment describing the variable I think is clear, and it uses "should"

Contradicted by my own comment it seems 😂 . I'll go with `shouldBreak`.

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -160,3 +151,74 @@ export function isCommentBreak(chunk: Chunk, nextChunk: Chunk): boolean {
...
+  groupsEnding: Set<number>,
+) {
+  for (const group of activeGroups) {
+    if (!chunk.text) {
+      throw new Error(INTERNAL_FORMAT_ERROR_MESSAGE);
+    }
+    group.size += chunk.text.length;
+    // PERF: Right now we include dbgText always, even though it's only used for debugging.
+    // It does not seem to have any significant performance downsides, but only doing so
+    // when e.g. a flag is set might be a more prudent choice.
```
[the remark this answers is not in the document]
tomasnyberg: Yeah if it's a small difference I think always including it is better to avoid the complexity and risk for error. But on a similar note, isn't adding strings like we're doing here slow, since it creates immutable objects of each string as we are adding things? Perhaps it would be better to have an array of strings which is then joined together. 

Regardless, profiling should be done before we make any decisions

packages/language-support/src/tests/formatting/comments.test.ts
```diff
@@ -492,29 +535,22 @@ LIMIT "6pkMe6Kx"`;
...
-     p, q
-UNWIND // Expands lists into multiple rows
-       [1, 2, 3] AS num;`;
+USE graph // Specifies the graph or database to use
+MATCH (m)-[:RELATION]->(n) // Matches patterns in the graph
+MERGE (p:Person {name: "Alice"}) // Ensures a pattern exists in the graph
+CREATE (q:Person {name: "Bob"})-[:KNOWS]->(p) // Creates new nodes or relationships
+DELETE r // Deletes nodes or relationships
+WITH p, q // Passes results to the next clause
+UNWIND [1, 2, 3] AS num; // Expands lists into multiple rows`.trimStart();
```
ncordon, line 544: Is this changing where the comment is with respect to the original query?

packages/language-support/src/tests/formatting/linebreaks.test.ts
```diff
@@ -602,12 +739,500 @@ MATCH (Alice123:Person)-[FRND_REL:friendship]->
...
+                )
+return n`;
+    const expected = `MATCH (n)
+WHERE
+  single(
+    this12 IN
+    [
+      (this11)<-[:REPRESENTS]-(this12:Identity)
+      WHERE ($jwt.email IS NOT NULL AND this12.identity = $jwt.email)
+      | "p4cI8VGS"
```
ncordon, line 1002: Why is this line not indented?

packages/language-support/src/tests/formatting/comments.test.ts
tomasnyberg: Yes, inline comments get attached to the the chunk to the left of them (except for when there are comments at the start of the query), and then they are inserted after a newline is chosen in the layout. So e.g. the first one is attached to e.g. the `USE` chunk, and is inserted after `graph` because a newline is coming after it.

This is similar to what Prettier does for some comments, for instance this case 

```javascript
// before formatting
function abc // comment
(
a
) {
 return a;
}
// after
function abc(a) { // comment
  return a;
}
```

tomasnyberg: ACtually nevermind this is just a query=expected case, in fact it seems like we've made the test easier (?) not sure why. Let me look into this.

tomasnyberg: We changed it because we thought it was correct now but actually it should look like this 

```
USE // Specifies the graph or database to use
 graph
MATCH // Matches patterns in the graph
 (m)-[:RELATION]->(n)
MERGE // Ensures a pattern exists in the graph
 (p:Person {name: "Alice"})
```
Which is what it looks like in the last version, but not in this one. I didn't realize the way this is handled changed between the PRs. I've added a comment to the last PR (which will be the only one after this one) to make sure we keep this change.

ncordon (approved):
lgtm

packages/language-support/src/tests/formatting/linebreaks.test.ts
```diff
@@ -602,12 +739,500 @@ MATCH (Alice123:Person)-[FRND_REL:friendship]->
...
+    b.genre = 'Fantasy'
+MERGE (a)-[:WROTE]->(b)
+RETURN a, b`;
+    verifyFormatting(query, expected);
+  });
+
+  /* TODO: this does not fully work yet since the parenthesized expr
+ * adds too much indentation. Really, it should be the CASE taking 
+ * responsibility for the indentation here. It's almost correct though.
+  test('avoidbreak and mustbreak should cooperate', () => {
```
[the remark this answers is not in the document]
tomasnyberg: Yeah that would have been better. It's included in the final version though so I'll leave it as is in this case

PR #463, opened 2025-04-08
Removes solution search from Formatter (wrong target branch)

simonthuresson:
# Description
This PR removes the solution search from the formatter. This is a tech debt from #460 which is getting fixed. 

### TODO
 [] Change target branch

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -364,7 +364,6 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       dbgText: '',
       size: 0,
       align: 0, // Irrelevant here
-      breakCost: 0,
     };
     lastChunk.groupsStarting.push(newGroup);
     this.groupStack.push(newGroup);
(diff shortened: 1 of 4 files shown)


packages/language-support/src/formatting/formatting.ts
```diff
@@ -364,7 +364,6 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
       dbgText: '',
       size: 0,
       align: 0, // Irrelevant here
```
tomasnyberg, line 366: remove align as well?

packages/language-support/src/formatting/formattingSolutionSearch.ts
```diff
@@ -134,15 +115,11 @@ function getNextIndentationState(
...
-// Very useful for debugging but not actually used in the code
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-function stateToString(state: State) {
-  const result = reconstructBestPath(state);
-  const resultString = decisionsToFormatted(result.decisions);
-  return resultString;
-}
-
-function getNeighbourState(state: State, choice: Choice, split: Split): State {
+function createStateTransition(
```
tomasnyberg, line 118: remove align and breakCost in the "newGroup" obejct that gets created as well

PR #465, opened 2025-04-09
Adds mock tests for the parameters pane

ncordon:
Adds more unit tests for the parameters pane.

Up until now we had a few e2e tests, but I actually discovered a couple of bugs making this PR.

This PR also does some cleanup:

* Moves many error messages to the `CONSTANTS` file.
* Adds a `DEBUG_VSCODE_TESTS` variable that we only set to true in the debugger, so that we don't timeout when stopping on variables.


packages/vscode-extension/src/commandHandlers/params.ts
```diff
@@ -94,25 +95,25 @@ export async function addParameter(): Promise<void> {
...
     return;
   }
-  const existingParam = getParameter(paramItem.id);
+  const existingParam = getParameterByKey(paramItem.id);
   if (!existingParam) {
     return;
   }
   const schemaPoller = getSchemaPoller();
-  const dbSchema = schemaPoller.metadata.dbSchema;
+  const dbSchema = schemaPoller?.metadata?.dbSchema ?? {};
```
ncordon, line 108: This should not happen in theory, but just to be sure

packages/vscode-extension/tests/runWebviewTests.ts
```diff
@@ -150,7 +150,10 @@ export const config: Options.Testrunner = {
   // See the full list at http://mochajs.org/
   mochaOpts: {
     ui: 'tdd',
-    timeout: 120000,
+    timeout:
+      process.env.DEBUG_VSCODE_TESTS === 'true'
+        ? Number.POSITIVE_INFINITY
+        : 120000,
```
ncordon, line 156: I've added a `DEBUG_VSCODE_TESTS` variable that prevents the tests from timing out (this is a problem when you are running several tests and you are inspecting things in the debugger)

packages/vscode-extension/src/parameterService.ts
```diff
@@ -36,6 +37,7 @@ export function getParameters(): Parameters {
 export async function clearParameters(): Promise<void> {
   const context = getExtensionContext();
   await context.globalState.update(PARAMETERS_KEY, {});
+  await sendParametersToLanguageServer();
```
ncordon, line 40: This was a bug, when clearing the parameters those should go out of scope for the language server.

anderson4j (commented):
Nice! Had some small notes

anderson4j (changes requested):
You missed one of the comments from last review (one test is idententical to the one above it). Other than that LGTM

packages/vscode-extension/tests/testRunnerDebug.ts
```diff
@@ -5,5 +5,6 @@ export async function run(): Promise<void> {
   const container = await createAndStartTestContainer();
   process.env.NEO4J_PORT = container.getMappedPort(7687).toString();
   process.env.CYPHER_25 = 'true';
+  process.env.DEBUG_VSCODE_TESTS = 'true';
```
anderson4j, line 8: Actually one more thing: This should be false when pushing any PR to main right? Should we maybe have a test that checks this to be false, and just accept that the test fails while debugging the rest with this flag?

This way we could never accidentaly push with the flag as true but all tests looking green

packages/vscode-extension/tests/specs/unit/paramCommandHandlers.spec.ts
```diff
@@ -0,0 +1,327 @@
...
+      .getChildren()
+      .find((param) => param.id === 'a');
+    await editParameter(param);
+    sandbox.assert.calledOnceWithExactly(
+      showErrorMessageStub,
+      CONSTANTS.MESSAGES.ERROR_EMPTY_PARAM_VALUE,
+    );
+  });
+
+  test('Editing parameters should fail if the new value is empty', async () => {
```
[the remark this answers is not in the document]
ncordon: Cool, I missed this

packages/vscode-extension/tests/testRunnerDebug.ts
ncordon: This file is only used when debugging the tests through the VSCode run and debug menu, so it needs to stay true here.

For tests run from the CI we set the variables to false in the package.json:

```
    "test:e2e": "DEBUG_VSCODE_TESTS=false && npm run build:dev && npm run test:apiAndUnit && npm run test:webviews",
    "test:apiAndUnit": "DEBUG_VSCODE_TESTS=false && npm run build:dev && rm -rf .vscode-test/user-data && node ./dist/tests/runApiAndUnitTests.js",
    "test:webviews": "DEBUG_VSCODE_TESTS=false && npm run build:dev && wdio run ./dist/tests/runWebviewTests.js"
 ```

anderson4j: Yeah, by default value I meant the one set where env is defined, in launch.json - it's set to true there too, so I suppose manual runs of the test files will get infinite timers (not that I've ever used anything but premade commands for e2e tests yet, but who knows what the future holds)

anderson4j (approved):

PR #467, opened 2025-04-09
Solves v3 tech debt: simplify 'breakBefore' logic  (STACKED PR; WRONG TARGET BRANCH)

tomasnyberg:
# Note: this PR currently targets another branch that is a PR stacked on the original V3 PR. This is to make sure the diff looks correct. Make sure to move the target before merging

## Description
In the original implementation of V3 as described in #460, we have an idea of `breakBefore`, which is essentially just to enable this type of pattern:

```
RETURN 
  a,
  b,
  c
```
The first group starts at `a`, and so in order to achieve this, we came up with some convoluted logic where a group could start breaking before it actually starts. Because of this, we also moved group starts to the chunk prior to where they actually start, in order to enable this pattern. This was noted as a bit messy in a few TODO comments, as well as [in the original PR](https://github.com/neo4j/cypher-language-support/pull/460#:~:text=The%20%22split%20before%22%20logic,placed%20to%20simplify%20this)

This PR instead uses the much more sound approach of
1. Moving group starts back to where they actually start rather than the group before
2. Wrapping every clause in a group, meaning that in the example above, the group between `RETURN` and `a` naturally happens because the whole `RETURN` group doesn't fit
3.  Removing anything related to `breakBefore`, since none of it is actually needed anymore.
4. Moving all the group start / ending logic out of the `createStateTransition` function; it now only contains logic for determining the indentation state and where the current chunk ends.

Overall, this should *greatly* simplify both the conceptual logic as well as the code.

### Testing
All old tests pass except for 2; one of them is technically more correct now but should be fixed by moving things around hard-break comments like we do for inline ones, and the other has to do with us having multiple chunkLists (a separate issue). I've changed the tests to pass and included TODO Comments that describe this.

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -108,7 +108,6 @@ import {
   CommentChunk,
   fillInRegularChunkGroupSizes,
   findTargetToken,
-  getActiveGroups,
   getParseTreeAndTokens,
   IndentationModifier,
   INTERNAL_FORMAT_ERROR_MESSAGE,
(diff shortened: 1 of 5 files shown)


tomasnyberg:
Closing this in favor of https://github.com/neo4j/cypher-language-support/pull/470

PR #469, opened 2025-04-09
Solves v3 tech debt: only one chunk list (STACKED PR; WRONG TARGET BRANCH)

tomasnyberg:
## Description
As noted in the original V3 pr (#460), we no longer need multiple chunk lists. Previously, we separated the problem of formatting between each clause, such that we formatted a list of chunks one clause at a time, and then put the result together for the full formatted output. This was necessary because otherwise the search space for the solution search would be way too large. Since we now no longer do a solution search, this can be dropped, which is what this PR does.

This also enables us to format things that contain nested queries, such as EXISTS expressions. Some of those were left commented out in the original PR, but they are now reenabled. An example would be this one:

```
MATCH (person:Person)
RETURN
  person.name AS name,
  COLLECT {
    MATCH (person)-[r:HAS_DOG]->(dog:Dog)
    WHERE r.since > 2017
    RETURN dog.name
  } AS youngDogs
```
This is the correct output according to the Prettier style; we split between every item in the return list (the COLLECT expression counts as one item). It was not possible to do this previously, as we had to end all groups at the start of the subquery in COLLECT. Now, all clauses are formatted like any other language construct.
 

## Testing
Uncommented a few of the older complicated EXISTS / CASE tests that now look good as a result of this

### DIFF
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -2172,7 +2172,10 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     }
     this.removeIndentation(clauseIndent);
     this.breakLine();
-    this._visit(ctx.RPAREN());
+    this._visitTerminalRaw(ctx.RPAREN(), {
+      dontConcatenate: true,
+      spacingChoice: 'SPACE_AFTER',
+    });
   };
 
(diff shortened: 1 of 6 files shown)


tomasnyberg:
Closing this in favor of https://github.com/neo4j/cypher-language-support/pull/470

PR #470, opened 2025-04-09
Fixes tech debt for formatter

simonthuresson:
### Description
This PR builds on top of #468. 

This PR accomplishes the final state of the formatter. (as we discussed) This means it includes a complete redesigning of how the formatter goes from a chunk list to formatted string. Types and concepts such as Split, Choice and Decision. This is replaced by a much more straightforward for loop which processes the chunk list linear.

### DIFF
--- a/packages/language-support/package.json
+++ b/packages/language-support/package.json
@@ -45,7 +45,6 @@
     "antlr4": "4.13.2",
     "antlr4-c3": "*",
     "fastest-levenshtein": "^1.0.16",
-    "heap-js": "^2.6.0",
     "vscode-languageserver-types": "^3.17.3"
   },
   "scripts": {
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -108,7 +108,6 @@ import {
   CommentChunk,
   fillInRegularChunkGroupSizes,
   findTargetToken,
-  getActiveGroups,
   getParseTreeAndTokens,
   IndentationModifier,
   INTERNAL_FORMAT_ERROR_MESSAGE,
@@ -206,7 +205,9 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
         const groupsEnding = new Set<number>(
           chunk.groupsEnding.map((g) => g.id),
         );
-        activeGroups = getActiveGroups(activeGroups, groupsEnding, chunk);
+        for (const group of chunk.groupsStarting) {
+          activeGroups.push(group);
+        }
         if (chunk.type === 'REGULAR') {
           fillInRegularChunkGroupSizes(chunk, activeGroups, groupsEnding);
         } else if (chunk.type === 'COMMENT') {
@@ -217,6 +218,13 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
           if (activeGroups.length > 0) {
             activeGroups[0].breaksAll = true;
           }
+        }
+        for (const group of chunk.groupsEnding) {
+          if (group.dbgText.at(-1) === ' ') {
+            group.size--;
+            group.dbgText = group.dbgText.slice(0, -1);
+          }
+          activeGroups = activeGroups.filter((g) => g.id !== group.id);
         }
       }
     }
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -134,7 +134,7 @@ type SpacingChoice = 'SPACE_AFTER' | 'EXTRA_SPACE';
 export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   root: StatementsOrCommandsContext;
   query: string;
-  buffers: Chunk[][] = [];
+  chunkList: Chunk[] = [];
   indentation = 0;
   indentationSpaces = 2;
   targetToken?: number;
@@ -162,7 +162,6 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     super();
     this.root = root;
     this.query = query;
-    this.buffers.push([]);
     if (unParseable) {
       this.unParseable = unParseable;
     }
@@ -176,7 +175,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     this._visit(this.root);
     this.removeAllPendingGroups();
     this.fillInGroupSizes();
-    const result = buffersToFormattedString(this.buffers);
+    const result = buffersToFormattedString(this.chunkList);
     this.cursorPos += result.cursorPos;
     const resultString = result.formattedString + this.unParseable;
     const originalNonWhitespaceCount = this.query.replace(/\s/g, '').length;
@@ -198,38 +197,34 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   };
 
   fillInGroupSizes = () => {
-    for (const chunkList of this.buffers) {
-      let activeGroups: Group[] = [];
-      for (let i = 0; i < chunkList.length; i++) {
-        const chunk = chunkList[i];
-        activeGroups = activeGroups.concat(chunk.groupsStarting);
+    let activeGroups: Group[] = [];
+    for (let i = 0; i < this.chunkList.length; i++) {
+      const chunk = this.chunkList[i];
+      activeGroups = activeGroups.concat(chunk.groupsStarting);
 
-        if (chunk.type === 'REGULAR') {
-          fillInRegularChunkGroupSizes(chunk, activeGroups);
-        } else if (chunk.type === 'COMMENT') {
-          // If we have a hard-break comment, i.e. one that is on its own line (and thus part of the chunkList)
-          // the groups it is a part of will always break.
-          activeGroups.forEach((group) => {
-            group.breaksAll = true;
-          });
-        }
-        chunk.groupsEnding.forEach((group) => {
-          if (group.dbgText.at(-1) === ' ') {
-            group.size--;
-            group.dbgText = group.dbgText.slice(0, -1);
-          }
-          // PERF: this is O(n) and doesn't really need to be.
-          activeGroups = activeGroups.filter((g) => g.id !== group.id);
+      if (chunk.type === 'REGULAR') {
+        fillInRegularChunkGroupSizes(chunk, activeGroups);
+      } else if (chunk.type === 'COMMENT') {
+        // If we have a hard-break comment, i.e. one that is on its own line (and thus part of the chunkList)
+        // the groups it is a part of will always break.
+        activeGroups.forEach((group) => {
+          group.breaksAll = true;
         });
       }
+      chunk.groupsEnding.forEach((group) => {
+        if (group.dbgText.at(-1) === ' ') {
+          group.size--;
+          group.dbgText = group.dbgText.slice(0, -1);
+        }
+        // PERF: this is O(n) and doesn't really need to be.
+        activeGroups = activeGroups.filter((g) => g.id !== group.id);
+      });
     }
     // If we made an error when calculating the group sizes, that counts as an internal bug.
-    verifyGroupSizes(this.buffers);
+    verifyGroupSizes(this.chunkList);
   };
 
-  currentBuffer = () => this.buffers.at(-1);
-
-  lastInCurrentBuffer = () => this.currentBuffer().at(-1);
+  lastInChunkList = () => this.chunkList.at(-1);
 
   removeAllPendingGroups = () => {
     while (this.groupStack.length > 0) {
@@ -241,11 +236,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     // groups might be active when breakLine() is called, but it does not make sense to keeep them active
     // so clear all groups when this happens.
     // TODO: v3 tech debt; we should not end groups between breakLine() calls.
-    this.removeAllPendingGroups();
-
-    if (this.currentBuffer().length > 0) {
-      this.buffers.push([]);
-    }
+    this.mustBreakBetween();
   };
 
   // If two tokens should never be split, concatenate them into one chunk
@@ -253,8 +244,8 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     // Loop since we might have multiple comments or special chunks anywhere, e.g. [b, C, C, a, C]
     // but we should still be able to concatenate a, b to ba
     const indices: number[] = [];
-    for (let i = this.currentBuffer().length - 1; i >= 0; i--) {
-      if (!(this.currentBuffer()[i].type === 'COMMENT')) {
+    for (let i = this.chunkList.length - 1; i >= 0; i--) {
+      if (!(this.chunkList[i].type === 'COMMENT')) {
         indices.push(i);
         if (indices.length === 2) {
           break;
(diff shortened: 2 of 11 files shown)


packages/language-support/src/formatting/formatting.ts
```diff
@@ -88,42 +104,49 @@ import {
...
 
 type SpacingChoice = 'SPACE_AFTER' | 'EXTRA_SPACE';
 
 export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   root: StatementsOrCommandsContext;
   query: string;
-  buffers: Chunk[][] = [];
+  chunkList: Chunk[] = [];
   indentation = 0;
   indentationSpaces = 2;
```
simonthuresson, line 139: Can be removed

packages/language-support/src/formatting/lineFormatter.ts
```diff
@@ -0,0 +1,192 @@
...
+): FinalResultWithPos {
+  const state = createInitialState();
+
+  for (let i = 0; i < chunkList.length; i++) {
+    const chunk = chunkList[i];
+    const nextChunk = chunkList[i + 1];
+
+    applyIndentationIfNeeded(state);
+    checkAndSetCursorPosition(state, chunk);
+    updateActiveGroups(state, chunk);
```
simonthuresson, line 97: Be consistent so that groups and indentation starts at the same place and with the same behavior?

tomasnyberg:
Fix this test 

https://github.com/neo4j/cypher-language-support/pull/460/files#r2039262762

We made it easier for no reason in the original V2 pr, but it works properly now, so it should be included.

packages/language-support/src/formatting/lineFormatter.ts
tomasnyberg, line 1: rename to `layout.ts`?

packages/language-support/package.json
```diff
@@ -45,7 +45,6 @@
     "antlr4": "4.13.2",
     "antlr4-c3": "*",
     "fastest-levenshtein": "^1.0.16",
-    "heap-js": "^2.6.0",
```
tomasnyberg, line 48: 🔥

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -29,25 +29,25 @@ export interface BaseChunk {
...
 }
 
 // Regular chunk specific properties
 export interface RegularChunk extends BaseChunk {
-  type: 'REGULAR';
-  node?: TerminalNode;
   noSpace?: boolean;
   noBreak?: boolean;
-  mustBreak?: boolean;
+  type: 'REGULAR';
```
tomasnyberg, line 39: unnecessary diff

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -29,25 +29,25 @@ export interface BaseChunk {
   // Comment that is attached to a chunk. Not to be confused with a comment
   // that is in the chunklist (one with a newline before it.)
   comment?: string;
+  mustBreak?: boolean;
```
tomasnyberg, line 32: this should be called `shouldBreak` as discussed in the original pr https://github.com/neo4j/cypher-language-support/pull/460

packages/language-support/src/formatting/formatting.ts
```diff
@@ -191,66 +190,57 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
-          // the groups it is a part of will always break.
-          // NOTE: This might not be entirely sound, but it seems to work flawlessly for comments
-          // so far.
-          if (activeGroups.length > 0) {
-            activeGroups[0].shouldBreak = true;
-          }
-        }
-        activeGroups = getActiveGroups(activeGroups, groupsEnding, chunk);
+    let activeGroups: Group[] = [];
+    for (let i = 0; i < this.chunkList.length; i++) {
```
simonthuresson, line 200: Cant this be a for (const group of groups)

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -177,6 +162,29 @@ export function fillInRegularChunkGroupSizes(
...
+      }
+    }
+  }
+}
+
+const openingCharacters = [CypherCmdLexer.LPAREN, CypherCmdLexer.LBRACKET];
+
+export function doesNotWantSpace(chunk: Chunk, nextChunk: Chunk): boolean {
+  return (
+    (nextChunk?.type !== 'COMMENT' &&
```
tomasnyberg, line 179: could this method be refactored a bit for readability=?

packages/language-support/src/formatting/lineFormatter.ts
```diff
@@ -0,0 +1,194 @@
...
+
+const INDENTATION_SPACES = 2;
+
+interface FinalResultWithPos {
+  formattedString: string;
+  cursorPos: number;
+}
+
+interface State {
+  formatted: string;
```
tomasnyberg, line 17: this should arguably be a string[] that is then joined together at the end. I think we tried this and it didn't have a huge performance diff though so we can leave it like this and then implement performance improvments after actually profiling

packages/language-support/src/formatting/lineFormatter.ts
```diff
@@ -0,0 +1,194 @@
...
+  doesNotWantSpace,
+  IndentationModifier,
+  INTERNAL_FORMAT_ERROR_MESSAGE,
+  MAX_COL,
+} from './formattingHelpers';
+
+const INDENTATION_SPACES = 2;
+
+interface FinalResultWithPos {
+  formattedString: string;
```
tomasnyberg, line 12: why is this called `formattedString` while the other one is called just `formatted`?

packages/language-support/src/formatting/formattingHelpers.ts
tomasnyberg: Similarly to the `shouldBreak helper`

packages/language-support/src/formatting/formatting.ts
```diff
@@ -802,26 +775,34 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    this._visitReturnBodyNotItems(ctx.returnBody());
     this._visit(ctx.whereClause());
   };
 
   visitMatchClause = (ctx: MatchClauseContext) => {
+    const matchClauseGrp = this.startGroup();
     this._visit(ctx.OPTIONAL());
+    if (ctx.OPTIONAL()) {
+      this.avoidBreakBetween();
+    }
```
simonthuresson, line 799: Use group with indentation around instead of avoid break? A more dynamic solution

packages/language-support/src/formatting/lineFormatter.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  return (
+    chunk.mustBreak ||
+    chunk.doubleBreak ||
+    state.activeGroups.at(-1)?.shouldBreak
+  );
+}
+
+function updateActiveGroups(state: State, chunk: Chunk): void {
+  for (const group of chunk.groupsStarting) {
```
tomasnyberg, line 51: nit: empty groups (ones starting and ending on the same chunk) should ideally be filtered out in the visitor. I think it's fine to leave as is right now but it in some cases makes debugging more tedious since it can be harder to see the right. groups.

packages/language-support/src/formatting/lineFormatter.ts
```diff
@@ -0,0 +1,194 @@
...
+      } else {
+        throw new Error(INTERNAL_FORMAT_ERROR_MESSAGE);
+      }
+    }
+  }
+}
+export function chunksToFormattedString(
+  chunkList: Chunk[],
+): FinalResultWithPos {
+  const state = createInitialState();
```
tomasnyberg, line 91: nit: I think the helper methods make it a bit needlessly verbose. Feel free to leave as is though.

packages/language-support/src/formatting/formatting.ts
```diff
@@ -191,66 +190,57 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
-  lastInCurrentBuffer = () => this.currentBuffer().at(-1);
-
-  removeAllPendingGroups = () => {
-    while (this.groupStack.length > 0) {
-      this.endGroup(this.groupStack.at(-1).id);
-    }
-  };
+  lastInChunkList = () => this.chunkList.at(-1);
 
   breakLine = () => {
```
simonthuresson, line 228: Feels like this now has another use than just breakLine. Perhaps change to using mustBreakBetween() in relevant places and rename this to its actually use

packages/language-support/src/formatting/lineFormatter.ts
```diff
@@ -0,0 +1,194 @@
...
+  };
+}
+
+function createInitialState(): State {
+  return {
+    formatted: '',
+    cursorPos: 0,
+    indentation: 0,
+    column: 0,
+    activeGroups: [] as Group[],
```
tomasnyberg, line 131: I don't think the typecasts are needed here?

packages/language-support/src/tests/formatting/linebreaks.test.ts
```diff
@@ -1190,46 +1206,23 @@ RETURN
     verifyFormatting(query, expected);
   });
 
-  test("formats Greg's query in a reasonable way", () => {
+  test('reduce function should not break between name and left paranthesis', () => {
```
tomasnyberg, line 1209: nit typo: `parenthesis`

tomasnyberg (commented):
very nice to get rid of the whole solution search file 👍

packages/language-support/src/formatting/formatting.ts
```diff
@@ -1915,8 +1929,14 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
   // https://neo4j.com/docs/cypher-manual/current/styleguide/#cypher-styleguide-indentation-and-line-breaks
   visitMergeAction = (ctx: MergeActionContext) => {
+    this.mustBreakBetween();
     const mergeActionIndent = this.addIndentation();
-    this.breakAndVisitChildren(ctx);
+    this._visit(ctx.ON());
+    this.avoidBreakBetween();
+    this._visit(ctx.MATCH());
+    this._visit(ctx.CREATE());
+    this.avoidBreakBetween();
```
simonthuresson, line 1938: This avoidBreakBetween feels missplaced, all the tests seems to work without it

packages/language-support/src/formatting/lineFormatter.ts
```diff
@@ -0,0 +1,194 @@
...
+  for (let i = 0; i < chunkList.length; i++) {
+    const chunk = chunkList[i];
+    const nextChunk = chunkList[i + 1];
+
+    applyIndentationIfNeeded(state);
+    checkAndSetCursorPosition(state, chunk);
+    updateActiveGroups(state, chunk);
+    appendChunkText(state, chunk);
+    updateIndentationState(state, chunk);
+
```
simonthuresson, line 102: Remove unecessary space

packages/language-support/src/formatting/formattingHelpers.ts
simonthuresson: Agreed, should be easier.

packages/language-support/src/formatting/lineFormatter.ts
simonthuresson: Agreed, it should be named formatted

packages/language-support/src/formatting/lineFormatter.ts
simonthuresson: String[] is probably faster. Lets focus on performance later though

packages/language-support/src/formatting/lineFormatter.ts
simonthuresson: I agree, they should be filtered out in the visitor. I will leave as is for now though

packages/language-support/src/formatting/lineFormatter.ts
simonthuresson: Hmm, I can see why you think that, I like it however. What does @OskarDamkjaer and @ncordon think?

packages/language-support/src/formatting/lineFormatter.ts
simonthuresson: Nope it is not needed, remedy of letting Clause code for you

simonthuresson:
> Fix this test

Fixed

packages/language-support/src/formatting/lineFormatter.ts
OskarDamkjaer: I also think it's fine to leave as is, but generally I'm not a big fan of small helper functions that mutate their input parameters or depend heavily on stateful `this` members. Usually it means you have to go into each of the helper functions to see what's going on, so the abstraction is does not provide much value. Here's a little contrived example, if you have:

```
 const abc = add(varone, vartwo)
 ```
 
 You don't really need to read the implementation, but if you have just 
 ```
 add(varone, vartwo)
 ```
 
then you kind of have to check what's going on, to follow along the code, the abstraction isn't helping if you need to read the source. That said, it's also a matter of taste, happy for it to remain as is

File: packages/language-support/src/tests/formatting/comments.test.ts [hunk 18]
```diff
@@ -325,6 +325,8 @@ AND post.likes >= 50
    - Adjust the threshold based on campaign feedback.
 */
 RETURN user.username, post.title, post.likes;`;
+    // TODO: It would be nice if the AND here moved up to the previous line, like we do when
+    // there are inline comments. We just haven't implemented it for hard-break ones just yet.
```
OskarDamkjaer, line 329: agreed, good to know it's on your radar

OskarDamkjaer: is it the same for some of the other examples below?

OskarDamkjaer (approved):

packages/language-support/src/tests/formatting/comments.test.ts
simonthuresson: Not when it comes to AND, OR etc etc, as far as I can tell. However we have discussed if the relationship node should be able to do something similar. Because a query becomes very split when you comment between a node and relationship node. 
How it works today: 
```
MATCH
  (p:Project)-
    // A project might have multiple owners
    [:OWNED_BY]->
  (user:User)
```
How it could work:
```
MATCH
  (p:Project)-[:OWNED_BY]->
  // A project might have multiple owners
  (user:User) 
```
Currently we are leaning towards not moving the relationship node with the motivation that what the comment refers to is moved making it less obvious what the comment refers to. Interested to see what you think.

packages/language-support/src/formatting/lineFormatter.ts
OskarDamkjaer: Even if it's as easy or easier to read I find that one line helper functions that aren't reusable (aka depend on state or highly specific) usually make it harder to build higher value abstractions later on in the project life cycle, similar to problems around pre-mature optimization. This article touches a little on it https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction

that said, I think it makes sense to keep as is, it's a tiny detail where people have different opinions, I just like sharing my favorite articles :)

packages/language-support/src/formatting/formatting.ts
```diff
@@ -191,66 +189,55 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
           `Unable to format query due to syntax error near ${this.firstUnParseableToken?.text} at line ${this.firstUnParseableToken?.line}`,
         );
       }
-      throw new Error(INTERNAL_FORMAT_ERROR_MESSAGE);
+      // throw new Error(INTERNAL_FORMAT_ERROR_MESSAGE);
```
ncordon, line 192: Any reason why we commented out this?

ncordon (approved):
LGTM

packages/language-support/src/formatting/formattingHelpers.ts
simonthuresson: No I did address this, but I can see why you think that. The comment or git diff does point at the wrong property. What Tomas meant was the property in Group, which was before allBreaks.

packages/language-support/src/formatting/formatting.ts
simonthuresson: Nice catch! It should not have been commented out.

PR #472, opened 2025-04-10
WIP downloads language server dependent on neo4j server and uses this instead

anderson4j:
Fetches and unpacks files from npm to a .neo4jLang folder in the users home folder. (ex. Users/Isak on Windows)

Switches to use the fitting language server from the downloaded folder(s) when connecting to a neo4j server.

Has the stub for a mapping between neo4j and language server

Very WIP

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,9 @@
         "packages/*",
         "vendor/*"
       ],
+      "dependencies": {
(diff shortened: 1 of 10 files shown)


anderson4j:
We switched from this approach a while ago, so removing for cleanup

PR #473, opened 2025-04-15
Fix e2e tests flakiness

ncordon:
Right now the CI takes ~15 min

### DIFF
--- a/packages/vscode-extension/tests/specs/api/autoCompletion.spec.ts
+++ b/packages/vscode-extension/tests/specs/api/autoCompletion.spec.ts
@@ -247,42 +247,6 @@ suite('Auto completion spec', () => {
     });
   });
 
-  test('Parameters are available in completions even when disconnected from neo4j', async () => {
-    await vscode.commands.executeCommand(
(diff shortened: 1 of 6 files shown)


packages/vscode-extension/tests/specs/api/autoCompletion.spec.ts
```diff
@@ -246,45 +247,6 @@ suite('Auto completion spec', () => {
...
-    await testCompletionContains({
-      textFile: textDocument.uri,
-      position: position,
-      expected: expecations,
-    });
-
-    await vscode.commands.executeCommand(
-      CONSTANTS.COMMANDS.INTERNAL.FORCE_CONNECT,
-      0,
-    );
```
ncordon, line 285: I think there's a race condition between this and the evaluate parameter of the next test. Just because when we connect to a new connection, I think we might not immediately set the `schemaPoller.connection` inside it?

ncordon: So I've swapped the order of the tests, but not sure if there's something else more elegant we could do

ncordon:
Superseded by #474

PR #474, opened 2025-04-16
Improve CI speed

OskarDamkjaer:
Based on #461

- Splits e2e jobs to run in parallel
- Adds cache on commit to avoid rebuilding
- Cancel old jobs on the same pull request
- Fixes parameter can't be set on system flaky test by improving the UX. We not instantly give the error (same as when disconnected), instead of letting the user fill in all the values and then giving an error ✨ 

On flakyness:
- [Ideally each test should be as isolated as possible](https://playwright.dev/docs/best-practices#make-tests-as-isolated-as-possible). This is so they can run / re-run in parallel, etc. This means they need to ensure their own connection etc in a `beforeEach`. I'm not sure if this applies to the vscode tests though as it's not got multiple browser windows like playwright tests do 🤔

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,18 +7,28 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 env:
   NODE_OPTIONS: '--max_old_space_size=8192'
 
 jobs:
-  lint-and-format:
-    name: Run eslint and check formatting
+  build:
+    name: Setup and build project
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-and-build
 
-      - name: Build project
-        uses: ./.github/actions/build-project
+  lint-and-format:
+    name: Run eslint and check formatting
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-and-build
 
       - name: Check there are no linting errors
         run: pnpm lint
@@ -30,38 +40,60 @@ jobs:
 
   unit-test:
     name: Unit tests
+    needs: build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-
-      - name: Build project
-        uses: ./.github/actions/build-project
+      - uses: ./.github/actions/setup-and-build
 
       - name: Run tests
         run: pnpm test
 
-  e2e-test:
-    name: E2E tests
-    timeout-minutes: 60
+  react-codemirror-e2e-test:
+    name: React Codemirror E2E tests
+    needs: build
+    timeout-minutes: 30
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-
-      - name: Build project
-        uses: ./.github/actions/build-project
+      - uses: ./.github/actions/setup-and-build
 
       - name: Install Playwright Browsers
-        run: pnpm exec playwright install --only-shell --with-deps
+        run: pnpm exec playwright install --only-shell --with-deps chromium
         working-directory: packages/react-codemirror
 
       - name: Run e2e test
-        # We need a graphics server to run integration tests with VSCode
-        # https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions
-        run: xvfb-run -a pnpm test:e2e
+        run: pnpm --filter react-codemirror test:e2e
 
-      - uses: actions/upload-artifact@v4
+      - name: Upload Playwright Report
+        uses: actions/upload-artifact@v4
         if: always()
         with:
           name: playwright-report
           path: packages/react-codemirror/playwright-report/
           retention-days: 30
+
+  vscode-webview-test:
+    name: VSCode Webview tests
+    needs: build
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-and-build
+
+      - name: Run e2e test
+        # We need a graphics server to run integration tests with VSCode
+        run: xvfb-run -a pnpm --filter neo4j-for-vscode test:webviews
+
+  vscode-api-and-unit-tests:
+    name: VSCode API and Unit tests
+    needs: build
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-and-build
+
+      - name: Run tests
+        run: xvfb-run -a pnpm --filter neo4j-for-vscode test:apiAndUnit
--- a/.github/workflows/publish-vscode-extension.yaml
+++ b/.github/workflows/publish-vscode-extension.yaml
@@ -17,8 +17,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Build project
-        uses: ./.github/actions/build-project
+      # todo verify caching works well here too
+      - name: Setup and build project
+        uses: ./.github/actions/setup-and-build
 
       - name: Publish to VSCode Marketplace
         env:
--- a/packages/react-codemirror/package.json
+++ b/packages/react-codemirror/package.json
@@ -28,7 +28,7 @@
     "test": "vitest run",
     "test:e2e": "playwright test -c playwright-ct.config.ts",
     "test:e2e-ui": "playwright test -c playwright-ct.config.ts --ui",
-    "benchmark": "BENCHMARKING=true npx playwright test -c playwright-ct.config.ts -g \"benchmarking & performance test session\""
+    "benchmark": "BENCHMARKING=true playwright test -c playwright-ct.config.ts -g \"benchmarking & performance test session\""
   },
   "repository": {
     "type": "git",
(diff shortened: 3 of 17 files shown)


.github/workflows/publish-vscode-extension.yaml
```diff
@@ -17,8 +17,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Build project
-        uses: ./.github/actions/build-project
+      # todo verify caching works well here too
+      - name: Setup and build project
+        uses: ./.github/actions/setup-and-build
```
OskarDamkjaer, line 22: not sure how to best check for this

packages/react-codemirror/playwright-ct.config.ts
```diff
@@ -42,16 +42,7 @@ export default defineConfig({
...
-    {
-      name: 'chromium',
-      use: { ...devices['Desktop Chrome'] },
-    },
-    {
-      name: 'firefox',
-      use: { ...devices['Desktop Firefox'] },
-    },
-  ],
+  projects: [{ name: 'chromium', use: { ...devices['Desktop Chrome'] } }],
```
OskarDamkjaer, line 47: i think it makes sense to create a test suite dedicated for webkit + safari at a later stage and avoid the flakyness from firefox now. also happy to add firefox back if you prefer

.github/workflows/publish-vscode-extension.yaml
ncordon: We could just make a card and check the next time we do a VSCode release?

packages/react-codemirror/package.json
```diff
@@ -28,7 +28,7 @@
     "test": "vitest run",
     "test:e2e": "playwright test -c playwright-ct.config.ts",
     "test:e2e-ui": "playwright test -c playwright-ct.config.ts --ui",
-    "benchmark": "BENCHMARKING=true npx playwright test -c playwright-ct.config.ts -g \"benchmarking & performance test session\""
```
ncordon, line 31: Don't we need `pnpx` here?

packages/react-codemirror/playwright-ct.config.ts
ncordon: I think we have firefox here because at some point we broke something in the semantic analysis that was firefox specific. To avoid regressions I think it'd be nice to still keep it in the CI?

packages/vscode-extension/src/commandHandlers/params.ts
```diff
@@ -103,6 +111,15 @@ export async function editParameter(paramItem: ParameterItem): Promise<void> {
     );
     return;
   }
+
+  const db = getCurrentDatabase();
+  if (db.type === 'system') {
+    void window.showErrorMessage(
+      'Parameters cannot be edited when on the system database. Please connect to a user database.',
+    );
```
ncordon, line 119: Also extract this?

packages/vscode-extension/tests/specs/webviews/params.spec.ts
```diff
@@ -68,10 +69,11 @@ suite('Params panel testing', () => {
   }
 
   async function forceDisconnect() {
-    await browser.executeWorkbench(async (vscode) => {
+    void browser.executeWorkbench((vscode) => {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
-      await vscode.commands.executeCommand('neo4j.internal.forceDisconnect');
+      void vscode.commands.executeCommand('neo4j.internal.forceDisconnect');
     });
```
ncordon, line 75: At some point I was using the `void`s as well, but that meant we had a notification that unless closed, the method wasn't finishing? 

I think we should just swap all of the:

```
await window.showErrorMessage
```

for:

```
void window.showErrorMessage
```

?

packages/vscode-extension/tests/webviewUtils.ts
```diff
@@ -212,3 +216,18 @@ export async function executeFile(
...
+export async function ensureNotificationsAreDismissed(
+  browser: WebdriverIO.Browser,
+): Promise<void> {
+  const wb = await browser.getWorkbench();
+  const notifications = await wb.getNotifications();
+  for (const notification of notifications) {
+    await notification.dismiss();
+  }
+
+  const remainingNotifications = await wb.getNotifications();
```
ncordon, line 229: Interesting. Why do we need to get notifications twice?

.github/workflows/ci.yaml
```diff
@@ -7,18 +7,28 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
```
ncordon, line 13: I like this because it prevents us from kicking jobs for different commits at the same time, but there's one thing about this I don't quite like: if two different jobs fail for the latest commit, you either have to retrigger them all or re-trigger one, wait for that to pass and re-trigger the second one.

I've looked it up and there's no way to tweak the concurrency per individual job, only per workflow. Any thoughts?

packages/vscode-extension/tests/webviewUtils.ts
OskarDamkjaer: In my testing overflowing ones didn't close get dismissed, so I needed to call it more than once. I think it makes sense to take some time and look at the docs to see if there's some "clear notifications" thing we could run instead

packages/vscode-extension/tests/specs/webviews/params.spec.ts
OskarDamkjaer: >we had a notification that unless closed, the method wasn't finishing?
what does this mean? This method still blocks like before due to the new `await waitUntilNotification` below

packages/vscode-extension/src/commandHandlers/params.ts
OskarDamkjaer: makes sense :+1:

packages/vscode-extension/src/commandHandlers/params.ts
```diff
@@ -54,6 +54,14 @@ export async function addParameter(defaultParamName?: string): Promise<void> {
     return;
   }
 
+  const db = getCurrentDatabase();
+  if (db.type === 'system') {
+    void window.showErrorMessage(
+      'Parameters cannot be added when on the system database. Please connect to a user database.',
+    );
```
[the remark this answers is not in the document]
OskarDamkjaer: The reason I made them slightly different was because the evaluating is at a later stage, differentiating between trying to add something and trying to validate the type of something while adding. Not sure if anyone really cares about the difference, so we could make both say "cannot be added"

packages/vscode-extension/package.json
```diff
@@ -426,8 +426,8 @@
     "build:dev": "tsc -b && pnpm gen-textmate && pnpm bundle-extension:dev && pnpm bundle-language-server && pnpm bundle-webview-controllers",
     "clean": "rm -rf dist",
     "test:e2e": "pnpm build:dev && pnpm test:apiAndUnit && pnpm test:webviews",
-    "test:apiAndUnit": "pnpm build:dev && rm -rf .vscode-test/user-data && node ./dist/tests/runApiAndUnitTests.js",
-    "test:webviews": "pnpm build:dev && wdio run ./dist/tests/runWebviewTests.js"
```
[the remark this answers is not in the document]
OskarDamkjaer: The cache isn't as aggressive as turborepo, so this will rebuild. It's also rebuilding a dev build instead of a production build so we're not testing the same behavior that we'd send out to production. 

For convenience locally it'd make more sense to have a new command, maybe `test:webviews:watch` or `test:webviews:dev` which rebuilds a development build and then runs the tests in some easy to debug mode

packages/react-codemirror/package.json
OskarDamkjaer: `pnpx` will always [download the latest from the internet](https://pnpm.io/cli/dlx) and not use the locally installed version

OskarDamkjaer: For calling dependencies, we don't need a prefix (see how we call eslint, esbuild, playwright in the above command etc)

packages/react-codemirror/playwright-ct.config.ts
OskarDamkjaer: We didn't fix the semantic analysis in firefox, it's still not working. I'll add it back :+1:

ncordon (commented):
So excited about this!!!! 🚀

packages/vscode-extension/tests/specs/webviews/params.spec.ts
ncordon: I'm not implying we should change anything here but that we should replace the `await window.showErrorMessage` with `void window.showErrorMessage` everywhere in the codebase? I don't see any reason why that should block the execution in general of some method when things went wrong

packages/react-codemirror/playwright-ct.config.ts
ncordon: > We didn't fix the semantic analysis in firefox, it's still not working.

Not completely you are right. But it does work in my Firefox, when it wasn't before. We definitely fixed something that wasn't working at some point that had to do with us setting the schema eagerly.

ncordon:
I'm going to merge this, given the CI is super flaky, we can address anything extra in other PRs

ncordon (approved):

packages/vscode-extension/tests/specs/webviews/params.spec.ts
OskarDamkjaer: i mean the test will fail if the show error message goes wrong no? The very next assertion is waiting for it to fire?

PR #475, opened 2025-04-16
update exports for noah

OskarDamkjaer:
For a Genai project in dashboards, @noahmay needs more of the Antlr4 exports. They need some basic schema aware linting (which we're not ready to add in a while), these extra exports lets them add some simple functionality in Dashboards without duplicating the parser from the one they already use through the react editor


noahmay (approved):
🙇

noahmay:
I assume a changeset I needed

PR #476, opened 2025-04-17
Fixes small details in the CI

ncordon:
This pr fixes a few details in the CI:

* The VSCode release job was not working
* The Archive VSCode artifacts job was not working either
* Adds back Firefox for the e2e tests (only in main)


packages/react-codemirror/playwright-ct.config.ts
```diff
@@ -14,9 +14,9 @@ export default defineConfig({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  retries: 2,
+  /* Use parallel testing */
+  workers: undefined,
```
OskarDamkjaer, line 19: Fully parallel isn't recommended on CI unless you have powerful machines, as you're already using much of the computing power. We can measure, but I think the stability vs. time computation isn't there unless we switch to a larger machine, see this github issue asking about the default on their page 

https://github.com/microsoft/playwright/issues/22666

packages/react-codemirror/playwright-ct.config.ts
```diff
@@ -14,9 +14,9 @@ export default defineConfig({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  retries: 2,
```
OskarDamkjaer, line 17: Why do you want retries when running locally?

scripts/clean.js
```diff
@@ -0,0 +1,33 @@
+// This script is run from package.json in the root of the project before
```
OskarDamkjaer, line 1: Usually [rimraf](https://www.npmjs.com/package/rimraf) is used for this purpose (cross OS file removal)

.github/actions/codemirror-e2e-tests/action.yaml
```diff
@@ -0,0 +1,34 @@
...
+      shell: bash
+      run: |
+        if [[ "${{ inputs.browser }}" != "chromium" && "${{ inputs.browser }}" != "firefox" ]]; then
+          echo "Invalid browser (should be one of firefox or chromium): ${{ inputs.browser }}"
+          exit 1
+        fi
+    
+    - name: Install Playwright Browsers
+      shell: bash
+      run: pnpm exec playwright install --only-shell --with-deps
```
OskarDamkjaer, line 21: it's good to specify firefox & chrome here, the default install includes webkit which makes the install take quite a bit longer

.github/workflows/fetch-dev-artifacts.yaml
```diff
@@ -29,98 +29,96 @@ jobs:
...
-      run: |
-        BRANCH_NAME="update-artifacts-$(date +%Y%m%d-%H%M%S)"
-        echo "BRANCH_NAME=$BRANCH_NAME" >> $GITHUB_ENV
+      - name: Build project
+        uses: ./.github/actions/setup-and-build-project
+      
+      - name: Create New Branch Name
+        run: |
+          BRANCH_NAME="update-artifacts-$(date +%Y%m%d-%H%M%S)"
+          echo "BRANCH_NAME=$BRANCH_NAME" >> $GITHUB_ENV
```
ncordon, line 97: GitHub is getting confused but the only changes to this file should have been adding this step to build the project (because we need the new generated TextMate grammar)

.github/workflows/fetch-dev-artifacts.yaml
```diff
@@ -29,98 +29,96 @@ jobs:
...
-        git add ./packages/language-support/src/antlr-grammar
-        git add ./packages/language-support/src/syntaxValidation
-        git commit -m "Updates artifacts to those from latest dev snapshot (${{ env.neo4j_version_base }}-SNAPSHOT)"
-        git push origin HEAD:${{ env.BRANCH_NAME }}
+      - name: Commit and Push Changes
+        if: steps.git-check.outputs.modified == 'true'
+        run: |
+          git add ./packages/language-support/src/antlr-grammar
+          git add ./packages/language-support/src/syntaxValidation
+          git add ./packages/vscode-extension/syntaxes/cypher.json
```
ncordon, line 111: And adding the updated TextMate grammar here

.github/workflows/publish-npm-packages.yaml
```diff
@@ -29,6 +29,8 @@ jobs:
     environment: publish-npm
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v4
```
ncordon, line 32: Not sure what we change but the publishing to npm step complains we are not in a git repo and it's broken in main 🤷‍♂️ :

Screenshot 2025-04-22 at 11 28 56

packages/react-codemirror/playwright-ct.config.ts
```diff
@@ -44,5 +44,14 @@ export default defineConfig({
...
 
-  projects: [{ name: 'chromium', use: { ...devices['Desktop Chrome'] } }],
+  projects: [
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] },
+    },
+    {
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
```
ncordon, line 54: Adds back firefox, but I've only enabled the execution of the job in `main`, not in PRs

packages/vscode-extension/package.json
```diff
@@ -22,6 +22,7 @@
   ],
   "icon": "resources/images/logo.png",
   "main": "./dist/extension.js",
+  "activationEvents": [],
```
ncordon, line 25: This we get a complain in the new version if we don't add it

packages/vscode-extension/package.json
```diff
@@ -424,6 +425,7 @@
     "bundle-webview-controllers": "esbuild ./src/webviews/controllers/* --bundle --outdir=dist/webviews/ --platform=browser",
     "build": "tsc -b && pnpm gen-textmate && pnpm bundle-extension && pnpm bundle-language-server && pnpm bundle-webview-controllers",
     "build:dev": "tsc -b && pnpm gen-textmate && pnpm bundle-extension:dev && pnpm bundle-language-server && pnpm bundle-webview-controllers",
+    "package": "pnpm vsce package --pre-release --no-dependencies",
```
ncordon, line 428: The release job to VSCode was broken because the tool to generate the VSCode packages does not work with pnpm: https://github.com/microsoft/vscode-vsce/issues/421.

Luckily we are bundling the plugin so it works if we pass the flag `--no-dependencies` but this is one to keep track of.

scripts/clean.js
ncordon: I just lifted the upx script. I've deleted it and changed the `package.json` to use `rimraf` now

packages/react-codemirror/playwright-ct.config.ts
ncordon: I don't think it was making the jobs faster anyway because our GitHub machines only have 2 cores

.github/workflows/publish-npm-packages.yaml
OskarDamkjaer: Would make sense to set to `onLanguage:cypher` maybe? 

https://code.visualstudio.com/api/references/activation-events#onLanguage

packages/vscode-extension/package.json
OskarDamkjaer: Oh that's unfortunate, thanks for looking into it

packages/vscode-extension/package.json
OskarDamkjaer: I think it'd make sense to add an activation event for onLanguage:cypher? https://code.visualstudio.com/api/references/activation-events#onLanguage

ncordon: > Note: Beginning with VS Code 1.74.0, languages contributed by your extension do not require a corresponding onLanguage activation event declaration for your extension to be activated.

We've never needed it but I'll add it

PR #478, opened 2025-04-22
Improve List/Map Formatting

simonthuresson:
# Description

This PR improves how List:s, Map:s and Map Projections are being formatted in certain cases. This was a tech debt from #460. 
The general idea of this PR is to fix the following:
```
// Before
RETURN 
  [
    item1,
    item2, 
    item3
  ]
// After
RETURN [
  item1,
  item2, 
  item3
]
```

There are a few caveats to this. This type of formatting only occurs if the List/Map is "alone". Meaning if two lists would occur in the return, the regular formatting applies. Which is similar to how prettier does it:
```
RETURN 
  [
    item1,
    item2, 
    item3
  ],
  [
    item4,
    item5, 
    item6
  ]
```
Another case is if a comment appear, then the regular style is also applied. This is because otherwise the comment will end up inside the list, which may cause the comment to refer to something else. 
This is how we do it:
```
RETURN // A Comment in weird place
  [
    item1,
    item2, 
    item3
  ]
```

## Indentation
The most challening thing about this PR was how to handle indentation. In the regular case, both the return statement and the list adds one level of indentation. However, with the "special" formatting, only one should be applied. If you just ignore adding of the the indentation levels, the formatter throws an error. This is by design to not allow removing indentation levels which has not been applied. The solution to this was to let the add indentation part include a reference to the remove indentation modifier, if the add indentation is not applied, a flag is set on the remove indentation. 

## Tests
 - About 10 tests has been modified to suit the new formatting style
 - About 10 tests added


packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -39,6 +39,8 @@ export interface BaseChunk {
   // that is in the chunklist (one with a newline before it.)
   comment?: string;
   mustBreak?: boolean;
+  specialSplit?: boolean;
+  oneItem?: boolean;
```
simonthuresson, line 43: Very unsure about these names, but have not come up with anything better.

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -65,6 +66,8 @@ export type Chunk = RegularChunk | CommentChunk | SyntaxErrorChunk;
 export interface IndentationModifier {
   id: number;
   change: 1 | -1;
+  removeIndentation?: IndentationModifier;
+  appliedIndentation: boolean;
```
simonthuresson, line 70: These names might not be the best either.

packages/language-support/src/formatting/formatting.ts
```diff
@@ -1904,6 +1910,12 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     }
     this.avoidSpaceBetween();
     this.removeIndentation(argsIndent);
+    // Check for right paranthesis
```
tomasnyberg, line 1913: nit: `paranthesis` -> `parenthesis`

tomasnyberg: same below

packages/language-support/src/formatting/formatting.ts
```diff
@@ -2013,6 +2025,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   visitMap = (ctx: MapContext) => {
     const mapGrp = this.startGroup();
     this._visit(ctx.LCURLY());
+    this.lastInChunkList().specialSplit = true;
```
tomasnyberg, line 2028: could this use the `setChunkProperty` helper we have?

packages/language-support/src/formatting/formatting.ts
```diff
@@ -2263,6 +2278,17 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     this.endGroup(callGrp);
     this.removeIndentation(callIndent);
   };
+
+  visitStringAndListComparison = (ctx: StringAndListComparisonContext) => {
```
tomasnyberg, line 2282: Are there any other places where a list can occur where it makes sense to have the `specialSplit` behavior? Alternatively, how did we decide where this special split behavior can happen?

I think we discussed this offline, but might be worth mentioning in the description

packages/language-support/src/formatting/formattingHelpers.ts
simonthuresson: hasSingleArgument and isBreakEligible?

packages/language-support/src/formatting/formattingHelpers.ts
simonthuresson: closingModifier and isApplied?

packages/language-support/src/formatting/layoutEngine.ts
```diff
@@ -148,23 +158,34 @@ function validateFinalState(state: State) {
   }
 }
 
+function applySpecialBreak(state: State, chunk: Chunk, nextChunk: Chunk) {
```
tomasnyberg, line 161: could this function clarify what all the booleans mean conceptually?  I don't really understand why we have to check for no comments, and whether the next groups are going to break. 

Also, not a huge fan of doing the overflowing calculation in here, I feel like it is risky to introduce that logic in more than one place (based on what we found previously)

packages/language-support/src/tests/formatting/edgecases.test.ts
```diff
@@ -691,18 +689,17 @@ RETURN f`;
...
-    .age,
-    .email,
-    .phone,
-    .address,
-    .occupation,
-    .nationality,
-    .birthdate,
-    .gender
-  } AS personInfo`;
+RETURN p {
```
tomasnyberg, line 692: nice 👍  I don't think the description mentions the choice to keep the map projection variable name thing on the same line as well though.

packages/language-support/src/formatting/formattingHelpers.ts
tomasnyberg: I can't really come up with anything better either, they're not awful though.

But I would appreciate a couple comments describing what they are for, can that be added? With that I think it's fine

packages/language-support/src/formatting/formatting.ts
```diff
@@ -369,10 +371,12 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     if (this.indentStack.length === 0 || this.indentStack.at(-1).id !== id) {
       throw new Error(INTERNAL_FORMAT_ERROR_MESSAGE);
     }
+    const lastModifier = this.indentStack.pop();
     const modifier: IndentationModifier = {
-      ...this.indentStack.pop(),
+      ...lastModifier,
       change: -1,
     };
+    lastModifier.removeIndentation = modifier;
```
tomasnyberg, line 379: this probably shouldn't have the same name as `this.removeIndentation`

tomasnyberg (commented):
big improvement for lists and maps 🔥 

discussed a bit offline to look at whether we can do without the reference check, will re-review

packages/language-support/src/formatting/formatting.ts
simonthuresson: Good catch

packages/language-support/src/formatting/formatting.ts
simonthuresson: This is the places I know of so far. It might be more places, but I think they will have to appear as the formatter is progressively more used.

packages/language-support/src/tests/formatting/edgecases.test.ts
simonthuresson: No it didnt, i have added a part which explain how and when a special split can appear.

packages/language-support/src/formatting/layoutEngine.ts
simonthuresson: I'll add comments to explain them.
The overflowing calculation was redundant, now removed.

simonthuresson: Added comment, re-review is required

ncordon (approved):

PR #479, opened 2025-04-22
Bump typescript to latest

OskarDamkjaer:
This lets us use [project references](https://www.typescriptlang.org/docs/handbook/project-references.html), which means we ensure caching of the `tsc` commands between builds. In my testing we cut ~20% of build times with this new configuration (most of the time is spend with esbuild bundling, more on that later)


packages/vscode-extension/tsconfig.json
```diff
@@ -4,12 +4,6 @@
...
     "outDir": "dist",
     "jsx": "react-jsx",
-    "types": [
-      "@wdio/mocha-framework",
-      "wdio-vscode-service",
-      "@wdio/globals/types",
-      "@wdio/types"
-    ],
-    "typeRoots": ["../../node_modules/@types", "../../node_modules"]
+    "types": ["@wdio/globals/types"]
```
OskarDamkjaer, line 7: From the docs on wdio 8 with typescript

https://webdriver.io/docs/typescript/#framework-setup

ncordon (approved):
I tried it out locally and seems to be working fine!

PR #480, opened 2025-04-22
Publish unbundled packages for ESM builds

OskarDamkjaer:
When we publish all the files and `sideEffects: "false"` in our package.json, the bundling tools will automatically exclude code that's not used in the target bundle. This means we (and most consumers) cut 4MB off our bundle size in `react-codemirror-playground` since it no longer bundles the semantic analysis 

When fiddling with the exports, I also folded in [this change](https://github.com/neo4j/cypher-language-support/pull/475/files) for @noahmay


packages/language-support/package.json
```diff
@@ -51,9 +52,8 @@
   },
   "scripts": {
     "gen-parser": "cross-env ANTLR4_TOOLS_ANTLR_VERSION=4.13.2 antlr4 -Dlanguage=TypeScript -visitor src/antlr-grammar/CypherCmdLexer.g4 src/antlr-grammar/CypherCmdParser.g4 -o src/generated-parser/ -Xexact-output-dir",
-    "build": "cd ../../vendor/antlr4-c3/ && pnpm build && cd - && pnpm gen-parser && concurrently 'npm:build-types' 'npm:build-esm' 'npm:build-commonjs'",
-    "build-types": "tsc --emitDeclarationOnly --outDir dist/types",
-    "build-esm": "esbuild ./src/index.ts --bundle --format=esm --sourcemap --outfile=dist/esm/index.mjs",
+    "build": "cd ../../vendor/antlr4-c3/ && pnpm build && cd - && pnpm gen-parser && concurrently 'npm:build-esm-and-types' 'npm:build-commonjs'",
+    "build-esm-and-types": "tsc --module esnext --outDir dist/esm/project/language-support && cp src/syntaxValidation/semanticAnalysis.js dist/esm/project/language-support/src/syntaxValidation/semanticAnalysis.js && mkdir -p dist/esm/vendor/antlr4-c3/dist/ && cp -r ../../vendor/antlr4-c3/dist/esm  dist/esm/vendor/antlr4-c3/dist/",
```
OskarDamkjaer, line 56: When not bundling, we can just do `tsc` for the build :100: 

However, the `semanticAnalysis.js` is ignored by the typescript config and it seemed the path outside the folder also didn't get built, so I added this to copy them in manually. The odd paths is for the import `../../../../vendor/etc` to be the same in src and dist

packages/language-support/package.json
```diff
@@ -23,16 +23,17 @@
...
     ".": {
       "require": "./dist/cjs/index.cjs",
-      "import": "./dist/esm/index.mjs",
+      "import": "./dist/esm/project/language-support/src/index.js",
       "default": "./dist/cjs/index.cjs",
-      "types": "./dist/types/src/index.d.ts"
+      "types": "./dist/esm/project/language-support/src/index.d.ts"
     }
   },
+  "sideEffects": false,
```
OskarDamkjaer, line 36: This allows bundlers to remove all code that's not imported

OskarDamkjaer:
@ncordon Seems there's something in the vscode extension loading the wrong version of the language support project. I'll give it a look, hopefully a small change :+1:

ncordon (commented):
I still see the language-server is still chunky 😢 . There's no way to fix that with these changes?

```
dist/cypher-language-server.js  8.4mb ⚠️
dist/lintWorker.js  7.6mb ⚠️
```

The `lintWorker` I would expect it to still be that size but not the language server.

PR #485, opened 2025-04-24
Add support for the access-mode console command to the grammar

daveajrussell:
This adds `:access-mode read/write` to the grammar.

Access mode is a console command that allows users to dynamically switch the transaction mode of the query they are executing between read and write mode.


packages/language-support/src/autocompletion/completionCoreCompletions.ts
daveajrussell, line 1: this provides a label override to display 'access-mode' in the completions

else "accessmode" is shown, itself leading to an erroneous lexer error "did you mean accessmode"

packages/language-support/src/antlr-grammar/CypherCmdParser.g4
```diff
@@ -50,13 +51,23 @@ welcomeCmd: WELCOME;
 
 playCmd: PLAY;
 
+accessModeCompletionRule: ACCESSMODE;
+
+accessModeArgs: (readCompletionRule | writeCompletionRule);
+
+accessModeCmd: accessModeCompletionRule accessModeArgs;
```
OskarDamkjaer, line 58: I think this does not allow `:access-mode` (without the read/write part) which we are fine with in query, so I think `accessModeCompletionRule accessModeArgs?;` is needed. It'd be good to get that case covered in a test too

packages/language-support/src/autocompletion/completionCoreCompletions.ts
OskarDamkjaer: We have some other cases of label being incorrect for keywords in `lexersymbols.ts`:

```
export const hasIncorrectSymbolicName: Record<number, string> = {
  [CypherLexer.ALL_SHORTEST_PATHS]: 'allShortestPaths',
  [CypherLexer.SHORTEST_PATH]: 'shortestPath',
  [CypherLexer.LIMITROWS]: 'LIMIT',
  [CypherLexer.SKIPROWS]: 'SKIP',
};
```

It could be that it's a similar problem and could be solved in the same way

OskarDamkjaer: It feels like an `[CypherLexer.ACCESSMODE]: 'access-mode'` here could do the trick, if i understand the issue correctly

packages/language-support/src/antlr-grammar/CypherCmdParser.g4
daveajrussell: sure thing - I figure `:access-mode` on its own should switch to `write` mode?

packages/language-support/src/autocompletion/completionCoreCompletions.ts
daveajrussell: yep - this works perfectly, thanks!

PR #487, opened 2025-04-25
Add file watching with automated reloads for VSCode & Codemirror

OskarDamkjaer:
I also update the eslint config to not ignore files when it's not needed. Instead I've included typescript configs for them, which means eslint can pick up and lint them


package.json
```diff
@@ -13,16 +13,17 @@
     "prepare": "husky install",
     "build": "pnpm --recursive build",
     "build-vscode": "pnpm --filter='neo4j-for-vscode...' build",
-    "dev-codemirror": "pnpm --filter='...@neo4j-cypher/react-codemirror-playground' build && pnpm --filter='@neo4j-cypher/react-codemirror-playground' dev",
+    "dev-codemirror": "pnpm --recursive --parallel --filter='@neo4j-cypher/react-codemirror-playground...' dev",
+    "dev-vscode": "pnpm --recursive --parallel --filter='neo4j-for-vscode...' dev",
```
OskarDamkjaer, line 17: This means run `dev` in `neo4j-for-vscode` and all it's dependencies in parallel

package.json
```diff
@@ -13,16 +13,17 @@
...
     "release": "pnpm build && pnpm publish -r --tag next --access=public",
     "snapshot-release": "pnpm changeset pre exit && pnpm changeset version --snapshot canary && pnpm install --frozen-lockfile=false && pnpm build && pnpm publish -r --tag canary --no-git-checks --access=public",
     "deep-clean": "rimraf -g **/{node_modules,dist}",
     "clean": "pnpm --recursive --parallel clean",
     "test": "pnpm --recursive --parallel test",
     "test:e2e": "pnpm --recursive test:e2e",
     "test:formattingIntegrity": "ts-node ./packages/language-support/src/tests/formatting/verification/verificationCheck.ts",
-    "lint": "eslint . --ext .ts",
-    "lint-fix": "eslint . --fix --ext .ts",
+    "lint": "eslint . --ext .ts,.tsx,.js,.mts",
```
OskarDamkjaer, line 25: Now we lint all project files, not just `.ts`

packages/language-support/package.json
```diff
@@ -27,10 +27,10 @@
   "types": "./dist/esm/project/language-support/src/index.d.ts",
   "exports": {
     ".": {
+      "types": "./dist/esm/project/language-support/src/index.d.ts",
```
OskarDamkjaer, line 30: apparently the order matters.. So types should go first and default last (we had warning in console on build about this)

packages/vscode-extension/esbuild-extension.mts
```diff
@@ -0,0 +1,23 @@
+import * as esbuild from 'esbuild';
+const production = process.argv.includes('--production');
+
+const ctx = await esbuild.context({
```
OskarDamkjaer, line 4: We can do top level await & esm imports in node because we have a modern node version 💯 

The issue with esm importing files in node is that all source files need the `.js` extensions, but we don't import any files here (only dependencies). So we're alright

packages/vscode-extension/package.json
```diff
@@ -419,12 +419,13 @@
     "gen-textmate": "node ./genTextMate.js",
     "vscode:prepublish": "pnpm build",
     "bundle-language-server": "cd ../language-server && pnpm bundle && cp dist/cypher-language-server.js ../vscode-extension/dist/ && pnpm bundle-worker && cp dist/lintWorker.js ../vscode-extension/dist/",
-    "bundle-extension": "node ./esbuild-extension.js --production",
+    "bundle-extension": "node --experimental-strip-types esbuild-extension.mts --production",
```
OskarDamkjaer, line 422: This allows us to skip using `ts-node` for scripts

packages/vscode-extension/src/extension.ts
```diff
@@ -64,6 +71,25 @@ export async function activate(context: ExtensionContext) {
...
+
+    watcher.onDidChange(() => {
+      void window.showInformationMessage('Extension rebuilt, reloading...');
+      void commands.executeCommand<void>(
+        'workbench.action.restartExtensionHost',
+      );
+    });
+
+    context.subscriptions.push(watcher);
+  }
```
OskarDamkjaer, line 92: All the docs I've found state there's no built in / clean way to do this. I'm open to removing it and requireing the dev to manually refresh the extensions instead if you prefer

PR #493, opened 2025-04-28
[WIP] Adding UI components package

nnecla:
This pr is work in progress at the moment. Don't merge it 🙃

This pr is to add the ui components package so that we can have the nvl visualization in the vscode extension. These changes are far from being done but here is a summary:

- **packages/**
  - **ui-components/**
    - **index.html:** To make it easier to develop components, renders `preview.tsx` and can be used by `pnpm dev` within `ui-components` package.
    - **package.json:** There are two important scripts here, `build:css` and `postbuild:css`. They are used to copy the tailwind styles to the vscode extension.
    - **src/**
      - **components/**
        - **nvl-graph-viz/:** Contains the main nvl related components, like the actual visualization, sidebar, overview and details. Mostly just copied from the UPX repository with lots of comments for the time being.
        - **shared/:** These are actually sub components used by the nvl related components, not sure of the naming of the folder :/ 

You can test the changes, by adding the following flag to your settings.json in vscode:
```
  "neo4j.features": {
    "enableVisualization": true
  }
```

**What's missing:** _(not written by ai, i've actually put the emojis, who knows, may be i will be a linkedin influencer one day 🔆 )_
- Relationships 👪 
- UI controls 🎮 
- Styling updates 🎨 
- Word color library and graph styling 🌈 

**A look from my local:**

https://github.com/user-attachments/assets/852bede6-14f4-43c1-8390-cd3736aa2941


packages/ui-components/LICENSE.md
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
OskarDamkjaer, line 1: I'm not 100% sure we can have this license here, i'll create a chat with us and greg

packages/ui-components/.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules/
+test-results/
+playwright-report/
+playwright/.cache/
```
OskarDamkjaer, line 4: can these go in the top level git ignore?

packages/ui-components/CHANGELOG.md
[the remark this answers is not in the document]
nnecla: Sorry, leftover from a copy paste 🙈 Removed it now

packages/ui-components/.gitignore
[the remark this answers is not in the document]
nnecla: Yes, changed it now!

pnpm-workspace.yaml
```diff
@@ -4,3 +4,5 @@ packages:
 
 catalog:
   'neo4j-driver': '5.12.0'
+  '@neo4j-ndl/base': '^3.4.1'
```
nnecla, line 7: I had some problems with the ndl styles, some tags did not have the background, and graph chips for relationships were looking funky. I've realised that the ndl css we link from the results webview was not up to date, and so, the nvl graph viz was looking differently on its own env and on the vs code extension.

I believe, this would be ok for now, so that it would keep the styles the same(as long as the ndl.css is generated on update). However, in the long term, we should find a similar solution to the tailwind styles. We should be able to generate these css files(which are to be linked from webviews) within ui-components library and export then to the consumer packages, for both tailwind and ndl styles.

packages/vscode-extension/src/webviews/controllers/resultTabs.tsx
```diff
@@ -51,8 +52,8 @@ function renderTable(rows: ResultRows) {
     <table>
       <thead>
         <tr>
-          {Object.keys(rows[0]).map((key) => (
-            <th>{key.toString()}</th>
+          {Object.keys(rows[0]).map((key, i) => (
```
OskarDamkjaer, line 55: I'd like to avoid using the indexes for keys as it's going to create bugs if we'd ever we re-order or remove items from the list. For things that don't have a natural key, I think we should do something like this 

```ts
  function withKey<A>(value: A): { key: string; value: A } {
    return { key: /* id generator like nanoid*/, value };
  }

  const mapItems = useMemo(() => originalItems.map(withKey), [row]);

File: packages/ui-components/package.json [hunk 5]
```diff
@@ -0,0 +1,61 @@
...
+    "@neo4j-ndl/react": "catalog:",
+    "@neo4j-nvl/base": "0.3.8",
+    "@neo4j-nvl/react": "0.3.8",
+    "@tailwindcss/vite": "^4.1.4",
+    "classnames": "2.5.1",
+    "neo4j-driver": "catalog:",
+    "re-resizable": "6.11.2",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "tailwindcss": "^3.4.17"
```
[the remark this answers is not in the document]
OskarDamkjaer: nvm let's fix in a followup pr

OskarDamkjaer (approved):
Looked nice when I tried it, let's fix the CSS bundling stuff in another PR. Good to merge after we fix the issues with `index` keys

PR #495, opened 2025-04-30
Updates language-support package type and node version

anderson4j:
The problem:
Running e2e tests in the react-codemirror package fails on main with a message like:
`SyntaxError: Named export 'testData' not found. The requested module '@neo4j-cypher/language-support' is a CommonJS module, which may not support all module.exports as named exports.`
This testData file is part of the language-support package, and we use an ESModule import to access it in the react-codemirror tests.

The fix:
Adding "type": "module" in the package.json of language-support makes it so .js files (the ones we build with the `build-esm-and-types` script) are treated as ESModule instead of CommonJS. 
In the `build-commonjs`  script the output file is already .cjs instead of .js so this should not interfere with the actual CommonJS file.

The node version we were using was also a bit old, and Oskar had suggested updating it, so I shifted it to LTS.


.github/actions/setup-and-build/action.yaml
```diff
@@ -17,7 +17,7 @@ runs:
     - name: Setup node.js
       uses: actions/setup-node@v4
       with:
-        node-version: 20
+        node-version: 22
```
anderson4j, line 20: I tried this locally after Oskar's tip, and it does seem to improve flakiness, if not fully fixing it...

OskarDamkjaer (approved):

PR #499, opened 2025-05-08
Updates vscode extension version to 1.11.0

anderson4j:


packages/vscode-extension/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # neo4j-for-vscode
 
+## 1.10.2
+
+- Adds support for the access-mode console command
+- Updates grammar and semantic analysis
+- Updates formatter handling of CALL
+
 ## 1.10.0
```
anderson4j, line 9: There was no entry for 1.10.1 but a 1.10.1 release is up on vscode marketplace. Difference between 1.10.0 and 1.10.1 looks to just be changes in build files, so I figure there's nothing to put into a 1.10.1 entry if we wanted one - maybe that's why it's missing

ncordon: Maybe we should have 1.11.0 instead given we are updating the grammar and the semantic analysis

ncordon (approved):

PR #503, opened 2025-05-13
Displays received error/warning messages in playwright

anderson4j:
Updates the helper method for testing linter notifications to display received linter notifications when expecting a specific message. This way, if for example the semantic analysis changes slightly, you can compare received message and expected message to easily find what changed.

Duplicated method from [this PR](https://github.com/neo4j/upx/pull/5950) in upx where we also do playwright testing of this type.

### DIFF
--- a/packages/react-codemirror/src/e2e_tests/e2eUtils.ts
+++ b/packages/react-codemirror/src/e2e_tests/e2eUtils.ts
@@ -82,7 +82,7 @@ export class CypherEditorPage {
     await expect(this.page.locator('.cm-tooltip-hover').last()).toBeVisible({
       timeout: 10000,
     });
-    await expect(this.page.getByText(expectedMsg)).toBeVisible();
+    await this.checkHoverMessage(expectedMsg);
     /* Return the mouse to the beginning of the query and
        This is because if for example we have an overlay with a 
        first interaction that covers the element we want to perform
@@ -93,5 +93,15 @@ export class CypherEditorPage {
     await expect(
       this.page.locator('.cm-tooltip-hover').last(),
     ).not.toBeVisible();
+  }
+
+  private async checkHoverMessage(expectedMsg: string) {
+    const tooltips = await this.page.locator('.cm-tooltip-hover').all();
+
+    const tooltipTexts = await Promise.all(
+      tooltips.map((t) => t.textContent()),
+    );
+
+    expect(tooltipTexts).toContain(expectedMsg);
   }
 }


packages/react-codemirror/src/e2e_tests/e2eUtils.ts
```diff
@@ -94,4 +94,27 @@ export class CypherEditorPage {
...
+    );
+    let hasExpectedMessage = false;
+
+    tooltipTexts.forEach((tooltip) => {
+      if (tooltip.includes(expectedMsg)) {
+        hasExpectedMessage = true;
+      }
+    });
+    if (!hasExpectedMessage) {
+      throw new Error(
```
anderson4j, line 112: With a check like this, we lose the nice coloring a native playwright matcher like `toContain()` would have, but since we want "to contain string with substring x" I think we have to have a custom check like this.
Message gets formatted like this:
Screenshot 2025-05-13 112253

packages/react-codemirror/src/e2e_tests/syntaxValidation.spec.tsx
```diff
@@ -182,6 +182,22 @@ test('Validation errors are correctly overlapped', async ({ page, mount }) => {
   );
 });
 
+test('Syntax highlighting works as expected with multiple separate linting messages', async ( {
+  page,
+  mount
+}) => {
+  const editorPage = new CypherEditorPage(page);
+  const query = `MATCH (n)--(m) CALL (n) {RETURN id(n) AS b} RETURN apoc.create.uuid(), a`;
```
anderson4j, line 190: I noticed we didnt have a test where we should get warnings on multiple locations (which should yield separate hover elements in the browser) so I added this one for that

packages/react-codemirror/src/e2e_tests/e2eUtils.ts
ncordon: It'd be nicer to read with some tabs:

```
Error: 
Expected linting to contain:
   Function idk is deprecated
but got:
   Function id is deprecated
```

ncordon (approved):

packages/react-codemirror/src/e2e_tests/syntaxValidation.spec.tsx
```diff
@@ -182,6 +182,24 @@ test('Validation errors are correctly overlapped', async ({ page, mount }) => {
...
+
+  await mount(<CypherEditor value={query} schema={testData.mockSchema} />);
+  await expect(
+    editorPage.page.locator('.cm-deprecated-element').last(),
+  ).toBeVisible({ timeout: 10000 });
+  await editorPage.checkWarningMessage('id', 'Function id is deprecated.');
+  await editorPage.checkWarningMessage('id', `The query used a deprecated function. ('id' has been replaced by 'elementId or an application-generated id')`);
```
anderson4j, line 197: Hovering on "id" will yield 2 warnings (1 only on the function itself and one on the entire call subquery). They become separate elements in the "checkHoverMessage" check and we get a test fail like this when a message we check for is off.
image

OskarDamkjaer (approved):

PR #506, opened 2025-05-19
adding a run button for cypher files to editor title

nnecla:
This pr is to add a run button to the top right corner of the editor when the open file has a `.cypher` extension.

Screenshot 2025-05-19 at 16 45 58

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -149,7 +149,7 @@
       "editor/title": [
         {
           "command": "neo4j.runCypher",
-          "when": "resourceExtname == .cypher",
+          "when": "resourceLangId == cypher",
           "group": "navigation"
         }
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/package.json
```diff
@@ -145,6 +146,13 @@
           "group": "z_commands"
         }
       ],
+      "editor/title": [
+        {
+          "command": "neo4j.runCypher",
+          "when": "resourceExtname == .cypher",
```
ncordon, line 152: It'd be better to use `"editorLangId == cypher"` here. The reason is that there are more extensions a cypher file can have (`cql`, `cy`, `cyp`) and we could even have an untitled file where we pick the language to be Cypher manually

ncordon (approved):

PR #508, opened 2025-05-21
Fixes cypher 25 check for semver servers

anderson4j:
New servers produce the following versions:
    
Screenshot 2025-05-21 160521

Since `semver.coerce(...)` expects a string without zero-padding this breaks the coercion in the check for whether the server supports Cypher 25.

This pr strips the zero padding with a regex - here's an image of what it would capture (and replace with a single period)
image

### DIFF
--- a/packages/language-support/src/version.ts
+++ b/packages/language-support/src/version.ts
@@ -1,17 +1,21 @@
 import semver from 'semver';
 
 export function cypher25Supported(serverVersion: string) {
-  const minSupportedVersion = '5.27.0-2025040';
+  const minSupportedVersion = '2025.8.0'; //TODO: set to correct value when cypher 25 is actually released
   const minSupported = semver.coerce(minSupportedVersion, {
     includePrerelease: false,
   });
-  const current = semver.coerce(serverVersion, { includePrerelease: false });
+
+  const cleanedServerVersion = serverVersion.replace(/"(\.0+)(?=\d)"/g, '.');
+  const current = semver.coerce(cleanedServerVersion, {
+    includePrerelease: false,
+  });
 
   if (minSupported && current) {
     const comparison = semver.compare(minSupported, current);
     if (comparison === 0) {
       const minPrelease = semver.prerelease(minSupportedVersion)?.at(0);
-      const currentPrelease = semver.prerelease(serverVersion)?.at(0);
+      const currentPrelease = semver.prerelease(cleanedServerVersion)?.at(0);
 
       return (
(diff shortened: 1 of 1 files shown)


packages/language-support/src/version.ts
```diff
@@ -1,17 +1,21 @@
 import semver from 'semver';
 
 export function cypher25Supported(serverVersion: string) {
-  const minSupportedVersion = '5.27.0-2025040';
+  const minSupportedVersion = '2025.8.0'; //TODO: set to correct value when cypher 25 is actually released
```
anderson4j, line 4: The old value doesnt seem right anymore. Im not sure if the august release will be the one that actually has Cypher 25, but it is at least not accepted yet on a dev-server. Figure we can leave it this way and change to the correct value on release?

packages/query-tools/src/queries/version.ts
```diff
@@ -1,23 +1,25 @@
...
-/**
- * Get dbms version
- */
-export function getVersion(): ExecuteQueryArgs<{
-  serverVersion: string | undefined;
+export function getCypherVersions(): ExecuteQueryArgs<{
+  languageVersions: string[] | undefined;
 }> {
-  const query = 'CALL dbms.components() YIELD versions';
+  const query = 'CALL dbms.components() YIELD name, versions';
```
anderson4j, line 7: originally I wanted this query to end with :  WHERE name = "Cypher" RETURN versions
but this fails on system db. 

Instead I opted to sort out the "Cypher" row and "versions" column in the resulttransformer.

packages/query-tools/src/queries/version.ts
```diff
@@ -1,23 +1,25 @@
 import { resultTransformers } from 'neo4j-driver';
 import { ExecuteQueryArgs } from '../types/sdkTypes';
 
-/**
- * Get dbms version
- */
-export function getVersion(): ExecuteQueryArgs<{
```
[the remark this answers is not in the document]
ncordon: If you are going to reintroduce the method, please don't delete it

packages/query-tools/src/queries/version.ts
```diff
@@ -1,23 +1,25 @@
...
-  serverVersion: string | undefined;
+export function getCypherVersions(): ExecuteQueryArgs<{
+  cypherVersions: string[] | undefined;
```
ncordon, line 5: We should call this `serverCypherVersions` so it's clear what it means

ncordon (approved):

PR #509, opened 2025-05-26
Fixes formatting e2e test for windows

anderson4j:
The formatting e2e test was failing on Windows because the line endings differ there. This PR adds a setting so the tests files run from `vscode-extension/tests/fixtures` default to LF line endings.

Before:
image

After:
image


packages/vscode-extension/tests/fixtures/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "files.eol": "\n"
```
anderson4j, line 2: I tried adding this line to the root settings.json instead, but it seems the window runTests opens doesn't use that one

PR #512, opened 2025-05-26
Changes formatQuery interface (breaking change), adds new maxColumn option

tomasnyberg:
### Up for debate
There are a few changes I would be happy to discuss a bit more before introducing this PR, namely:
- What should happen if the formatter gets passed a very low (or even negative) number as `maxColumn`?
  - I considered throwing in these cases, but figured that might be annoying if e.g. an editor setting it dynamically based on its width gets initialized with weird values, crashing the page. So in the PR right now it just always breaks lines
- What is the correct interface for `formatQuery`?? I've introduced a new interface which I think is reasonable, but it requires careful consideration since obviously everyone who uses the formatter has to interact with it.
- What kind of patch should this be? I made it a minor since it changes an interface which felt most appropriate.
### Background
Recently, I added a formatting button to dashboards in UPX:

https://github.com/neo4j/upx/pull/5800

But, one problem with the formatter there is that their editor is 68 columns wide, while we always assume 80 in the formatter. Generally, we avoid any configuration in the formatter, but I think an exception should be made in this case so that it's possible to configure the maximum column width for the formatter. In fact, Black and Prettier both support such a setting, despite otherwise being very anti-configuration, which to me indicates that it's fine in this case in order to be pragmatic. From what I can tell, there are multiple Neo4j tools with an editor where the column width isn't necessarily 80, so I think there's a clear need for it.

### Description
- Adds a maxColumn option to the formatter which it will adhere to
- Changes the interface for `formatQuery` (!!!) to this:
```
export function formatQuery(
  query: string,
  formattingOptions?: FormattingOptions,
): FormattingResult {
```
Rather than the previous:
```
export function formatQuery(query: string): string;
export function formatQuery(
  query: string,
  cursorPosition: number,
): FormattingResultWithCursor;
export function formatQuery(
  query: string,
  cursorPosition?: number,
): string | FormattingResultWithCursor {
```
This is obviously a breaking change, and anyone consuming the formatter will need to adapt to it. I played around with some diffferent interfaces, and this is what I think is most reasonable. It allows us to add more configurability in the future (hopefully not too much) and also add things to the result if necessary. I think that's preferable over e.g. adding a new argument, as that might just mean postponing the problem for later, when we need to add further things. 

### Testing
- Adds ~5 new tests that use non-default column widths

### DIFF
--- a/.changeset/true-olives-stay.md
+++ b/.changeset/true-olives-stay.md
@@ -1,0 +1,5 @@
+---
+'@neo4j-cypher/language-support': minor
+---
+
+New interface for formatQuery, support maxColumn option
--- a/packages/language-support/src/formatting/formatting.ts
+++ b/packages/language-support/src/formatting/formatting.ts
@@ -2279,7 +2279,7 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
   };
 }
 
-interface FormattingResultWithCursor {
+interface FormattingResult {
   formattedString: string;
   newCursorPos?: number;
 }
@@ -2294,11 +2294,11 @@ export function formatQuery(query: string): string;
 export function formatQuery(
   query: string,
   formattingOptions?: FormattingOptions,
-): FormattingResultWithCursor;
+): FormattingResult;
 export function formatQuery(
   query: string,
   formattingOptions?: FormattingOptions,
-): string | FormattingResultWithCursor {
+): string | FormattingResult {
   const { tree, tokens, unParseable, firstUnParseableToken } =
     getParseTreeAndTokens(query);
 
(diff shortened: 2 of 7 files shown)


simonthuresson:
> "What should happen if the formatter gets passed a very low (or even negative) number as maxColumn?"

My initial thought is that a too low or even negative value is not really a concern, the formatter will just attempt to break as much as it can in those cases, and if that is the parameters passed, so be it?

>"What is the correct interface"

I think the one you introduced looks reasonable. An interface like that is easy to extend later, even though we "should" minimize configuration :)

Otherwise LGTM. I like the shift from function overload to using interface.

OskarDamkjaer (approved):

PR #517, opened 2025-05-28
Add graph extraction helpers to make it easier to use NVL

OskarDamkjaer:


packages/query-tools/src/cypher-execution/extract-unique-nodes-and-relationships.ts
```diff
@@ -0,0 +1,130 @@
...
+ */
+export const extractUniqueNodesAndRels = (
+  records: Record[],
+  {
+    nodeLimit,
+    keepDanglingRels = false,
+  }: { nodeLimit?: number; keepDanglingRels?: boolean } = {},
+): DeduplicatedBasicNodesAndRels => {
+  let limitHit = false;
+  if (records.length === 0) {
```
anderson4j, line 49: This looks like it isnt needed. Do you think it makes the code clearer/gives a worthwhile optimization, or what do you think about instead having a test like:

```
test('should not break on empty bolt records', () => {
    const { nodes, relationships, limitHit } = extractUniqueNodesAndRels([]);

    if (
      nodes === undefined ||
      relationships === undefined || 
      limitHit === undefined
    ) {
      throw new Error('Invalid return values on empty array of paths');
    }

    expect(nodes.length).toBe(0);
    expect(relationships.length).toBe(0);
    expect(limitHit).toBe(false);
  });
  ```

Comment (OskarDamkjaer) [reply packages/query-tools/src/cypher-execution/extract-unique-nodes-and-relationships.ts:49]: I added a test similar to this now, thanks :+1:

File: packages/query-tools/src/cypher-execution/extract-unique-nodes-and-relationships.ts [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+} from 'neo4j-driver';
+import { isNode, isPath, isRelationship } from 'neo4j-driver';
+import { CypherProperty } from '../data-types/cypher-data-types';
+
+export type Properties = RecordShape<string, CypherProperty>;
+
+/**
+ * Result type containing deduplicated nodes and relationships extracted from Neo4j records.
+ */
+export type DeduplicatedBasicNodesAndRels = {
```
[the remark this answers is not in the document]
OskarDamkjaer: Orginally we named the nodes `BasicNode` to avoid clashing with the web browser api [Node](https://developer.mozilla.org/en-US/docs/Web/API/Node) and to distinguish from the Nodes used in NVL. Here we don't have that problem, so I'll drop the `basic` part for now

PR #520, opened 2025-06-03
Sets up the logic to check/switch lint workers

anderson4j:
Switches so the vscode extension downloads the appropriate linter version when we connect to an older neo4j server. This way, when we use older syntax that is no longer valid on newer servers - ex. call subqueries without a variable scope clause - we won't get linting for it when connected to an older server.

Ex. connected to 5.23
image
but connected to Aura
image


anderson4j:
closing since old commits made it messy

PR #527, opened 2025-06-16
Simplified download lint worker

anderson4j:
Motivation:
After a discussion with Oskar, we decided to go for a simplified first approach of publishing the lint-worker package. The other approach was pretty much done, but the versioning is still a bit up in the air, so we thought it would be good to get some eyes on it before publishing a bunch of versions to npm that might later need fixing. The idea is to just publish a single lint-worker, get some feedback, and then publish more linters after (restoring the logic to handle this). 

5.20 was chosen as the cutoff since this was the upcoming version on the server the last time the artifacts were introduced before introducing the 5.23 deprecations that we got an issue about on old servers.

**PR Content:**
With this version you would:
Download and use the 5.20 lint-worker if your server version is 5.20.x or lower -> get semantic analysis/grammar linting using 5.20 artifacts. 

Use standard linting if your server version is above 5.20.x.

Some examples:
5.20.0 server -
image

image

5.17.0 (should be same as 5.20.0, so ill take only one example)
image

5.27 (aura)
image

image


packages/lint-worker/src/helpers.ts
```diff
@@ -0,0 +1,50 @@
+import { compareMajorMinorVersions } from './version';
+import {
+  DbSchema as DbSchema2,
+  Neo4jFunction,
+  Neo4jProcedure,
+} from '@neo4j-cypher/language-support';
+import { Neo4jSchemaPoller } from '@neo4j-cypher/query-tools';
+import { DbSchema as DbSchema1 } from 'languageSupport-next.13';
+
+export function convertDbSchema(
```
anderson4j, line 10: In review of the previous PR it was proposed to move this to the language-server package. I'd like some motivation for why - to me it makes sense to keep here since it's only used when we are switching lint-worker, but it wouldnt be hard to move

anderson4j:
Oh, one thing. I call relintAllDocuments() when switching lintworker, but it seems flaky to me. Sometimes it requires a change in the document to see the new(/old) linting take effect. I'll have another look to see if I can figure out how to fix it, but it could be worth having an extra close eye on this part of the code. 

PS, ive updated [the drive](https://drive.google.com/drive/folders/1u2do395A-wTXmJ75fLV1JCIl1QiUK9Vv?usp=sharing) containing my verdaccio files to contain the updated package.json and 0.0.0 release needed to try it out locally.

pnpm-workspace.yaml
```diff
@@ -1,6 +1,6 @@
 packages:
-  - 'packages/*'
-  - 'vendor/*'
-
+  - packages/*
+  - vendor/*
 catalog:
-  'neo4j-driver': '5.12.0'
+  neo4j-driver: 5.12.0
+registry: http://localhost:4873/
```
anderson4j, line 6: Note to self in this PR too: Remove this and alter url in connectionService before merge

packages/lint-worker/src/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+  originalSchema: DbSchema2,
+  neo4j: Neo4jSchemaPoller,
+): DbSchema2 | DbSchema1 {
```
ncordon, line 13: I'd prefer this to be named `DbSchemaV2` and `DbSchemaV1` as I asked in the other PR so it's clear it's two different versions of the schema

packages/lint-worker/src/helpers.ts
ncordon: Because you are just passing it to the language server anyway and that's the only place where it's used. Plus you had to configure the package to expose all of these things, so it added clutter to the package.json.

You can leave it here, but it doesn't seem a good structure to me.

ncordon (approved):

PR #530, opened 2025-06-23
removing all references to showBottomPanel

nnecla:
Ups sorry! Second pr of the [clean-up](https://github.com/neo4j/cypher-language-support/pull/528/files). Removing all references to `showBottomPanel`(hopefully) so that bottom panel shows up without the feature flag.


anderson4j (approved):
Fixes the bug I had with not being able to run queries at all, and shows the panel tab from the start now - nice!

PR #531, opened 2025-06-24
Adds symbol table to the language support

ncordon:


packages/lint-worker/src/lintWorker.ts
```diff
@@ -12,7 +13,7 @@ function lintCypherQuery(
   query: string,
   dbSchema,
   featureFlags: { consoleCommands?: boolean; cypher25?: boolean } = {},
-): SyntaxDiagnostic[] {
+): { diagnostics: SyntaxDiagnostic[]; symbolTables?: SymbolTable[] } {
```
ncordon, line 16: The `symbolTables` is optional here thinking we might not set it at all for old linters

packages/vscode-extension/tests/specs/api/autoCompletion.spec.ts
```diff
@@ -188,6 +233,35 @@ suite('Auto completion spec', () => {
     });
   });
 
+  test('Does not complete properties for non node variables', async () => {
```
anderson4j, line 236: The test content can stay the same, but like I mentioned above, rels can have properties too, so I figure this name is more true to what we want
```suggestion
  test('Does not complete properties for non node/relationship variables', async () => {
```

anderson4j (changes requested):
Nice to see a use-case for the table already in this PR. Did find some small issues, most notably the old lintWorker breaking. Suggested a workaround to fix it

packages/lint-worker/src/lintWorker.ts
ncordon: What you propose is not really feasible because if you set the symbol table inside the linting code, that's executed in a different thread normally, so it won't get exposed outside of there to the completions.

> Yeah I think it would be very messy to get 'old semantic analysis + semantic analysis with symbol table' all in one.

I don't see the problem with this: for old language supports, you just set the `diagnostics` inside the dictionary, for new ones you set both the diagnostics and the symbol table

packages/language-support/src/syntaxValidation/semanticAnalysisWrapper.ts
```diff
@@ -54,6 +58,19 @@ function copySettingSeverity(
   }));
 }
 
+function copySymbolTable(symbolTable: SymbolTable): SymbolTable {
+  return symbolTable.map(
+    ({ variable, definitionPosition, types, references }) => {
+      return {
+        variable,
+        definitionPosition,
+        types: types,
```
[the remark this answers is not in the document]
ncordon: I've added the `Array.from(types)` but it wasn't really needed. It's in the references because what comes from the semantic analysis is an `Int32Array` not an Array.

packages/lint-worker/src/lintWorker.ts
anderson4j: yeah, setting old lintworkers to
`return {diagnostics: _lintCypherQuery(query, dbSchema as DbSchema), symbolTables: []};`
should work fine, not sure why I thought it wasnt fixable...

Anyway, maybe we re-release it and try that it works after lunch, and then merge this PR so other PRs work with the new linter?

PR #535, opened 2025-07-02
Adds e2e test for neo4j version specific linter

ncordon:
When we are connected to a specific neo4j version, we should download (unless it's already available locally) and use the linter specific for that database.

This PR also cleans up the `cypher25` feature flag, which should be enabled by default now


packages/vscode-extension/package.json
```diff
@@ -436,6 +436,16 @@
           ],
           "default": true,
           "description": "Enable linting errors for Cypher queries"
+        },
+        "neo4j.features.useVersionedLinters": {
```
ncordon, line 440: I needed to add this to the configuration because otherwise we cannot enable it from the tests

PR #540, opened 2025-07-15
Updates file missing for release to npm to work

ncordon:
## Why

Because the job to release to npm is broken:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,6 +23,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-and-build
+      - name: Check the TextMate grammar is up to date
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -23,6 +23,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-and-build
+      - name: Check the TextMate grammar is up to date
+        run: |
+          (git diff-index --name-status HEAD | grep "cypher.json" && (echo "Run pnpm build and commit an up to date cypher.json file please" && exit 1) || echo "\n\nTextMate grammar is up to date"
```
ncordon, line 28: Makes sure the TextMate grammar is kept up to date in the CI so this situation does not happen again.

OskarDamkjaer (approved):

PR #542, opened 2025-07-16
Enables linter switching feature

ncordon:
After releasing all linters for neo4j from version 5.23 to version 2025.07, this PR enables the switching between them.

## Some details
The scheme used to store the linters in disk is `${neo4jVersion}-lintWorker-${npmRelease}.cjs`, where only the first two digits are used to match the neo4j version. That means if we use `2025.05.0` and `2025.05.1` they should both use the same linter `2025.05-lintWorker-{{most up to date version here}}`.

And that means when releasing a new linter, only the major and minor version in the neo4j release should be used for the npm tag:

```
pnpm publish --tag neo4j-2025.02 --no-git-checks
```

### DIFF
--- a/packages/vscode-extension/README.md
+++ b/packages/vscode-extension/README.md
@@ -123,7 +123,7 @@ Once you've written your desired query (for example `CREATE (n)-[r:Rel]->(m) RET
 
 ![demo-execution](https://github.com/neo4j/cypher-language-support/blob/main/packages/vscode-extension/resources/images/demo-execution.png?raw=true)
 
-## Linting experience
+## Version tailored linting
 
 Our aim is to provide an experience that suits the different neo4j versions you could be connected to. When connecting to a database, a linter that matches that version of the database will be automatically downloaded.
 


packages/vscode-extension/src/linterSwitching.ts
```diff
@@ -179,29 +173,34 @@ export async function switchToLinter(
...
+        if (success) {
+          await switchWorkerOnLanguageServer(expectedFileName, storageUri);
+        } else {
+          await switchToLocalLinter(linterVersion);
+        }
       }
     }
+  } catch (e) {
+    // In case of error use default linter (i.e. the one included with the language server)
+    await switchWorkerOnLanguageServer();
```
ncordon, line 203: The main difference here is adding the try catch block to use the default language server (the one included in the plugin) in case something wrong happens during the process

packages/language-server/src/server.ts
```diff
@@ -43,7 +43,6 @@ async function lintSingleDocument(document: TextDocument): Promise<void> {
         });
       },
       neo4jSchemaPoller,
-      settings?.features?.useVersionedLinters,
```
ncordon, line 46: This PR cleans up this flag

packages/lint-worker/src/helpers.ts
```diff
@@ -7,28 +7,26 @@ import {
...
   }
-  if (
-    originalSchema.functions['CYPHER 5'] &&
-    originalSchema.procedures['CYPHER 5']
-  ) {
-    oldFunctions = originalSchema.functions['CYPHER 5'];
-    oldProcedures = originalSchema.procedures['CYPHER 5'];
-  }
 
-  if (compareMajorMinorVersions(linterVersion, oldLinter) <= 0) {
```
ncordon, line 31: This was not correct. The check needs to be any version lower than 2025.01, which is when we changed the schema.

packages/vscode-extension/package.json
```diff
@@ -478,9 +467,9 @@
     "dev": "tsc -b && pnpm gen-textmate && concurrently 'node --experimental-strip-types esbuild-extension.mts' 'pnpm bundle-webview-controllers --watch' ",
     "clean": "rm -rf {dist,tsconfig.tsbuildinfo}",
     "package": "pnpm vsce package --pre-release --no-dependencies",
-    "test:e2e": "DEBUG_VSCODE_TESTS=false && pnpm build:dev && pnpm test:apiAndUnit && pnpm test:webviews",
-    "test:apiAndUnit": "DEBUG_VSCODE_TESTS=false && pnpm build:dev && rm -rf .vscode-test/user-data && node ./dist/tests/runApiAndUnitTests.js",
-    "test:webviews": "DEBUG_VSCODE_TESTS=false && pnpm build:dev && wdio run ./dist/tests/runWebviewTests.js"
+    "test:e2e": "pnpm build:dev && cross-env DEBUG_VSCODE_TESTS=false pnpm test:apiAndUnit && cross-env DEBUG_VSCODE_TESTS=false pnpm test:webviews",
+    "test:apiAndUnit": "pnpm build:dev && rm -rf .vscode-test/user-data && cross-env DEBUG_VSCODE_TESTS=false node ./dist/tests/runApiAndUnitTests.js",
+    "test:webviews": "pnpm build:dev && cross-env DEBUG_VSCODE_TESTS=false wdio run ./dist/tests/runWebviewTests.js"
```
ncordon, line 472: There was a bug here (which did not have any implications until now) and the `DEBUG_VSCODE_TESTS` variable was not passed to the test job because you either do:

```bash
export VARIABLE=value && job
```

or you have to prepend setting the variable _just before_ every job.

packages/vscode-extension/tests/specs/api/versionSpecificLinting.spec.ts
```diff
@@ -48,7 +57,7 @@ suite('Neo4j version specific linting spec', () => {
             new vscode.Position(5, 7),
             new vscode.Position(5, 12),
           ),
-          "The query used a deprecated function. ('id' has been replaced by 'elementId or consider using an application-generated id')",
```
ncordon, line 51: This change is needed because now that the linter is actively downloaded, the message included in the linter for the 2025.05 version used in the tests was slightly different to the one the default linter (for the current neo4j) includes

packages/vscode-extension/README.md
```diff
@@ -123,11 +123,21 @@ Once you've written your desired query (for example `CREATE (n)-[r:Rel]->(m) RET
 
 ![demo-execution](https://github.com/neo4j/cypher-language-support/blob/main/packages/vscode-extension/resources/images/demo-execution.png?raw=true)
 
-## Upcoming features
+## Linting experience
```
OskarDamkjaer, line 126: Nice update! How about we call it:? 
```suggestion
## Version tailored linting
```

OskarDamkjaer:
>And that means when releasing a new linter, only the major and minor version in the neo4j release should be used for the npm tag
Is this something we're planning to set up later? If so, do we have a card for it?

PR #544, opened 2025-07-28
Automatic lint-worker release

anderson4j:
Normally we publish our npm packages automatically with the changeset framework, with updates coming from the changesets we generate for PRs. We ran into an issue with the lintworker package since the releases we want to use from there need to have a tag, which should change to reflect the neo4j server the release is for. For existing releases we therefor created them manually.

Since we know what server version we take the grammar/semantic analysis artifacts from however, we should be able to store this version, and use it to tag our lint-worker package releases. That is what this PR does, in 2 steps.

1. Store the current neo4j server version in a file when updating the artifacts. 

To test this out I ran the `Monthly Artifact Update` workflow in a private repo. The final line shows the full print of the file. As you can see the file content becomes the full tag `neo4j-2025.08` which is what we also have in the demo video in step 2 below.
image

2. Update the release script to use this tag for the lint-worker package

I tried this out by changing the release script to use my verdaccio proxy, and it seems to work as expected - see demo below
[See video](https://drive.google.com/file/d/1jrIDLe2K6w0Eb0Xi5JK63AA8tiWFjzUG/view?usp=sharing)

After some issues with the commit, I now have the whole modified `fetch-dev-artifacts` workflow running successfuly in my private copy of the repo. Starting off with a file containing a dummy version, we get the correct update
image


package.json
```diff
@@ -43,7 +43,7 @@
     "eslint": "^8.32.0",
     "eslint-config-prettier": "^8.6.0",
     "husky": "^8.0.0",
-    "lint-staged": "^15.5.0",
+    "lint-staged": "^16.1.2",
```
anderson4j, line 46: the old version had a bug that caused us to fail when overwriting a server-version-tag.txt from `neo4j-5.26` to `neo4j-2025.08`

ncordon (approved):

PR #549, opened 2025-08-04
Fixes flaky param test

anderson4j:
Sometimes the test failed with:
image

After calling a vscode command to switch connection.

The method used in the connection spec does not run into this issue, and switching for this test removed the flakiness


anderson4j:
If you click the check mark of the CI, you can see that I re-ran the webview tests 10 times, and it passed each time

OskarDamkjaer (approved):

PR #563, opened 2025-08-21
Adds polling of graph schema

anderson4j:
For schema aware completions we already get the symbol table for the current query, but also need the schema of the underlying graph we will run the query against. We decided to use `db.schema.visualization` for this, which lets us see which nodes (which labels) are connected via which rels (which rel type). See the image below
image

This PR introduces calls to `db.schema.visualization()` in the polling and saves the results in the dbSchema as a set of rels (containing to/from nodes as their label). Note that when a node has multiple labels, it shows up as multiple separate nodes in the schema visualization (for example DRIVER and PERSON are on a single node in the graph, but illustrated as 2 nodes above), so each node in the result only has one label.

Here i've printed out the new schema property as it it looks when in the completion part of the code.
Screenshot 2025-08-21 152839


packages/query-tools/src/queries/graphSchema.ts
```diff
@@ -0,0 +1,116 @@
...
+    nodes: {
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          labels: {
+            type: 'array',
+            items: { type: 'string' },
+          },
+          elementId: { type: 'string' },
```
OskarDamkjaer, line 32: can we verify this schema also works for data from a `4.4` database in the next commits? Since the `4 LTS` is still in support (and those usually get extended)

packages/query-tools/src/queries/graphSchema.ts
```diff
@@ -0,0 +1,116 @@
...
+function extractRelationshipsWithNamedNodes(
+  graphSchemas: GraphSchema[],
+): Set<{ from: string; to: string; relType: string }> {
+  const items = new Set<{ from: string; to: string; relType: string }>();
+
+  if (graphSchemas.length !== 1) {
+    return items;
+  }
+  const graphSchema = graphSchemas[0];
+  // const nodes = record.get('nodes').map(node => { return { labels: node.labels, id: node.elementId }});
```
OskarDamkjaer, line 101:
```suggestion
```

packages/query-tools/src/queries/graphSchema.ts
anderson4j: Ran against a 4.4.42 server and still got elementID/polled schema, so good!

OskarDamkjaer: Perfect! must be some early 4.x series or 3.5 that doesn't have elementId then!

packages/query-tools/src/queries/graphSchema.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  if (graphSchemas.length !== 1) {
+    return items;
+  }
+  const graphSchema = graphSchemas[0];
+  const nodes: Record<string, string> = {};
+
+  for (const node of graphSchema.nodes) {
+    //Query should only return 1 label
+    nodes[node.elementId] = node.labels[0];
```
OskarDamkjaer, line 101: I remembered there are nodes without a label. I don't know how / if they show up in the stats though! It would be a good thing just to quickly test

OskarDamkjaer: perf, thanks for checking!

packages/query-tools/src/tests/setupTestContainer.ts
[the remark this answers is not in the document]
OskarDamkjaer: duplication sounds like a good first step here

OskarDamkjaer (approved):

PR #565, opened 2025-08-22
adding rich clipboard copier to codemirror

nnecla:
This pr is to enable rich clipboard copier for the cypher editor. With these changes, we overwrite the default copy behaviour with a custom extension named `richClipboardCopier`.

`richClipboardCopier` copies the selection as a plain text, as well as an html. Since most of our styles are declared under classnames, within `getHTML`, we look for the styles for each classname(`getCSSStyleForClass`) and embed it to html.

You can test it out by copying some text from Cypher Editor to Google Docs, or some other rich text editor that may support.

### DIFF
--- a/packages/react-codemirror/src/richClipboardCopier.ts
+++ b/packages/react-codemirror/src/richClipboardCopier.ts
@@ -69,12 +69,12 @@ export function getHTML(view: EditorView, from: number, to: number): string {
     .forEach(replaceClassWithStyle);
 
   const editorElement = document.querySelector('.cm-editor');
-  const editorStyles = getComputedStyle(editorElement);
+  const editorStyles = editorElement ? getComputedStyle(editorElement) : null;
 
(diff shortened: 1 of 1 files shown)


packages/react-codemirror/src/richClipboardCopier.ts
```diff
@@ -0,0 +1,74 @@
+import { EditorView } from '@codemirror/view';
+
+export function getCSSStyleForClass(className: string): string {
```
OskarDamkjaer, line 3: I'm wondering, could we use [getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) here instead? This way we don't have to deal with the differences between class application vs inline styles etc

OskarDamkjaer (approved):
works great, nice research and work! I added a comment that could maybe make it easier to implement

PR #570, opened 2025-09-08
Fixes linter terminating on slow query

anderson4j:
Fixes an issue where in large and slow queries would crash the codemirror, since the worker returned by the pool for the current task got terminated by cancelling the previous task. In quick queries this would not show since the previous task would never be cancelled.

Looking at the docs for the workerpool, it seems that the worker performing the cancelled task is always terminated when no abortListener is added. I could not find a way to get the workerpool to give us a different worker, so we always got an error when the termination happened. Having an abortListener - even one that does nothing - allows the worker to be reused, and fixes our issue.

For even larger queries the workerTerminateTimeout would kick in during cancellation, terminating the active worker, so I increased this number as well.

### DIFF
--- a/packages/react-codemirror/package.json
+++ b/packages/react-codemirror/package.json
@@ -62,7 +62,7 @@
     "prismjs": "^1.29.0",
     "style-mod": "^4.1.2",
     "vscode-languageserver-types": "^3.17.3",
-    "workerpool": "^9.0.4"
+    "workerpool": "^9.3.3"
   },
   "devDependencies": {
     "@neo4j-ndl/base": "^3.2.10",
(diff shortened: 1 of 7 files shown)


packages/react-codemirror/package.json
```diff
@@ -62,7 +62,7 @@
     "prismjs": "^1.29.0",
     "style-mod": "^4.1.2",
     "vscode-languageserver-types": "^3.17.3",
-    "workerpool": "^9.0.4"
+    "workerpool": "^9.3.3"
```
anderson4j, line 65: Bumping this allowed us to reproduce the issue in our built-in react-codemirror playground

PR #575, opened 2025-09-18
Removes caret-dependencies

anderson4j:
Since there have been quite a few supply chain attacks on npm lately, the recommendation of handling npm dependencies have been updated. We should now specify the exact version we want to use, so we don't accidentaly download patch/minor updates that could contain a vulnerability/virus. 

We previously used versions like "^x.y.z" in most places, but this PR replaces those with exact versions. To avoid issues, I've used the versions that were currently in use, aka those in the pnpm-lock file


packages/vscode-extension/package.json
```diff
@@ -479,37 +479,36 @@
...
-    "axios": "^1.9.0"
+    "react": "18.3.1",
+    "react-dom": "18.3.1",
+    "vscode-languageclient": "8.1.0",
+    "tar": "7.4.3",
+    "axios": "1.9.0"
   },
   "devDependencies": {
-    "@testcontainers/neo4j": "^10.4.0",
-    "@types/glob": "^8.1.0",
```
anderson4j, line 490: I've removed this one, since it's apparently already included in the glob dependency (see https://www.npmjs.com/package/@types/glob?activeTab=versions)

PR #576, opened 2025-09-19
Fix archive-vscode-artifacts workflow

anderson4j:
It looks like there is a check in `vsce package` that ensures the `@types/vscode` version and the `engine.vscode` versions are compatible, and that this is based off the package.json.
image
Even though we already had 1.93.0 installed in the pnpm-lock file previous to my caret-fixing PR, this didnt trigger the check to fail. However, I think the idea is for the types to not be too new anyway, since this would mean the type system thinks we have newer apis than we do (so we dont error on incorrect calls), so installing 1.75.0 types would be the way to go.

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -491,7 +491,7 @@
     "@types/react": "18.3.5",
     "@types/react-dom": "18.3.0",
     "@types/sinon": "17.0.3",
-    "@types/vscode": "1.93.0",
+    "@types/vscode": "1.75.0",
     "@vscode/test-electron": "2.4.1",
     "@vscode/vsce": "3.3.2",
     "@wdio/cli": "8.41.0",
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/tests/mocks/mockExtensionContext.ts
```diff
@@ -12,14 +11,13 @@ import { InMemorySecretStorage } from './inMemorySecretStorage';
 import { getExtensionStoragePath } from '../helpers';
 
 export class MockExtensionContext implements ExtensionContext {
-  languageModelAccessInformation: LanguageModelAccessInformation;
```
anderson4j, line 15: These fields werent used in the test that we have MockExtensionContext in, so removing one and modifying the other (to fit the old ExtensionContext implementation) didnt need any changes there

Issue #554, opened 2025-08-05
Focus the Query Results panel after running a query in VS Code


nachocab:
If I'm in the Terminal and I run the "neo4j.runCypher" command, nothing happens.

It would be good if the Query Results panel was automatically focused:

<img width="139" height="71" alt="Image" src="https://github.com/user-attachments/assets/e13d7c94-d2fa-4b6f-a430-32aa6487b98d" />

anderson4j:
Thanks for yet another useful comment! Merged in [this](https://github.com/neo4j/cypher-language-support/pull/587) PR

anderson4j:
Yeah we're not too many working on it and sometimes it get's tricky to pick what to prioritise, but if you have other issues we'd be happy to hear them!

Issue #553, opened 2025-08-05
Add a "when clause" to the keybindings in VS Code


nachocab:
To avoid conflict with other VS Code extensions, a [when clause](https://code.visualstudio.com/api/references/when-clause-contexts) should probably be added to the default keybinding. 

Something like this:

```json
{
    "key": "ctrl+cmd+space",
    "command": "neo4j.runCypher",
    "when": "editorTextFocus && editorLangId == 'cypher'"
  }
```

Issue #552, opened 2025-08-05
Execute a single statement by default


nachocab:
Currently, the `Neo4j: Run Cypher statements` command runs all the statements in a file. To run a single statement, it must be selected before running the command. 

It would be good if, by default, the `neo4j.runCypher` command  only ran the statement where the cursor is currently placed and there was another command to run all the statements in a file.

This is how the [Snowflake extension](https://marketplace.visualstudio.com/items?itemName=snowflake.snowflake-vsc&ssr=false#user-content-query-execution) and the [Database Client](https://github.com/cweijan/vscode-database-client?tab=readme-ov-file#execute-sql-query) extensions work, for example.

anderson4j:
Thanks for the suggestion! This change is merged now as part of [this](https://github.com/neo4j/cypher-language-support/pull/587) PR

PR #587, opened 2025-10-16
Adds "run single cypher" command

anderson4j:
Adds command to run only the statement where the caret is, as suggested here https://github.com/neo4j/cypher-language-support/issues/552.

Trying to use a variation of similar keys to the current run command gave me some shortcut collisions. "ctrl+space" is already the keybind for autocompleting and "alt+space" collided with windows powertoys, which I use, and could see being an issue for other windows users. In my experience the enter key is most associated with executing, so I went with "ctrl+enter" which we also use for "run" in [query](https://console-preview.neo4j.io/tools/query).

I also added the command to the right-click context menu
image
but left the "run cypher" button in the top right as is (run all)
image

Also adds "when" checks to the shortcuts (the old one too), so we don't get overlap with other extensions when there is no cypher to run. (See this issue https://github.com/neo4j/cypher-language-support/issues/553)

Around these changes I also found something that looks like a bugged call intending to solve this https://github.com/neo4j/cypher-language-support/issues/554, so I made it work

### DIFF
--- a/packages/vscode-extension/src/commandHandlers/connection.ts
+++ b/packages/vscode-extension/src/commandHandlers/connection.ts
@@ -220,7 +220,7 @@ export function getCurrentStatement(): string | undefined {
   if (editor) {
     const currentOffset = editor.document.offsetAt(editor.selection.active);
     return getStatementAtCaret(editor.document.getText(), currentOffset);
-  } else return ''; //Maybe error notification here
+  } else return '';
 }
 
 //exported for testing


packages/vscode-extension/tests/specs/unit/cypherRunner.spec.ts
```diff
@@ -42,21 +60,73 @@ MATCH (n) RETURN n`,
...
+        caret: ' MATCH (n) RETURN n; \n RETURN '.length,
+        expected: 'RETURN 50',
+      },
+    ];
+    cases.forEach((c) => {
+      const statement = getStatementAtCaret(c.query, c.caret);
+      assert.strictEqual(
+        statement,
+        c.expected,
+        "Failed on case '" + c.name + "'",
```
anderson4j: Failure message will look like:
Screenshot 2025-10-16 120905

packages/vscode-extension/tests/specs/unit/cypherRunner.spec.ts
```diff
@@ -42,21 +60,73 @@ MATCH (n) RETURN n`,
       new Position(3, 10),
       new Position(3, 19),
     );
-    // We pick the `CREATE (n:Person)` from the first line and the `RETURN n` from the last one
+    // We pick the `CREATE (n:Person)` from the first line of the document and the `RETURN n` from the last one
```
anderson4j: modified a bit since I moved document away

packages/vscode-extension/tests/specs/unit/cypherRunner.spec.ts
```diff
@@ -42,21 +60,73 @@ MATCH (n) RETURN n`,
...
-    await runCypher(async () => {});
+    await runCypher(async () => {}, getSelectedText());
 
     const arg = runMethod.firstCall.args.at(0) as string;
 
     // We check that the first line and the RETURN n from the last one have been concatenated correctly
     assert.equal(arg, 'CREATE (n:Person) RETURN n');
   });
+
+  test('Running single query should work', async () => {
```
anderson4j: This just tests a simple method and doesnt need the setup with a server and a runner that the other tests do... but it's about the cypherrunner, and definitely a unit test, so maybe we just keep it in here anyway? Didn't find a better existing folder for it

packages/vscode-extension/package.json
```diff
@@ -91,6 +91,12 @@
         "category": "Neo4j",
         "icon": "$(play)"
       },
+      {
```
nnecla: I wonder if we should add another button to the `editor/title` for single cypher run. We can also change the icon used for the run all cypher statements to something else(I've used `run-all` icon). Wdyt?

Image

packages/vscode-extension/src/commandHandlers/connection.ts
```diff
@@ -205,18 +210,39 @@ export async function runCypher(
...
 
   if (editor) {
-    const selectedText = getSelectedText(editor);
+    const currentCaret = editor.document.offsetAt(editor.selection.active);
+    return getStatementAtCaret(editor.document.getText(), currentCaret);
+  } else return ''; //Maybe error notification here
+}
 
+//exported for testing
+export function getStatementAtCaret(input: string, caret: number): string {
```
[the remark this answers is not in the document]
anderson4j: Huh, the check 
  ```
currentStatement =
    !currentStatement && statements.statementsParsing
      ? statements.statementsParsing.at(-1)
      : currentStatement;
```
was specifically to handle the case of "caret after final token" but maybe it passes a bad query to run when the final query is empty (semicolon followed by nothing like your picture). Anyway, I'll have another look, good catch!

About which statement to run: I think of the grammar as the "source of truth" here. The main entry point looks like `statementOrCommand (SEMICOLON statementOrCommand)* ...` so a `statementOrCommand` is anything up until a semicolon, then a new `statementOrCommand` is started. Following this, we would go against your taste and run the latter query in your example.

To me `;` means "query over" and in with that way of thinking it makes sense that moving the caret after the semicolon wouldnt run it.

packages/vscode-extension/package.json
anderson4j: After discussion with Greg, I've gone with a single button with the following logic
`if hasSelection -> runQuery else runSingleQuery`

PR #592, opened 2025-10-22
adding support for the help command

nnecla:
This pr is to add :help command to the grammar. This command displays the available commands in query.

### DIFF
--- a/packages/language-support/src/antlr-grammar/CypherCmdParser.g4
+++ b/packages/language-support/src/antlr-grammar/CypherCmdParser.g4
@@ -21,6 +21,7 @@ consoleCommand: COLON (
     | styleCmd
     | playCmd
     | accessModeCmd
+    | helpCmd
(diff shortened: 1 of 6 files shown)


packages/language-support/src/antlr-grammar/CypherCmdParser.g4
```diff
@@ -55,6 +56,8 @@ accessModeArgs: (readCompletionRule | writeCompletionRule);
 
 accessModeCmd: ACCESSMODE accessModeArgs?;
 
+helpCmd: HELP;
```
anderson4j: Feature sugggestion: What do you think about making it `:help consoleCmd?` and having it so when one specifies a consoleCmd one can get more info only on this command?

anderson4j (approved):
LGTM. If you take my suggestion though, it would need a small alteration

PR #594, opened 2025-10-24
Remove slice

anderson4j:
Removes "temporary hack" cutting of query at the cursor for autocompletions. This changes the results of the parsing, most notably the stopNode is now at the proper end of the query, not at the cursor. This node is used for the "current/last rule" for some completions, so we needed to find it in a new way. I added a parser-listener for this.

### DIFF
--- a/packages/language-support/src/parserWrapper.ts
+++ b/packages/language-support/src/parserWrapper.ts
@@ -184,9 +184,12 @@ export function parse(query: string): StatementOrCommandContext[] {
 
 export function createParsingResult(
   query: string,
-  consoleCommandsEnabled: boolean,
-  caretPosition?: number,
+  options: {
+    consoleCommandsEnabled: boolean;
+    caretPosition?: number;
+  },
 ): ParsingResult {
+  const { consoleCommandsEnabled, caretPosition } = options;
   const parsingScaffolding = createParsingScaffolding(query);
 
   const results: ParsedStatement[] =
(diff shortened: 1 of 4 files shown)


packages/language-support/src/autocompletion/schemaBasedCompletions.ts
```diff
@@ -127,19 +127,21 @@ export function completeRelationshipType(
 
   // limitation: not checking PathPatternNonEmptyContext
   // limitation: not handling parenthesized paths
-  const callContext = findParent(
-    parsingResult.stopNode.parentCtx,
+  const patternContext = findParent(
```
anderson4j: Changes in this file are just renaming to patternContext. Think the name callContext came from copy-pasting a place where we tried to match the "rule containing `CALL`" in procedure calls.

packages/language-support/src/parserWrapper.ts
```diff
@@ -185,8 +184,12 @@ export function parse(query: string): StatementOrCommandContext[] {
 
 export function createParsingResult(
   query: string,
-  consoleCommandsEnabled: boolean,
+  options: {
+    consoleCommandsEnabled: boolean;
+    caretPosition?: number;
+  },
```
OskarDamkjaer: nice to add the options as an object :+1:

OskarDamkjaer (approved):

PR #595, opened 2025-10-28
Schema based completions for nodes in paths

anderson4j:
Implements similar logic that we used for paths like
`MATCH (n:A)-[:` to complete rels, to also work for node completions in paths like:
`MATCH (n:A)-[r:R]->(:`

Also uses the new `lastRule` property on the parsing result to get the latest node/rel pattern before the caret in both rel/node completions to avoid taking just "last pattern in parent ctx" failing on queries like `MATCH (n:A)-[r:R]->(:^)-[r2:R2]->(:` 

Closes DEV-197

### DIFF
--- a/packages/language-support/src/tests/autocompletion/schemaBasedPatternCompletion.test.ts
+++ b/packages/language-support/src/tests/autocompletion/schemaBasedPatternCompletion.test.ts
@@ -462,6 +462,45 @@ RETURN [(p)-[:`;
     });
   });
 
+  test('Handles undirected rels for node completions', () => {
+    const query = 'MATCH (p:Pokemon)-[r:KNOWS]-(:';
+    testCompletions({
+      query,
+      dbSchema,
+      computeSymbolsInfo: true,
+      expected: [
+        { label: 'Move', kind: CompletionItemKind.TypeParameter },
+        { label: 'Pokemon', kind: CompletionItemKind.TypeParameter },
+      ],
+      excluded: [
+        { label: 'Trainer', kind: CompletionItemKind.TypeParameter },
+        { label: 'Region', kind: CompletionItemKind.TypeParameter },
+        { label: 'Type', kind: CompletionItemKind.TypeParameter },
+        { label: 'Gym', kind: CompletionItemKind.TypeParameter },
+      ],
+    });
+  });
+
+  test('Handles undirected rels for rel completions', () => {
+    const query =
+      'MATCH (p:Pokemon)-[r:CHALLENGES]-(m:Gym)-[r2:IS_IN]->(reg:Region)-[:';
+    testCompletions({
+      query,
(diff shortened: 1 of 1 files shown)


packages/language-support/src/autocompletion/schemaBasedCompletions.ts
```diff
@@ -52,45 +52,47 @@ export const allReltypeCompletions = (dbSchema: DbSchema) =>
...
+    (c) => (union = union.union(walkLabelTree(labelToConnectedLabelsMap, c))),
   );
   return union;
 }
 
+// outgoingLabelsMap is here a map that takes a "label" = Label/ RelType
+//  and returns the "labels" of rels/nodes going out/in of the node/rel in the graph schema
 function walkLabelTree(
-  relsFromLabels: Map<string, Set<string>>,
+  labelToConnectedLabelsMap: Map<string, Set<string>>,
```
anderson4j: We can use this identically if it is 
Map<label, Set<rel types>> (rel completion)
Map<rel types, Set<labels>>, (node completion)

I struggle to find a nice name that covers both. Maybe I should add the above comment into the code to help clarify things that way?

packages/language-support/src/tests/autocompletion/schemaBasedPatternCompletion.test.ts
```diff
@@ -64,6 +64,45 @@ describe('completeRelationshipType', () => {
     });
   });
 
+  test('Simple node completion pattern', () => {
+    const query = 'MATCH (t:Trainer)-[r:CATCHES]->(:';
```
OskarDamkjaer: It'd be nice to add a few more test cases (longer pattern, different relationships). Kind of like we have for the rel types?

packages/language-support/src/tests/autocompletion/completionAssertionHelpers.ts
```diff
@@ -38,6 +38,9 @@ export function testCompletions({
   manualTrigger?: boolean;
   computeSymbolsInfo?: boolean;
 }) {
+  if (query.length > offset) {
+    query = query.slice(0, offset);
+  }
```
OskarDamkjaer: is this not done in the auto completion function now?

packages/language-support/src/autocompletion/schemaBasedCompletions.ts
OskarDamkjaer: Just a thought: is there a risk anywhere in our logic around people using the same names for a Reltype or label?

OskarDamkjaer: I have some ideas on how we maybe could make this data structure better or more efficient, but I think we can wait until the feature matures to consider that refactoring

PR #667, opened 2026-04-01
adding auto-completion for :auto

nnecla:
This pr is to add auto-completion support for the `:auto` command which is used by the Query app:
- `:auto` to preview a small explanation on auto-committing transactions
- `:auto [STATEMENT]` to run the statement with forced auto-committing transactions

### DIFF
--- a/packages/language-support/src/antlr-grammar/CypherCmdParser.g4
+++ b/packages/language-support/src/antlr-grammar/CypherCmdParser.g4
@@ -22,6 +22,7 @@ consoleCommand: COLON (
     | playCmd
     | accessModeCmd
     | helpCmd
+    | autoCmd
 );
 
 paramsCmd: PARAM paramsArgs?;
@@ -57,6 +58,8 @@ accessModeArgs: (readCompletionRule | writeCompletionRule);
 accessModeCmd: ACCESSMODE accessModeArgs?;
 
 helpCmd: HELP;
+
+autoCmd: AUTO statement SEMICOLON? EOF | AUTO EOF;
 
 // These rules are needed to distinguish cypher <-> commands, for exapmle `USE` and `:use` in autocompletion
 listCompletionRule: LIST; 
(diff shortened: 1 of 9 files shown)


packages/language-support/src/antlr-grammar/CypherCmdParser.g4
```diff
@@ -59,9 +59,9 @@ accessModeCmd: ACCESSMODE accessModeArgs?;
 
 helpCmd: HELP;
 
-autoCmd: AUTO statement SEMICOLON? EOF | AUTO EOF;
+autoCmd: AUTO statement?;
```
anderson4j: autoCmd is already inside statementsOrCommands that covers the semicolon/EOF

packages/language-support/src/syntaxValidation/completionCoreErrors.ts
```diff
@@ -73,7 +72,6 @@ export function completionCoreErrormessage(
   // Exception: inside console commands (e.g. :auto) we do want to report
   // statement-level errors since there's no database error stack for those
   if (
-    !insideConsoleCommand &&
```
anderson4j: We only got the "expected a statement" error here, which is not very helpful. I figure it's better we lint with `:auto` as we normally do (so ignoring this)

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -347,7 +347,10 @@ export function lintCypherQuery(
     const statements = resolvedParsingResult.statementsParsing;
     const result = statements.map((current) => {
       const cmd = current.command;
-      if (cmd.type === 'cypher' && cmd.statement.length > 0) {
+      if (
+        (cmd.type === 'cypher' || cmd.type === 'auto') &&
```
anderson4j, line 351: Not having this before meant we didnt get regular linting on the statement part of an `:auto` command

nnecla: Makes sense!

anderson4j (approved):
Looked good to me except the linting, after my fix im happy with that too

nnecla:
Thanks @anderson4j Your updates are looking good from my side!

PR #668, opened 2026-04-01
Fix formatting on antlr error recovery

anderson4j:
Previosly we would error when ANTLR did error recovery for nodes in the middle of a query. We could already handle cases like<br>`MATCH  (n) WHERX (n:Person) RETURN n` where the parsing fails completely in the middle. The formatter then only formats up until the "unparseable" part and inserts the rest of the query as-is.

Sometimes though, like in <br>`MATCH (n:-)   RETURN n` antlr recovers, inserting an error-node on the `-` but correctly parsing the "RETURN n" part afterward. Before this fix we would throw an error in this case, because we didnt visit the error node, causing it to not be inserted to any of the formatted chunks - thus effectively being deleted. The fix we have here is to insert missed preceding error nodes (with the content as-is) when creating a chunk in the `visitTerminal` and `visitTerminalRaw` for properly parsed rules.

Claude got a bit lost while testing things, so I interrupted it and made the final steps myself, but it's an agent-driven PR.

Closes [LS-138](https://linear.app/neo4j/issue/LS-138/formatter-panic-on-template-query)<br>Fixes [LS-138](https://linear.app/neo4j/issue/LS-138/formatter-panic-on-template-query)


OskarDamkjaer:
I had an old branch migrating to `antlr4-ng` and there I needed to do some updates to the error recovery, so I wanted to see if we'd done the same thing twice! Turns out this was a separate problem, so no issue there. Apologies for the delay in checking that

packages/language-support/src/formatting/formatting.ts
```diff
@@ -548,6 +550,57 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     }
   };
 
+  private _collectErrorNodes(node: ParserRuleContext | TerminalNode): void {
+    if (node instanceof TerminalNode) {
+      // @ts-expect-error isErrorNode exists on ErrorNode at runtime but not in TerminalNode types
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
OskarDamkjaer, line 556: I guess this is the fault of antlrs types not being great?

packages/language-support/src/formatting/formatting.ts
```diff
@@ -580,7 +641,12 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
         this.previousTokenIndex + 1,
         errorTokenIndex,
       );
-      gapText = skippedTokens.map((t) => t.text).join('');
+      // Normalize \r\n to \n so the layout engine (which uses \n) works
+      // consistently across platforms.
+      gapText = skippedTokens
+        .map((t) => t.text)
+        .join('')
+        .replace(/\r\n/g, '\n');
```
OskarDamkjaer, line 649: This is a bit strange to me, is this so that the tests work on your machine? I would have assumed we'd just keep what was there before and not "normalize" across plattforms 🤔

OskarDamkjaer (approved):

PR #671, opened 2026-04-07
Manual update to 2026.04 artifacts

anderson4j:
Update grammar and semantic analysis to 2026.04 

Fix some tests that changed, and formatting that needed updating

Closes LS-139


.lintstagedrc.json
```diff
@@ -7,6 +7,7 @@
   ],
   "vendor/antlr4-c3/index.ts": ["bash check-imports.sh"],
   "*.js": ["oxlint --fix --type-aware --max-warnings 0"],
+  "packages/vscode-extension/**/cypher.json": [],
```
anderson4j, line 10: had a clash with oxfmtrc for cypher.json. This file called oxfmt on all .json files, but oxfmtrc (rightly, since it's generated) removed cypher.json, causing an "Expected at least one target file"-error.

packages/lint-worker/server-version-tag.txt
```diff
@@ -1 +1 @@
-neo4j-2026.03
+neo4j-2026.04
```
anderson4j, line 1: not forgetting it this time!

packages/language-support/src/tests/syntaxValidation/proceduresValidation.test.ts
```diff
@@ -745,25 +745,6 @@ meaning that it expects at least 3 arguments of types NODE, STRING, ANY
         dbSchema: testData.mockSchema,
       }),
     ).toEqual([
-      {
-        severity: 1,
-        message:
-          'Procedure call inside a query does not support naming results implicitly (name explicitly using `YIELD` instead)',
```
anderson4j, line 751: Looks like semantic analysis got smart enough to realise the procedure doesnt yield any columns at all, so this error doesnt make sense. Added another test below with a procedure that does have return columns

packages/language-support/src/antlr-grammar/Cypher25Parser.g4
```diff
@@ -265,6 +266,10 @@ unwindClause
    : UNWIND expression AS variable
    ;
 
+forListClause
+   : FOR variable IN expression
```
anderson4j, line 270: Think we should add formatter handling for this new clause. Realised that we didnt consider this for a lot of new rules in the past, so thinking maybe I skip it for now and have a look at it all at once later? (this also lets us get in the updated artifacts to upx quicker - before the browser bundling on thursday)

anderson4j: Made a linear card about it here https://linear.app/neo4j/issue/DEV-534/look-into-formatting-of-new-rules-from-grammar-updates

packages/language-support/src/formatting/formatting.ts
```diff
@@ -1130,6 +1131,23 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
...
+    this._visitTerminalRaw(ctx.COLON());
+    this.avoidBreakBetween();
+    if (ctx.IS()) {
+      this._visit(ctx.IS());
+      if (ctx.NOT()) {
+        this._visit(ctx.NOT());
+      }
+      if (ctx.LABELED) {
+        this._visit(ctx.LABELED());
+      }
```
alisasanib: I think `LABELED` is a method similar to `IS` and `NOT`. If so, maybe we need to call it here as well?

anderson4j: Ah I see what you mean - I check that method exists not method value, good catch!

packages/language-support/src/antlr-grammar/Cypher25Parser.g4
anderson4j: PS. I think it can be confusing that I made changes in `formatting.ts` when I here say I want to fix formatting things later. The change below is more of a fix, since the `labelComparison` used to have formatting handled by a visitor of it's children, but now that the child-rule changed (it's no longer using labelExpression, which we handle, instead it's a straight up labelExpression4) we need to handle it manually to avoid regression

alisasanib (approved):

PR #674, opened 2026-04-16
Split semantic analysis into two entry points

anderson4j:
Updates the transpiled `semanticAnalysis.js` to have 2 entry points for results, 1 for the symbol tables, and 1 for the linting. The old `analyzeQuery` method now only returns linting (like in days of yore) and `calculateSymbolTable` returns the symbol table. Both have had their parsing pipeline trimmed to not include unnecessary steps.

Most significant change for review is how these are now used. Just like the transpiled file which they call on, the worker files `lintWorker.cjs` and `lintWorker.mjs` now export 2 methods, 1 for linting and 1 for symbols. This will of course only happen in workers created from now on, so we have a split in how workers are handled - for old workers both linting and symbol tables are handled as before. For newer workers however, we run the appropriate method only.

Closes LS-137


packages/language-server/src/server.ts
```diff
@@ -81,18 +82,36 @@ class SymbolFetcher {
         const docUri = this.nextJob.uri;
         this.nextJob = undefined;
         const fixedDbSchema = convertDbSchema(dbSchema, this.linterVersion);
+        let symbolTables: SymbolTable[] | undefined;
 
-        const result = await proxyWorker.lintCypherQuery(query, fixedDbSchema);
+        if (
+          (this.linterVersion && this.linterVersion >= '2026.04') ||
```
OskarDamkjaer: Does this string comparison work for all our version numbers? I guess so because of the 0padding? In UPX that is discarded as the leading zero doesn't work in semvar parseing. Are we sure this is a robust comparison?

packages/language-support/src/tests/syntaxValidation/helpers.ts
```diff
@@ -21,5 +24,5 @@ export function getSymbolTablesForQuery({
   query,
   dbSchema = {},
 }: SyntaxValidationTestArgs) {
-  return lintCypherQuery(query, dbSchema).symbolTables;
+  return getSymbolTables(query, dbSchema);
 }
```
OskarDamkjaer: This function seems kind of useless now? Thinking we could just get rid of it? 

If we want to keep the `getSymbolTablesForQuery` name we can rename the inner one and export it instead? WDYT?

packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -322,12 +322,13 @@ class CodemirrorSymbolFetcher {
         const dbSchema = this.nextJob.schema;
         this.nextJob = undefined;
 
-        const result = await proxyWorker.lintCypherQuery(query, dbSchema);
+        //Add a similar check here as in SymbolFetcher of language-server when we add per-server linting
```
OskarDamkjaer: Ah you mean the version check is not needed here?

>per-server linting

Don't we already lint depending on the neo4j-server, or do you mean something else?

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -330,6 +333,50 @@ function fixSymbolTableOffsets({
...
+      parsingResult ?? createParsingResult(query, { consoleCommandsEnabled });
+    const statements = resolvedParsingResult.statementsParsing;
+    const symbolTables = statements.map((current) => {
+      const cmd = current.command;
+      if (
+        (cmd.type === 'cypher' || cmd.type === 'auto') &&
+        cmd.statement.length > 0
+      ) {
+        const rawSymbolTable = wrappedSymbolTableCollection(
+          cmd.statement,
```
OskarDamkjaer, line 358: I'm wondering if the `:auto` prefix is stripped here 🤔 I'll test it out.

packages/language-server/src/server.ts
OskarDamkjaer: as well as for the 5x series

packages/language-support/src/syntaxValidation/syntaxValidation.ts
OskarDamkjaer: I was seeing errors with tailwind when trying to start the demo page? I tried 
```
pnpm clean
pnpm install
pnpm build
pnpm dev-codemirror
```
Is it working for you?

OskarDamkjaer: a `pnpm install --force` did the trick!

OskarDamkjaer: When I try this in the codemirror demo, i get thousands of logs `CypherEditor.tsx:336 Symbol table calculation failed undefined`. Seems there's some crash looping?

OskarDamkjaer: Have you tested `:auto` works in vscode for variables?

packages/react-codemirror/src/CypherEditor.tsx
anderson4j: I mean we get our linting logic from transpiling parts of the server, but the linting in upx is not different per server you connect to

anderson4j: Unlike in vscode where we have one linter per server version

packages/language-server/src/server.ts
anderson4j: Ah you're right, I should be calling our helper `compareMajorMinorVersions`

packages/language-support/src/syntaxValidation/syntaxValidation.ts
anderson4j: Hadn't tested `:auto`, but I thought it should, since we set cmd.statement to just be the statement bit on parsing auto cmd. (see `parseToCommand` in `cypherLanguageService.ts`)

Did a double-check now by activating console commands in vscode playground, heres a screengrab
image

packages/react-codemirror/src/CypherEditor.tsx
OskarDamkjaer: can we mention in the comment what the check is instead of refferening to it?

anderson4j: It takes up a few rows so I dont want to inline the entire implementation. I've modified the comment a bit so its hopefully clearer

packages/language-server/src/server.ts
```diff
@@ -91,9 +91,10 @@ class SymbolFetcher {
           ? convertDbSchema(dbSchema, this.linterVersion)
           : dbSchema;
         let symbolTables: SymbolTable[] | undefined;
+        // The split in symbol table creation / linting was first introduced for 2026.05
         const versionComparison = compareMajorMinorVersions(
           this.linterVersion ?? '',
-          '2026.04',
+          '2026.05',
```
anderson4j, line 97: since we already built the 2026.04 linter, the old check was wrong

packages/language-support/src/tests/testData.ts
```diff
@@ -10,6 +10,59 @@ function asNeo4jProcedures<T extends Record<string, Neo4jProcedure>>(obj: T) {
 }
 
 const mockSchema: DbSchema = {
+  graphSchema: [
```
anderson4j, line 13: This way we can try out schema based completions in codemirror demo

OskarDamkjaer (approved):

PR #675, opened 2026-04-16
Expand formatter coverage

anderson4j:
Things have been busy over the past couple of months, and the formatter has not been kept up to date with recent grammar updates. In addition, it seems like some old syntax like `CREATE DATABASE` was not covered. This PR is to fix this, updating the formatter and adding test with example syntax.

Closes LS-145


.gitignore
```diff
@@ -18,5 +18,6 @@ generated
 *.npmrc
 benchmarks.txt
 .idea/
+.claude/
```
OskarDamkjaer: Git-ignoring this one but commiting CLAUDE.md?

OskarDamkjaer: IMO it could be interesting to add some skills/etc to the repo, for new contributors, or for automated code review? Let's discuss later in our 121?

CLAUDE.md
```diff
@@ -0,0 +1,109 @@
+# CLAUDE.md
```
OskarDamkjaer: I think a shared claude.md would be a good idea, but I think it'd be nice to collaborate on it. I found some of the best practises from anthropic counter intuitive:

https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md

They say to exclude for example anything Claude can figure out by reading code.

packages/language-support/src/tests/formatting/commands.test.ts
```diff
@@ -127,9 +127,11 @@ OPTIONS {
   });
 
   test('show indexes with where and return', () => {
-    const query = `SHOW INDEXES YIELD aaaaaa, aaaaaa, aaaaaa, aaaaaa, aaaaaa
+    const query = `SHOW INDEXES
```
OskarDamkjaer, line 130: How did you find the queries we fail to format? I'm thinking it could be nice if we could find them automatically? Perhaps we could extract samples from the main repos tests?

OskarDamkjaer (approved):
Formatting changes good to go, the claude changes would be nice to pair on or take in another PR

packages/language-support/src/tests/formatting/commands.test.ts
anderson4j: I agree, ive discussed this with some surface people before, and they mentioned some `.feature` files that often contain queries with new syntax. I've made a card [here ](https://linear.app/neo4j/issue/LS-153/automatic-tests-for-formatting-of-new-syntax-based-off-feature-tests)

.gitignore
anderson4j: Like we said in 1-2-1 I agree. Will clean up this PR when it's ready for a full review

CLAUDE.md
anderson4j: 👍

anderson4j (commented):
Here are some replies to your comments. I still have a few months of changes to the grammar left, so I'll fix that first before a re-review

packages/language-support/src/formatting/formatting.ts
```diff
@@ -568,6 +612,14 @@ export class TreePrintVisitor extends CypherCmdParserVisitor<void> {
     }
   }
 
+  _emitResidualErrorNodes = () => {
```
anderson4j, line 615: We already have a check to emit missed error nodes when visiting a terminal node. The one case we missed was when the error node was in the end. This change is to emit the leftover error nodes too.

anderson4j:
To find new syntax we hadn't covered I checked changes to the [Cypher25Parser.g4 file in he monorepo](https://github.com/neo-technology/neo4j/commits/dev/public/community/cypher/front-end/parser/v25/parser/src/main/antlr4/org/neo4j/cypher/internal/parser/v25/Cypher25Parser.g4). 

To then figure how it should be formatted I checked what PRs the changes were for and checked what the canonical syntax looked like for this feature in the [corresponding CIPs](https://drive.google.com/drive/folders/1HoJG_ZY-8GflKoIpsWrQz8-7UegFwL0I). I also checked the cypher styling guide for the general formatting rules.

OskarDamkjaer (approved):
:+1:

PR #681, opened 2026-05-18
Schema based linting

anderson4j:
Adds linting using the symbol table for invalid paths in a query. Warns on each segment that is invalid like
`MATCH (:Trainer)-[:WEAK_TO]->(:Type)` -> "Path segment does not exist on graph." marking the `(:Trainer)-[:WEAK_TO]->` segment

or

`MATCH (:Trainer)-[:WEAK_TO]->(:Gym)` -> "Path segment does not exist on graph." x2, once marking `(:Trainer)-[:WEAK_TO]->` and once marking `-[:WEAK_TO]->(:Gym)`

Limitation: Re-uses a modified version of the completion logic, so it currently only supports a single label on the second variable in a path segment - otherwise ignores the segment.

Warnings are deliberately skipped inside CREATE / INSERT / MERGE clauses.

While making the PR I ran across and fixed some places where the tighter TS checks of TS 6 (now default in vscode) showed some type issues (usually ~"value not checked for null when it could be")

Closes LS-103

### DIFF
--- a/packages/language-support/src/autocompletion/completionCoreCompletions.ts
+++ b/packages/language-support/src/autocompletion/completionCoreCompletions.ts
@@ -307,11 +307,11 @@ const parameterCompletions = (
         ? { insertText: `$${backtickedName}` }
         : {};
       // If there is a preceding token and it's not empty, compute the suffix
-      if (previousToken.type !== CypherLexer.SPACE) {
+      if (previousToken && tokens && previousToken.type !== CypherLexer.SPACE) {
         const param = maybeInsertText.insertText ?? `$${paramName}`;
         const hasDollar =
           previousToken.type === CypherLexer.DOLLAR ||
-          tokens.at(previousToken.tokenIndex - 1).type === CypherLexer.DOLLAR;
+          tokens.at(previousToken.tokenIndex - 1)?.type === CypherLexer.DOLLAR;
         // If the $ symbol is already there, we need to have the insert
         // text without the starting $ in VSCode, otherwise when we have
         // 'RETURN $' and we get offered $param we would complete
--- a/packages/language-support/src/autocompletion/schemaBasedCompletions.ts
+++ b/packages/language-support/src/autocompletion/schemaBasedCompletions.ts
@@ -202,7 +202,7 @@ const reltypesToCompletions = (reltypes: string[] = []) =>
 export const allReltypeCompletions = (dbSchema: DbSchema) =>
   reltypesToCompletions(dbSchema.relationshipTypes);
 
-function walkCNFTree(
+export function walkCNFTree(
   incomingLabels: Map<string, Set<string>>,
   outGoingLabels: Map<string, Set<string>>,
   labelTree: LabelOrCondition,
@@ -271,7 +271,7 @@ function walkCNFTree(
   return { inLabels, outLabels };
 }
 
-function getRelsFromNodesSets(dbSchema: DbSchema): {
+export function getRelsFromNodesSets(dbSchema: DbSchema): {
   toNodes: Map<string, Set<string>>;
   fromNodes: Map<string, Set<string>>;
 } {
--- a/packages/language-support/src/autocompletion/schemaBasedCompletions.ts
+++ b/packages/language-support/src/autocompletion/schemaBasedCompletions.ts
@@ -5,13 +5,7 @@ import {
 } from 'vscode-languageserver-types';
 import { DbSchema } from '../dbSchema.js';
 import { ParsedStatement } from '../cypherLanguageService.js';
-import {
-  ConditionNode,
-  isLabelLeaf,
-  LabelLeaf,
-  LabelOrCondition,
-  SymbolsInfo,
-} from '../types.js';
+import { isLabelLeaf, LabelOrCondition, SymbolsInfo } from '../types.js';
 import { findParent } from '../helpers.js';
 import {
   NodePatternContext,
@@ -26,6 +20,11 @@ import {
   isNotAnyNode,
   removeInnerAnys,
 } from '../labelTreeRewriting.js';
+import {
+  getNodesFromRelsSet,
+  getRelsFromNodesSets,
+  walkCNFTree,
+} from '../labelTreeWalking.js';
 
 export function getShortPathCompletions(
   lastNode: RelationshipPatternContext,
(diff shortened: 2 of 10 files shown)


packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -307,11 +307,11 @@ const parameterCompletions = (
         ? { insertText: `$${backtickedName}` }
         : {};
       // If there is a preceding token and it's not empty, compute the suffix
-      if (previousToken.type !== CypherLexer.SPACE) {
+      if (previousToken && tokens && previousToken.type !== CypherLexer.SPACE) {
```
anderson4j, line 310: just cleaning up some warnings from the tighter typescript checks

packages/language-support/src/autocompletion/schemaBasedCompletions.ts
```diff
@@ -202,120 +201,6 @@ const reltypesToCompletions = (reltypes: string[] = []) =>
 export const allReltypeCompletions = (dbSchema: DbSchema) =>
   reltypesToCompletions(dbSchema.relationshipTypes);
 
-function walkCNFTree(
```
anderson4j, line 205: shared logic with the new schema based linting - moved to its own file

packages/language-support/src/autocompletion/schemaBasedCompletions.ts
```diff
@@ -328,7 +213,7 @@ function findLastVariable(
           // Because the anonymous variable created in the AST is "made up", it doesnt have a position of its own in the query.
           // It therefor inherits the parent-nodes (The NodePattern/RelationshipPattern = lastValidElement) position
           entry.definitionPosition >= lastValidElement.start.start &&
-          entry.definitionPosition <= lastValidElement.stop.stop,
+          entry.definitionPosition <= (lastValidElement.stop?.stop ?? -1),
```
anderson4j, line 216: also for a ts check (stop can be undefined)

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -258,6 +287,168 @@ function warnOnUndeclaredLabels(
...
+  return diagnostics;
+}
+
+function findPatternElements(
+  ctx: ParserRuleContext,
+  acc: PatternElementContext[],
+): PatternElementContext[] {
+  for (const c of ctx.children ?? []) {
+    if (
+      c instanceof MergeClauseContext ||
```
anderson4j, line 513: We would never want to warn of "Path segment doesnt exist" for clauses that create new paths. Currently we dont parse the labels of variables in these clauses anyway, but if we do we would need to skip them. No harm in skipping early here, and we future proof if we start parsing labels there. `InsertClause` currently doesnt use `PatternElement` either, but once again - if it did we wouldnt want to warn

packages/vscode-extension/tests/specs/api/syntaxValidation.spec.ts
```diff
@@ -236,6 +236,14 @@ suite('Syntax validation spec', () => {
     await testSyntaxValidation({
       docUri,
       expected: [
+        new vscode.Diagnostic(
+          new vscode.Range(
+            new vscode.Position(0, 8),
+            new vscode.Position(0, 32),
+          ),
+          'Path segment does not exist on graph.',
+          vscode.DiagnosticSeverity.Warning,
```
OskarDamkjaer, line 245: nice!

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -381,6 +565,20 @@ export function lintCypherQuery(
...
+        });
+
+        let missingPathWarnings: SyntaxDiagnostic[] = [];
+        if (dbSchema.graphSchema) {
+          missingPathWarnings = warnOnPathDirectionalityIssues(
+            current,
+            dbSchema,
+            symbolTable,
+          );
+        }
```
OskarDamkjaer: it's purely a stylistic/opinion thing but I would write this like 
```ts
const missingPathWarnings = dbSchema.graphSchema ?  warnOnPathDirectionalityIssues(current, dbSchema, symbolTable) : []
```

Because then I don't need to type out the `SyntaxDiagnostic` type annotation, I avoid a nested code block (even if it's a very simple block), as well as avoiding mutating the variable

packages/language-support/src/labelTreeWalking.ts
```diff
@@ -0,0 +1,119 @@
+import { DbSchema } from './dbSchema.js';
+import {
+  ConditionNode,
+  isLabelLeaf,
+  LabelLeaf,
+  LabelOrCondition,
+} from './types.js';
+
```
OskarDamkjaer, line 8: This was the moved function right? assuming there's not many changes here

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -258,6 +287,161 @@ function warnOnUndeclaredLabels(
...
+          if (
+            possibleRels &&
+            nextSymbolLabels &&
+            'children' in nextSymbolLabels &&
+            nextSymbolLabels.children.length === 1 &&
+            isLabelLeaf(nextSymbolLabels.children[0])
+          ) {
+            if (!possibleRels.has(nextSymbolLabels.children[0].value)) {
+              diagnostics.push({
+                message: 'Path segment does not exist on graph.',
```
OskarDamkjaer: Do we want to have a dynamic message here? For example:

`:Movie has no outgoing :ACTED_IN relationship`

exact wording tbd.

packages/language-support/src/tests/syntaxValidation/symbolValidation.test.ts
```diff
@@ -0,0 +1,668 @@
...
+    ]);
+  });
+
+  test('Does not warn on valid path segments. <-Rel-Node', () => {
+    const query = 'MATCH ()<-[:WEAK_TO]-(:Pokemon) RETURN ""';
+    const diagnostics = getDiagnosticsForQuery({ query, dbSchema });
+    expect(diagnostics).toEqual([]);
+  });
+
+  test('Warns on multiple invalid path segments at once', () => {
```
OskarDamkjaer, line 175: esp in cases like this it'd be nice to have the specificity I meantioned earlier

packages/language-support/src/tests/syntaxValidation/symbolValidation.test.ts
```diff
@@ -0,0 +1,668 @@
...
+            character: 33,
+            line: 0,
+          },
+        },
+        severity: 1,
+      },
+    ]);
+  });
+
+  test('Should not fail on broken rel', () => {
```
OskarDamkjaer, line 535: nice to check we don't get warnings on errors

packages/language-support/src/tests/syntaxValidation/symbolValidation.test.ts
```diff
@@ -0,0 +1,668 @@
...
+            character: 24,
+            line: 0,
+          },
+        },
+        severity: 1,
+      },
+    ]);
+  });
+
+  test('Warns both on missing label and path segment at once', () => {
```
OskarDamkjaer, line 715: Do we want to show both here? Maybe it's more useful to fix the mistyped labels/rels first

OskarDamkjaer (approved):

packages/language-support/src/labelTreeWalking.ts
anderson4j: Yep these 3 were copypasted from `schemaBasedCompletions.ts`, just so that it's more clear
Not
"schema based linting" -needs-> "schema based completions" 
but rather 
 ["schema based linting", "schema based completions"] -need-> "tree logic"

packages/language-support/src/tests/syntaxValidation/symbolValidation.test.ts
anderson4j: Yeah I was in 2 minds about this one. My thoughts why it is as is:
Reasons for:
 We dont need to pass in parsed labels/missing label warnings to the schema based linting (cleaner). Doing this we would also run into cases like "variable has label X that doesnt exist, but it's not in the specified path. Is it still not interesting to have the warning? Will it be misleading if we have an overlap of both errors, but hide the one until the other is resolved?

ex. 
```
MATCH (n:Gibberish)
... (maybe many many lines of query)
MATCH (n:Person)-[r:KNOWS]->()
RETURN r
```
A( weak reason): if `Gibberish` doesnt exist but `Person` does, do we hide "path segment does not exist" on `MATCH (n:Person)-[:KNOWS]->()`? Thinking a user might focus on the `r` part of the query when they dont get any rows.
B(etter reason): if `Gibberish` doesnt exist, but neither does the path segment `(n:Person)-[r:KNOWS]` we would first not show "path segment doesnt exist" and then add it back when `n:Gibberish` is removed. Might confuse a user to see that adding `n:Gibberish` "fixes" the path segment.

Reasons against:
 missing label always causes "path segment could not exist". This is true, but not more useful than just "label does not exist"

packages/language-support/src/syntaxValidation/syntaxValidation.ts
anderson4j: I agree that its more useful in this example. I think it would be quite a mess to write out for more complex label trees though. Theres no limit to the depth of course, but even `(n:Person & (Friend | Neighbour)` is going to be quite messy to write out imo. When we (hopefully) support multiple labels on the second variable it's going to get even more ugly with 2 trees. 

One solution would be to only give a "good" message on single-label pairs, which I think are quite common, and then fall back to `Path segment does not exist on graph.` on trees with multiple labels.

anderson4j: Small question mark on how something like
```
MATCH (n:Trainer)
... (possible many lines) 
MATCH (n)-[:TYPE]->() 
```
Would look. Saying `:Trainer has no outgoing :TYPE relationship` when `Trainer` is not specified at the segment might be a bit confusing

OskarDamkjaer: >Theres no limit to the depth of course, but even (n:Person & (Friend | Neighbour) is going to be quite messy to write out imo. 

yes it's messy but how are you going to debug what's wrong if you can't see it? It's just like type errors in other programming languages, they get long and possibly confusing but it's the best we have

OskarDamkjaer: How useful would it be if typescript just said "Type error"?

packages/language-support/src/tests/syntaxValidation/symbolValidation.test.ts
OskarDamkjaer: I think it'd be better if the warning is more clear what's going on. for example 
` (Person&Gibberish)-[:KNOWS] path segment does not exist`, but I'm still on the fence whether I think they should be in the same category/happen at the same time

packages/language-support/src/syntaxValidation/syntaxValidation.ts
anderson4j: you're right, it's quite a bit more useful with types, and now that ive implemented it I'd say even some quite big cases dont look that bad in the end.

packages/language-support/src/tests/syntaxValidation/symbolValidation.test.ts
anderson4j: I'll leave it as my preference with the double warnings then

PR #682, opened 2026-05-20
Update neovim config

meistermeier:


anderson4j (approved):
Looks correct! Had to install nvim for the first time to try it out, and I came across your old blog post, where I see you even get linting. Just installing with lspconfig without any extra fluff, most things (formatting/syntax highlighting/auto completions) work out of the box, but I don't seem to get any linting from the server.

Do you still get linting (and my conf is just incomplete), or have we broken the server since your blog post?

meistermeier:
> or have we broken the server since your blog post?

I wonder the same, because _after_ I raised the PR, I updated the server because the example JS project has multi statement support, which I was missing. After the update there was no linting anymore.
Need to figure out if I would need to do something else regarding configuration or if this is a server problem. Will not merge and first iterate on this when I find time.

PR #683, opened 2026-05-21
Brow-341: inline widget panel

alisasanib:
This PR adds an inlinePanel prop to CypherEditor for rendering a  panel as a block widget inside the editor controlled by the caller. The prop has pos, optional placement, and onMount, onUnmount callbacks. The caller renders its own content into the provided DOM container, via createPortal for example, so react-codemirror will continue to not have react-dom as a dep.([ref to similar discussions](https://discuss.codemirror.net/t/rendering-react-components-or-similar-in-decoration-todom/3492/6)) While the panel is open, new onMount / onUnmount references are picked up without remounting and caller don't need to memoize callbacks. Changing pos or placement repositions the block widget, which removes the existing DOM and remounts at the new position.
The widget also manages a ResizeObserver so the editor remeasures its layout when host content resizes.

### DIFF
--- a/.changeset/five-ghosts-joke.md
+++ b/.changeset/five-ghosts-joke.md
@@ -1,0 +1,6 @@
+---
+"@neo4j-cypher/react-codemirror-playground": patch
+"@neo4j-cypher/react-codemirror": patch
+---
+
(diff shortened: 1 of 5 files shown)


OskarDamkjaer (approved):
Works nicely! thanks for adding the extra demo piece

PR #685, opened 2026-05-27
Improve readability of schema based warnings

anderson4j:
Neo4j has some errors when the query is unfinished, which is almost all cases we work with. The errors cover a large chunk of the query, and obscure some other errors/warnings. In an effort to make other warnings more visible, this PR moves the startposition to the end of the error for errors containing "Query cannot conclude with", so we still get the error, but only at the point where the invalid ending is, not the entire clause.
Ex:
`MATCH (n:Person)-[:KNOWS]->(:Person)` would get linting from start to finish. Now we show it only at the finish point.

The schema based warnings also became quite verbose in the last iteration, so we switched to a shorter variant (see examples in tests).

Closes LS-150


packages/react-codemirror/src/e2e_tests/e2eUtils.ts
```diff
@@ -69,6 +69,30 @@ export class CypherEditorPage {
     });
   }
 
+  //So we can test errors without needing to match text
```
anderson4j: Reminder to self - update this msg

packages/vscode-extension/tests/specs/api/syntaxValidation.spec.ts
```diff
@@ -241,7 +241,7 @@ suite('Syntax validation spec', () => {
             new vscode.Position(0, 8),
             new vscode.Position(0, 32),
           ),
-          'Relationship with label ACTED_IN has no incoming connection to a node with label(s) Person.',
+          '[:ACTED_IN] has no incoming (:Person)',
```
OskarDamkjaer, line 253: Nice!

packages/language-support/src/tests/syntaxValidation/syntacticValidation.test.ts
```diff
@@ -1687,4 +1687,38 @@ describe('Syntactic validation spec', () => {
       }),
     ).toEqual([]);
   });
+
+  test('No long message on unfinished query', () => {
```
OskarDamkjaer: Nit/Doesn't need to be changed, but i find it's easier to read what we want to see, rather than what we dont't want to see. So `Unfinished query error covers only end of query string` for example

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -641,9 +617,18 @@ export function lintCypherQuery(
...
         );
+        const cleanedErrors = errors.map((x) =>
+          x.message.includes('Query cannot conclude with')
+            ? {
+                ...x,
+                offsets: { start: x.offsets.end, end: x.offsets.end },
+                range: { start: x.range.end, end: x.range.end },
+              }
+            : x,
+        );
```
OskarDamkjaer, line 643: I'd like us to be descriptive and avoid `cleaned/fixed/updated/helper`. `adjustedPositionErrors` or `offsetErrors` could be better. If it's hard to find a good outer name we can name the inner one.

```ts
function moveUnfinishedErrorToEnd(x) {
  return x.message.includes('Query cannot conclude with')
            ? {
                ...x,
                offsets: { start: x.offsets.end, end: x.offsets.end },
                range: { start: x.range.end, end: x.range.end },
              }
            : x
}

const cleanedErrors = errors.map(moveUnfinishedError);
```

OskarDamkjaer: More importantly: I also wonder if we can make the check more stable? The error message can be changed without being a breaking change. Is there also an error code we could look for?

anderson4j: There is a code, 41N72. I too was thinking this would be nice to use, but we throw out that info on the java side, only passing on { message, startposition, endposition } to js.

We could adjust on the Java side to pass on the Gql error object/adjust based on GQL error
(**edit: removed some inaccuracy here, I mixed up this change with the "remove param errors" pr**)

We also don't use the GQL message right now which might be a bit confusing to the users, compared to say browser. (Although I guess that mean the message wont change, since they only modify the GQL one iirc)

anderson4j: After sitting with it a bit I think we definitely should include GQL errors in the linter to some extent, at the very least modifying messages to use GQL messages (but then, why not include the codes etc too). Right now we have a disparity between what is shown when you run a query in Query (the app with unfortunate naming), and what the linter shows. Would need to do some digging/thinking to decide what else from the GQL object we want to include (all of it?)

OskarDamkjaer: I think it'd make sense to pass on the whole GQL error unless there's some reason it doesn't work

OskarDamkjaer (approved):
I'm happy to ship this change as is now and then add the more robust GQL errors in a followup PR

PR #688, opened 2026-05-28
Filter out "parameter missing" errors when not connected

anderson4j:
When disconnected one can not add new params, since we rely on evaluating their value against the current db. (see browser example)
image
It's therefor not very helpful to give this linting to the user - rather we can wait until they connect, and then provide the linting + let them add the param.

In order for the change to be retroactive for old linter files ive added the removal of this error as a filtering on the results from the lint worker in the language-server/in react-codemirror. The check for whether to do this is if we have a schema with dbs or not (no dbs = disconnected from dbms).

Here it is in action:
Connected -> error
image
Disconnected -> no error
image

We also had a really flaky test suite that made the CI painful. Added a fix for that too (that suite now passed 7 times in a row after primary fix 🥳, where before it would fail more often than not)

Closes LS-151

### DIFF
--- a/packages/language-support/src/syntaxValidation/syntaxValidation.ts
+++ b/packages/language-support/src/syntaxValidation/syntaxValidation.ts
@@ -89,12 +89,12 @@ function detectNonDeclaredLabel(
 
 type GenericDiagnostic = { message: string };
 
-export function filterParams<T extends GenericDiagnostic>(
-  diagnostics: T[],
-): T[] {
-  const paramRegex = /Parameter .+ is not defined./;
-  return diagnostics.filter(
-    (x) => x.message.length > 1000 || !x.message.match(paramRegex),
+export function isNotParamError<T extends GenericDiagnostic>(
+  diagnostic: T,
+): boolean {
+  return (
+    !diagnostic.message.startsWith('Parameter ') ||
+    !diagnostic.message.endsWith(' is not defined.')
   );
 }
 
@@ -812,6 +812,8 @@ function warningOnDeprecatedFunction(
   }
   return warnings;
 }
+
+export const paramMsgStart = 'Parameter ';
 
 function errorOnUndeclaredParameters(
   parsingResult: ParsedStatement,
--- a/packages/language-support/src/syntaxValidation/syntaxValidation.ts
+++ b/packages/language-support/src/syntaxValidation/syntaxValidation.ts
@@ -87,6 +87,15 @@ function detectNonDeclaredLabel(
   return undefined;
 }
 
+type GenericDiagnostic = { message: string };
+
+export function filterParams<T extends GenericDiagnostic>(
+  diagnostics: T[],
+): T[] {
+  const paramRegex = /Parameter .+ is not defined./;
+  return diagnostics.filter((x) => !x.message.match(paramRegex));
+}
+
 export function clampUnsafePositions(
   diagnostics: SyntaxDiagnostic[],
   document: TextDocument,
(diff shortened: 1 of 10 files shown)


packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -87,6 +87,15 @@
   return undefined;
 }
 
+type GenericDiagnostic = { message: string };
+
+export function filterParams<T extends GenericDiagnostic>(
+  diagnostics: T[],
+): T[] {
+  const paramRegex = /Parameter .+ is not defined./;
+  return diagnostics.filter((x) => !x.message.match(paramRegex));
```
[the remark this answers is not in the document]
anderson4j: added bailout if message is really long

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -87,6 +87,17 @@ function detectNonDeclaredLabel(
...
+type GenericDiagnostic = { message: string };
+
+export function filterParams<T extends GenericDiagnostic>(
+  diagnostics: T[],
+): T[] {
+  const paramRegex = /Parameter .+ is not defined./;
+  return diagnostics.filter(
+    (x) => x.message.length > 1000 || !x.message.match(paramRegex),
+  );
+}
```
OskarDamkjaer, line 99: could we break out this functino and name it according to what it does? Why do we check the length btw?

OskarDamkjaer: So that the `.filter` happens at the call site

anderson4j: Got a message from the github bot that regex on arbitrary string can be very slow for large strings. Added this as a safeguard against that, though I don't expect us to run into it. Not sure how the background logic for regex would work here but `message.startsWith()` + `message.endsWith` should be constant, so Ill just use that instead and remove the limit.

packages/vscode-extension/tests/specs/api/lintSwitching.spec.ts
```diff
@@ -29,6 +29,21 @@ suite('Lint switching spec', () => {
   let showErrorMessageSpy: sinon.SinonSpy;
 
   before(async () => {
+    textDocument = await newUntitledFileWithContent(`MATCH (n) RETURN m`);
+    await saveDefaultConnection();
+    await testSyntaxValidation({
```
anderson4j, line 34: This way we ensure initial connection is settled to where linting shows

packages/vscode-extension/tests/specs/api/lintSwitching.spec.ts
```diff
@@ -50,7 +65,6 @@ suite('Lint switching spec', () => {
       'switchWorkerOnLanguageServer',
     );
     sendNotificationSpy = sandbox.spy(mockLanguageClient, 'sendNotification');
-    await saveDefaultConnection();
```
anderson4j, line 53: connecting triggers "dynamicallyAdjustLinter" so doing this pre every test caused racing

packages/vscode-extension/tests/specs/api/syntaxValidation.spec.ts
```diff
@@ -74,7 +77,8 @@ suite('Syntax validation spec', () => {
     await toggleLinting(true);
   });
 
-  test('Suggests replacements for deprecated functions/procedures', async () => {
+  test('Suggests replacements for deprecated functions/procedures', async function () {
+    this.timeout(60000);
```
anderson4j, line 81: The big flaky test fix was to lintSwitching.spec.ts, but some flakes remained flaky in other suites. Changes to timeout (and `eventually` are attempts to mitigate those flaky tests, but it's a bit tricky (and time consuming!) to measure how big an impact these other changes have). I figure we go with this, seems pretty stable, if not perfect

OskarDamkjaer (approved):

PR #689, opened 2026-05-29
Test CI issues

anderson4j:
CI looks to be having issues in react-codemirror e2e and webview tests. Pretty much a copy of main to run the CI on and see if we can fix it


anderson4j:
Looks like this doesnt test what I want, think I need to manually revert changes, rather than remove the commit, otherwise CI tests "main + new changes" and considers this an out of date branch, rather than a reverting branch

PR #690, opened 2026-06-01
Investigate CI failures

anderson4j:
First tried reverting commit 2d5fa772499eaf0f8ee3b019e56e3d837cf4080a.

This did not fix things. Oskar's Claude PR mentions issues with playwright and node. One comment suggested issues would be fixed in playwright `1.60.0` - trying that here (+ updating webdriverio)

Looks like updating fixes playwright, but not webdriverio


anderson4j:
Using Oskar's fix of pinning node version for now, made a linear card in case we want to unpin later

PR #693, opened 2026-06-03
Brow 326 add merge view

alisasanib:
This PR adds diff view to the Cypher editor using `@codemirror/merge`. When consumers pass `diff={{ original: '<query>' }}`, the editor renders a read-only diff between the original and the current document. We are using ndl design tokens for diff view color for dark/light themes. 
The diff is computed between `original` and the `current` editor document, so external updates to the document re-diff automatically against the same original.
An example has also been added to `react-codemirror-playground`.


packages/react-codemirror-playground/src/App.tsx
```diff
@@ -80,6 +81,19 @@ export function App() {
...
+  const [diff, setDiff] = useState<DiffProps | null>(null);
+  let beforeValue: string = '';
+  const showDiff = useCallback(() => {
+    beforeValue =
+      editorRef.current?.editorView.current?.state.doc.toString() ?? '';
+    setDiff({ original: beforeValue });
+  }, []);
+  const acceptDiff = useCallback(() => setDiff(null), []);
+  const rejectDiff = useCallback(() => {
+    setValue(beforeValue);
```
anderson4j, line 141: This is the change I showed yesterday to improve the codemirror demo. Have a look and see if you want to modify anything @alisasanib

packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -732,6 +742,25 @@ export class CypherEditor extends Component<
...
     }
 
+    if (prevProps.diff?.original !== this.props.diff?.original) {
+      this.editorView.current.dispatch({
+        effects: diffCompartment.reconfigure(
+          this.props.diff ? createDiffExtension(this.props.diff) : [],
+        ),
+      });
+
+      // When the diff first appears the inline panel widget may have drifted
```
anderson4j: Im struggling to find the problem this fixes. Tried commenting it out and behaviour looks the same? See videos I sent on slack

anderson4j (commented):
Had one piece of code I couldnt figure out why we need, but otherwise LGTM.

anderson4j (approved):
Nice improvement on the demo! Seems like having the new "demo diff" as well could be useful for debugging behaviour like the jumping we discussed earlier. 

One note to confirm it's intended:
The inline panel no longer jumps to the current line (which is what I thought was intended originally) - I think this is on purpose?

If so, LGTM

alisasanib:
> Nice improvement on the demo! Seems like having the new "demo diff" as well could be useful for debugging behaviour like the jumping we discussed earlier.
> 
> One note to confirm it's intended: The inline panel no longer jumps to the current line (which is what I thought was intended originally) - I think this is on purpose?
> 
> If so, LGTM

Yes the jump behavior was a bug and not intended.

PR #697, opened 2026-06-24
add editor action buttons overlay to CypherEditor

nnecla:
This PR is to add support for a host-rendered cluster of action buttons pinned to the top-right corner of the editor, via a new `editorActions` prop.

This overlay uses two cooperating pieces so it can both reserve space and stay pinned (one element can't do both): a floated spacer widget at the start of the doc reserves the footprint, and an absolutely-positioned container holds the actual buttons so they don't move as the doc grows or scrolls.

Also refactors the shared {onMount, onUnmount} lifecycle contract into `hostCallbacks.ts`, reused by both `editorActions` and the existing `inlinePanel`.


alisasanib (approved):
:fire: :fire:

PR #698, opened 2026-06-25
Symmetric schema based linting

anderson4j:
Previously the the schema based linting was basically the completions logic, and checking if the RHS label was in completions. To make this work, we limited the RHS to a single label, so we could only handle segments like `(:Person&Friend)-[:KNOWS]` or `[:KNOWS|IS]->(:President)` but not `[:KNOWS|IS]->(:President|(Senator&Presidential_Candidate)` This PR expands the logic (still reusing completions) to handle generic labels

The new logic works like:
Split incoming tree to Disjunctive Normal Form (similar to existing Conjunctive Normal Form). The tree is then guaranteed to be on shape:
OR(A, !B, AND(C,D), AND(E,!F)...)
Also write the second tree to DNF
Ex. OR(X, AND(Y,Z))
if any of the OR-ed labels/ANDs of the LHS (left hand side) tree have a matching label in the OR of the RHS tree, there is a possible match. 

How do we deal with the ANDs then? My implementation takes "what labels come out of AND(C,D)?" from the existing completion logic. If the RHS is also AND, we can say - given that the label can't be any of the NOT-ed labels - can find all the required labels in the labels coming out of the LHS? If so, we've found a viable candidate, and the segment should not have a warning. For the single labels, the logic is like before - if a single completion of what could come after one of the possible LHS labels matches a RHS label we have a viable segment

I also found some bugs in the old logic that I fixed along the way

### DIFF
--- a/packages/language-support/src/labelTreeRewriting.ts
+++ b/packages/language-support/src/labelTreeRewriting.ts
@@ -17,9 +17,55 @@ function copyLabelTree(labelTree: LabelOrCondition): LabelOrCondition {
 }
 
 /**
- * Takes a label tree with an AND-root and converts it to Conjunctive Normal Form
+ * Takes a label tree with an AND-root and converts it to Distjunctive Normal Form
+ * also simplifies away redundant conditions
  * @param root - the original label tree
- * @returns a an equivalent CNF tree
+ * @returns a an equivalent DNF tree
+ */
+export function convertToSimplifiedDNF(
+  root: LabelOrCondition,
+): LabelOrCondition {
+  const dnfRoot: LabelOrCondition = convertToDNF(root);
+  return simplifyAndRemoveDNFContradictions(dnfRoot);
+}
+
+/**
+ * Takes a label tree with an AND-root and converts it to Distjunctive Normal Form
+ * Uses a trick with De Morgan's law (ex. !AND(X,Y) = OR(!X, !Y)) to turn CNF to DNF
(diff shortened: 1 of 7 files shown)


packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -318,7 +308,9 @@ function labelTreeToString(node: LabelOrCondition): string {
   if (isLabelLeaf(node)) {
     return node.value;
   } else if (node.condition === 'any') {
-    return 'ANY';
+    return '%';
```
anderson4j, line 317: Had some issues in this from before - printed ANY instead of %, which I now think looks less nice. Also NOT printing was broken

packages/language-support/src/labelTreeRewriting.ts
```diff
@@ -269,68 +313,224 @@ function pushInOr(orCondition: LabelOrCondition): LabelOrCondition {
 export function simplifyAndRemoveTautologies(
```
anderson4j, line 313: Found some issues/room for improvement here. The switch to make `const simplifiedChildren: (LabelOrCondition | undefined)[]` and then call `.filter(x => x !== undefined)` is not a change to functionality, but to be type-safe (LabelOrCondition.children is just `LabelOrCondition[]`)

anderson4j:
Added in a simplified approach on a [separate branch](https://github.com/neo4j/cypher-language-support/pull/700)

PR #700, opened 2026-06-30
Simplify and expand schema based linting

anderson4j:
Previously the the schema based linting was basically the completions logic, and checking if the RHS label was in completions. To make this work, we limited the RHS to a single label, so we could only handle segments like `(:Person&Friend)-[:KNOWS]` or `[:KNOWS|IS]->(:President)` but not `[:KNOWS|IS]->(:President|(Senator&Presidential_Candidate)` This PR expands the logic (still reusing completions) to handle generic labels.

[My initial approach](https://github.com/neo4j/cypher-language-support/pull/698) to do this involved rewriting trees to Disjunctive Normal Form (DNF) (which can be done to any tree of ands/ors/nots/values), and then reason on a tree with a known shape. This worked, but took quite some reasoning and thought to verify when reading through. This PR instead uses the fact that for any valid path segment, we can have one and only one reltype on the relationship. We can thus assemble the valid rel types, and for each check if there is any combination of labels on a node that would pass the label declaration (through brute-force, so if n different labels are part of the label tree of the node, we would check 2^n combinations). Exponential complexity, so it can blow up, but even for 10 different labels (quite a large tree) we just get 1024 checks. This approach is also a lot easier to read through and understand in my opinion.

Like with DNF solution, we can now handle any labels on either end of a segment (:<...>)-[:<...>] or [:<...>]-(:<...>) for example
`(n:Person&Friend)-[:MELTS_AT|CONSTRUCTED_IN]` <- Relationship with relationship type(s) (MELTS_AT | CONSTRUCTED_IN) has no incoming connection to a node with label(s) (Person & Friend).

I also expanded the testing to contain many more edgecases, such as contradictions matches with "!%" etc.

The new file for schema based validation (`schemaBasedValidation.ts`) is Claude's creation, but I've gone through the whole thing and adjusted some things I thought were unclear/unnecessary

Closes LS-165


packages/language-support/src/helpers.ts
```diff
@@ -209,6 +209,23 @@ export function resolveCypherVersion(
   return cypherVersion;
 }
 
+export function getDirection(
```
anderson4j, line 212: shared logic for SB-completions and SB-linting, could see us wanting it at other times

packages/language-support/src/labelTreeRewriting.ts
```diff
@@ -595,9 +595,12 @@ export function pushInNots(labelTree: LabelOrCondition): LabelOrCondition {
 }
 
 export const isAnyNode = (n: LabelOrCondition) =>
-  !isLabelLeaf(n) && n.condition === 'any';
+  !isLabelLeaf(n) && n.condition === 'any' && n.children.length === 0;
```
anderson4j, line 598: Should always be the case that if we have any/not that they have 0/1 children

packages/language-support/src/autocompletion/schemaBasedCompletions.ts
```diff
@@ -44,7 +44,7 @@ export function getShortPathCompletions(
 
   const { toRels: nodesToRelsSet, fromRels: nodesFromRelsSet } =
     getNodesFromRelsSet(dbSchema);
-  const assumedDirection = lastNode.leftArrow() ? 'outgoing' : 'incoming';
+  const assumedDirection = lastNode.leftArrow() ? 'left' : 'right';
```
anderson4j, line 47: So we follow the same naming convention as `getDirection(rel)`in `helpers.ts`

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,421 @@
...
+} from '../labelTreeRewriting.js';
+import { getNodesFromRelsSet } from '../labelTreeWalking.js';
+import { isLabelLeaf, LabelOrCondition, SymbolTable } from '../types.js';
+import { SyntaxDiagnostic } from './syntaxValidation.js';
+import { getDirection } from '../helpers.js';
+
+// A relationship has exactly one type; this stand-in represents "some
+// relationship type that is not mentioned anywhere" when checking whether a
+// relationship type expression is satisfiable at all.
+const NON_SCHEMA_RELTYPE = '__non_schema_reltype__';
```
anderson4j, line 25: Ex. `CREATE ()-[:IS_INVISIBLE]->()` would make a rel that doesnt show in the graphSchema. I suppose we could fetch even these with `db.relationshipTypes()` but they would all be treated the same (accepted for the case where the node has no label)

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+ * of labels that the schema allows on this side (`candidates`)?
+ *
+ * `evalNodeTree` only inspects labels that appear in the tree, so any candidate
+ * NOT named in the tree is interchangeable with every other such candidate. We
+ * therefore only enumerate subsets of the labels that are both named in the
+ * tree and valid candidates (`candInTree`). That keeps the 2^n search bounded by
+ * the number of labels written in the query's label expression, not by the
+ * (potentially schema-sized) candidate set.
+ *
+ * `otherAvailable` records that at least one valid candidate is not named in the
```
anderson4j, line 291: Quite a massive comment, but when I first read through I got a bit confused about this `otherAvailable` part, so I figured it could be helpful to include the bigger explanation

packages/language-support/src/tests/syntaxValidation/symbolValidation.test.ts
```diff
@@ -294,15 +321,254 @@ describe('Schema based linting spec', () => {
...
   });
 
-  test('Limitation: Bails on multiple labels on second variable in path segment (rel)', () => {
+  test('Handles multiple labels on second variable in path segment (rel)', () => {
     const query = 'MATCH (:Gym)<-[:WEAK_TO|IS_IN]-(:Pokemon) RETURN ""';
     const diagnostics = getDiagnosticsForQuery({ query, dbSchema });
+    expect(diagnostics).toEqual([
+      {
+        message:
+          'Relationship with relationship type(s) (WEAK_TO | IS_IN) has no outgoing connection to a node with label(s) Gym.',
```
anderson4j: Redo as `[:WEAK_TO | IS_IN] has no outgoing (:Gym)`

packages/language-support/src/tests/syntaxValidation/symbolValidation.test.ts
```diff
@@ -765,6 +1041,400 @@ describe('Schema based linting spec', () => {
     ]);
   });
 
+  test('Warns on a single (non-ANDed) negated label on the end node. Rel->Node', () => {
```
anderson4j: v1, v2 here?

packages/language-support/src/tests/syntaxValidation/symbolValidation.test.ts
```diff
@@ -765,6 +1041,400 @@ describe('Schema based linting spec', () => {
...
+    ]);
+  });
+
+  test('Does not warn on a single negated label when another target is viable. Rel->Node', () => {
+    const query = 'MATCH ()-[:CHALLENGES]->(:!Pokemon) RETURN ""';
+    const diagnostics = getDiagnosticsForQuery({ query, dbSchema });
+    expect(diagnostics).toEqual([]);
+  });
+
+  test('Warns on a single negated relationship type that removes the only valid type', () => {
```
anderson4j: maybe look into making name more similar to one above with !Label

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+  parseResult: ParsedStatement,
+  dbSchema: DbSchema,
+  symbolTable: SymbolTable,
+): SyntaxDiagnostic[] {
+  if (!dbSchema.graphSchema) {
+    return [];
+  }
+
+  const diagnostics: SyntaxDiagnostic[] = [];
+  try {
```
angrykoala: Is this try needed? Seems like it may hide implementation errors, as nothing here looks like would throw under normal execution

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+          : '';
+      }
+      const separator = tree.condition === 'and' ? ' & ' : ' | ';
+      return '(' + tree.children.map(renderLabelTree).join(separator) + ')';
+    }
+  }
+}
+
+/** All read-side `PatternElement`s, skipping CREATE/INSERT/MERGE subtrees. */
+function collectReadPatternElements(
```
angrykoala: This seems a bit ad-hoc way of collecting patterns, couldn't this be a collector or use a listener instead?

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+          const right = asNode(elements[i + 1]);
+          collectSegmentWarnings(
+            rel,
+            left,
+            right,
+            symbolTable,
+            schemaRelTypes,
+            fromRels,
+            toRels,
+          ).forEach((d) => diagnostics.push(d));
```
angrykoala: **nitpick** we could just push the collected array:

```js
const newWarnings=collectSegmentWarnings()
diagnostics.push(...newWarnings);
```

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  return diagnostics;
+}
+
+function asNode(
+  element: NodePatternContext | RelationshipPatternContext | undefined,
+): NodePatternContext | undefined {
+  return element instanceof NodePatternContext ? element : undefined;
+}
```
angrykoala: This way of filtering out relationshipPattern reads a bit awkward, the name `asNode` sounds to me like casting, more than checking the type

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+
+type Direction = 'right' | 'left' | 'undirected';
```
angrykoala, line 83: Could this type be used in `getDirection`?

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+    }
+    return subsetSatisfiable(cleaned, from) || subsetSatisfiable(cleaned, to);
+  });
+
+  if (supported) {
+    return undefined;
+  }
+
+  const nodeStr = renderLabelTree(nodeTree);
+  if (position === 'left') {
```
angrykoala, line 223: It looks like position is only used here. Couldn't the role be used instead?

Same question for qualifier

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+  // can be disproven against a label-keyed schema.
+  if (isAnyNode(cleaned) || isNotAnyNode(cleaned)) {
+    return undefined;
+  }
+
+  const supported = validRelTypes.some((relType) => {
+    const from = fromRels.get(relType) ?? new Set<string>();
+    const to = toRels.get(relType) ?? new Set<string>();
+    if (role === 'from') {
+      return subsetSatisfiable(cleaned, from);
```
angrykoala: This function may benefit for a longer name describing a bit more what it does?

Something like `isLabelTreeSatisfiable`? There is probably a better name than this though

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+    if (relTypeMatches(tree, t)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+function labelsIn(
+  tree: LabelOrCondition,
+  acc: Set<string> = new Set<string>(),
```
angrykoala, line 372: This parameter feels a bit awkward, as it is not really used when calling this function, just in the recursive part, which may be a bit confusing when trying to reuse this

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+ * is a valid candidate: the empty `candInTree` subset evaluates the tree as
+ * true, and `otherAvailable` confirms a real (non-empty) node can back it.
+ */
+function subsetSatisfiable(
+  tree: LabelOrCondition,
+  candidates: Set<string>,
+): boolean {
+  const mentioned = labelsIn(tree);
+  const candInTree = [...candidates].filter((c) => mentioned.has(c));
+  const otherAvailable = [...candidates].some((c) => !mentioned.has(c));
```
angrykoala: alternatively, `const otherAvailable=candidates.length > candInTree.length`. I'm not sure if this alternative is easier to read, but seems closer to the intention, plus we remove a loop

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+      return tree.children.some((c) => evalNodeTree(c, labelSet));
+    case 'not':
+      return !evalNodeTree(tree.children[0], labelSet);
+    case 'any':
+      return labelSet.size > 0;
+  }
+}
+
+/** Whether a relationship of exactly type `t` matches the type expression. */
+function relTypeMatches(tree: LabelOrCondition, t: string): boolean {
```
angrykoala: Not sure if it is worth it, but `evalNodeTree` could be reused for relationships

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+  const otherAvailable = [...candidates].some((c) => !mentioned.has(c));
+
+  const n = candInTree.length;
+  for (let mask = 0; mask < 1 << n; mask++) {
+    const labelSet = new Set<string>();
+    for (let i = 0; i < n; i++) {
+      if (mask & (1 << i)) {
+        labelSet.add(candInTree[i]);
+      }
+    }
```
angrykoala, line 322: I wonder if there is a way of making the intention of this bit clearer, took me a while to figure out what this is doing

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+    // a named candidate, or relies on some other (unnamed) candidate existing.
+    if (evalNodeTree(tree, labelSet) && (labelSet.size > 0 || otherAvailable)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+/** Evaluate a node label tree as a boolean over the node's label set `S`. */
+function evalNodeTree(tree: LabelOrCondition, labelSet: Set<string>): boolean {
```
angrykoala: Maybe there is a better name for this function? This name implies we are evaluating the node tree itself, when in reality is the label predicate

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+): Set<string> {
+  if (isLabelLeaf(tree)) {
+    acc.add(tree.value);
+  } else {
+    tree.children.forEach((c) => labelsIn(c, acc));
+  }
+  return acc;
+}
+
+//TODO: verify that this is how switch-case works (see 'and')
```
angrykoala: This looks fine, anything left to do in this comment?

angrykoala (commented):
A few minor stylistic nitpicks that are not too important and a few questions due to me trying to understand the project

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
anderson4j: Mm we could but im a bit unsure about what's nicest here. 

I went with this way because:
`'right' | 'left' | 'undirected'` in signature (instead of type `Direction`) makes it clear when we import getDirection what strings it actually returns (no need to click into import to find the type and its possible values)
I still kept the type here though, because repeating the possible values in `roleFor/qualifierFor` seemed a bit clunky.

Im open to changing it though, but curious to hear your opinion on this reasoning

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
anderson4j: nope, just forgot to remove the todo

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
anderson4j: Yep, although currently it will occur for each parse, also those done for syntax highlighting and autocompletions. That's true for the other collectors too though, and will be not be a problem once we fix caching of the parse.

I adjusted it. (Needs a bit different approach to ensure we're in a read clause though, but it will also centralize that logic (i used `couldCreateNewLabel` that we are fixing to handle `INSERT` etc). We should double check that we don't get issues with `SET` in this new usecase when rewriting it

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
anderson4j: Role depends on both position and direction. I figure this way makes it more clear than inverting that logic here (figuring out position from role+direction)

anderson4j: For qualifier it's just a helper to get the right string to put in the message
direction = right =>
[:B]->(:C)  "(:C) has no incoming [:B]"
(:A)-[:B] "[:B] has no incoming (:C)" (both become 'incoming')

and the opposite goes for direction = left

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
anderson4j: It's a pattern ive seen (and used) on the scala side (the server monorepo) to pass in an accumulating object and pass that on to recursive calls. If you don't object im happy to keep it here too

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
anderson4j: yeah still need to spread the array, so to me neither is nicer than the other. ill keep as is if you dont mind

packages/language-support/src/syntaxValidation/schemaBasedValidation.ts
anderson4j: Not so happy with any of the other solutions either, so I tried to make this a bit more readable with comment + renaming the subset.

(Found this type of solution was an option on the second result when googling "all subsets of subsets javascript" btw, so I don't think its that uncommon)

angrykoala (approved):

PR #703, opened 2026-07-02
Replace readme diagram with mermaid

angrykoala:
Closes LS-193


anderson4j (approved):
Showed to team during standup and they also approved

PR #706, opened 2026-07-13
Warn on missing property

angrykoala:
Adds warnings for property access on nodes and relationships when the property doesn't exist in the database.

To achieve this:
1. Add a new `propertiesCollector`, which collects properties information in the Cypher query, along with the variable name they belong to and whether they are used in a _write_ or _read_ clause
2.  In syntax validation, these properties are checked against the database, ignoring the following cases:
    * Properties that have been updated in the same query
    * Properties on variables that are not node or relationships


CONTRIBUTING.md
```diff
@@ -41,6 +41,7 @@ pnpm test
 End to end tests:
 
 ```
+cd packages/react-codemirror
```
nnecla: nit: this scopes `pnpm test:e2e` to only `react-codemirror` ignoring the `vscode-extension` e2e tests

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -267,10 +273,78 @@ function warnOnUndeclaredLabels(
...
+      if (!isNodeOrRelationshipProperty(propInCypher, symbolTable)) {
+        return false;
+      }
+
+      // Ignore all properties that are being modified in the query
+      if (writeProperties.has(propInCypher.propertyName)) {
+        return false;
+      }
+
+      return !propertiesInSchema.includes(propInCypher.propertyName);
```
nnecla, line 309: We normalize the label names [here](https://github.com/neo4j/cypher-language-support/blob/main/packages/language-support/src/syntaxValidation/syntaxValidation.ts#L37), so, a valid property with backticks still gives a warning

Image

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -267,10 +273,78 @@ function warnOnUndeclaredLabels(
...
+    },
+  );
+
+  for (const property of missingProperties) {
+    const message = property.propertyName + ' not available';
```
nnecla: nit. maybe we should check this message with Greg and keep it consistent with the other missing messages (aka `{property} is not present in the database. Make sure you didn't misspell...`)?

packages/language-support/src/syntaxValidation/syntaxValidation.ts
```diff
@@ -267,10 +273,78 @@ function warnOnUndeclaredLabels(
...
+  }
+  const symbol = symbolTable.find((symbol) => {
+    return symbol.variable === variable.name;
+  });
+  return symbol && symbolIsNodeOrRel(symbol);
+}
+
+/** Checks all possible types of a symbol are Node or Relationship */
+function symbolIsNodeOrRel(symbol: Symbol): boolean {
+  return symbol.types.every((type) => {
```
nnecla, line 349: Probably will never be the case 🙈 but maybe return false when types is empty?

angrykoala: oh yeah, good catch!

nnecla (approved):
LGTM!

PR #707, opened 2026-07-13
Add Claude skills

anderson4j:
Adds some skills that could help Claude (or other agents) to work in the repo. For now I've kept the skills in .claude as that's what ive used for now. We can adjust later for other agents.

finish-grammar-artifact-bump:
What it says, help fix an automated artifact bump PR. Common and repetitive job we do where solutions are often similar.

parse-cypher-with-antlr:
again what it says. An issue that came up in another claude session, where it didnt know how to best analyze the grammar of a query.

release-and-publish-flows:
Contained some handy info on how we handle releases. Many things like creating changeset and making the release are things I think we should do by-hand. Claude calls this an "ADVISORY skill" - so one to help a user, more than providing commands for claude to call.

run-cypher-language-support:
CLI way for an agent to E2E test language support features via language-server, formatter CLI and codemirror.

run-vscode-extension-tests:
Another one where we dont want the bot to run the tests without approval since they pop a window. Contains some handy info on how runs work

Also adds a general `AGENTS.md` with instructions on building, formatting code etc. 

Andres made a good point about the divide between README.md and AGENTS.md/SKILL.md:s . Maybe some things from the new MD files can be removed in lieu of better commands. Maybe agent MDs can be simplified by referencing the "human-oriented" readme. Could divide this per-subject (ex. antlr parsing) similar to skills to not clog context window. Let's discuss after your review @angrykoala 

Closes LS-163
Closes LS-172


.claude/skills/run-cypher-language-support/SKILL.md
anderson4j, line 1: This one kind of feels like we should just have tests to cover. Maybe a run of the tests with xvfb like in CI would be a workaround to not have the agent disturb user with pop-up?
Also another point of maintenance to make sure this still works with any new feature updates etc. 

I vote that we skip it, feel free to just skim through and see if you agree Andres

.claude/skills/run-vscode-extension-tests/SKILL.md
```diff
@@ -0,0 +1,37 @@
+---
```
anderson4j, line 1: This one has some handy information on how tests are run. Seems like info a person would also want to know

.claude/skills/run-vscode-extension-tests/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Environment gotchas
+
+- **Node must be < 24.16.0** (CI pins 24.15.0): Node 24.16 hangs playwright's browser-archive extraction and breaks the wdio webview tests. See the comment in `.github/actions/setup-and-build/action.yaml` and playwright issue #40724. If a suite hangs immediately on browser/VS Code setup, check `node --version` first.
+- **Two module graphs** (the big one, also in AGENTS.md): the extension under test is the bundled `dist/extension.js`, while spec files compile separately from `src/`. Anything invoked via `commands.executeCommand(...)` runs the **real bundled handlers** — sinon stubs applied to `src/` modules don't reach them, so those paths do real npm fetches, real downloads, real language-server calls. To test with stubs applied, import and call the `src/` function directly.
+- **Linter download cache**: downloaded lint workers live in the real VS Code global storage — `%APPDATA%/Code/User/globalStorage/neo4j-extensions.neo4j-for-vscode` on Windows (`tests/helpers.ts` `getExtensionStoragePath()`) — NOT in `.vscode-test/user-data`. The suite's user-data wipe therefore does not give you cold-download behavior; wipe the globalStorage path for that.
```
anderson4j, line 27: Something I didnt think about before, but we actually delete the same globalstorage cache used by the actual extension here. So if you have a bunch of old linters and run the tests, you will need to redownload after. Shouldnt really cost too much data/time, but maybe worth looking into a better way to do it (not pressing imo)

packages/vscode-extension/syntaxes/cypher.json
anderson4j, line 41: we forgot to build this on "update dev artifacts" PR. We should be able to generate it in the workflow, but I've accidentally broken that in [this fix](https://github.com/neo4j/cypher-language-support/pull/650) in response to the workflow crashing on failed build.

AGENTS.md
```diff
@@ -0,0 +1,68 @@
+# AGENTS.md
+
+This file provides guidance to coding agents (Claude Code, Codex, Cursor, …) when working with code in this repository.
+
+## Repository overview
+
+pnpm monorepo (workspaces: `packages/*` and `vendor/*`) implementing Neo4j's Cypher language support: syntax highlighting, autocompletion, linting, and formatting. Requires Node >= 24.11.1 (engineStrict). Generating the parser additionally requires `antlr4-tools` (`pip install antlr4-tools`) and Java.
```
anderson4j: Should probably ref `.nvmrc` for node version to avoid mismatch.

.claude/skills/release-and-publish-flows/SKILL.md
```diff
@@ -0,0 +1,40 @@
+---
```
anderson4j, line 1: Not sure how much value this gives compared to a normal readme. Maybe the added commands help agents a bit

.claude/skills/parse-cypher-with-antlr/SKILL.md
```diff
@@ -0,0 +1,44 @@
+---
+name: parse-cypher-with-antlr
+description: Inspect how a Cypher query is parsed by the ANTLR grammar — print the parse tree or token stream for any query straight from the .g4 grammar files, without building the project. Use when debugging why a query parses a certain way, investigating grammar/lexer behavior, checking which rule or token matches some syntax, or after a grammar bump.
+---
+
+# Parse Cypher queries with ANTLR
```
anderson4j, line 6: This one I noticed claude had an issue with before, so I think a skill describing the commands is handy. (Also for a person)

.claude/skills/finish-grammar-artifact-bump/SKILL.md
```diff
@@ -0,0 +1,39 @@
+---
+name: finish-grammar-artifact-bump
+description: Reconcile a Cypher grammar / semantic-analysis artifact bump — the monthly "Automated update of artifacts" bot PR or any change to Cypher25*.g4 / semanticAnalysis.js. Covers regenerating and committing cypher.json, categorizing new lexer tokens, fixing completion/highlighting/formatting wiring, and updating tests. Use when working on an artifact-bump PR, when the build breaks after grammar files changed, or when CI fails the "TextMate grammar is up to date" check.
+---
+
+# Finish a grammar/artifact bump
```
anderson4j, line 6: We have a card to trigger an agent job to autofix these. I think this skill gives some good info on how that would look.

anderson4j (commented):
Left some comments, but awaiting your review @angrykoala before further changes

.claude/skills/parse-cypher-with-antlr/SKILL.md
```diff
@@ -0,0 +1,43 @@
...
+
+`antlr4-parse` (from `antlr4-tools`, same prerequisite as the build — see AGENTS.md) interprets the grammar files in `packages/language-support/src/antlr-grammar/` directly. No build step needed, and it reflects grammar edits instantly. The generated TypeScript parser in `src/generated-parser/` is produced from these same grammars, so the trees match what the library sees.
+
+## The commands
+
+**Must run from the grammar directory** — the grammars use `import CypherPreLexer;` etc., and `antlr4-parse` resolves those imports relative to the current working directory, not the grammar file's location. Running from the repo root with full grammar paths fails with `error(110): can't find or load grammar CypherPreLexer`. From the repo root, use a subshell:
+
+```sh
+# Parse tree (LISP notation), query piped on stdin — works fine for an agent, no interactive input needed
+(cd packages/language-support/src/antlr-grammar && echo 'MATCH (n:Person) RETURN n' | antlr4-parse ./CypherCmdLexer.g4 ./CypherCmdParser.g4 statementsOrCommands -tree)
```
angrykoala, line 16: We may want to consider having these common tools as scripts to make them easier to run

AGENTS.md
angrykoala: Maybe consider using the package.json engine as a source of truth instead? (which is probably where this data comes from in the first place, judging from engineStrict)

CLAUDE.md
```diff
@@ -0,0 +1 @@
+@AGENTS.md
```
angrykoala, line 1: This looks good to me, but doesn't clause automatically checks the `AGENTS.md` file? Or is there more advantages to explicitly define the `CLAUSE.md`?

.claude/skills/run-cypher-language-support/playground-screenshot.png
angrykoala, line 1: I'm not sure I understand why do we need this screenshot for

CLAUDE.md
anderson4j: I thought this was the way to do it (what claude proposed when asked) - but looking [here](https://dev.to/deployhq/claudemd-agentsmd-and-every-ai-config-file-explained-4pde) it looks like we can remove the file entirely

.claude/skills/run-cypher-language-support/playground-screenshot.png
anderson4j: Yeah I would remove it from git if we keep the workflow. Ive proposed we skip this skill and look into making the e2e tests easier to run without pop-up though, so in that case we would remove the entire folder. 

If you like this approach however (no container-manager needed is nice) we can un-check out the generated image.

angrykoala (approved):

CLAUDE.md
anderson4j: Then again, [these docs](https://code.claude.com/docs/en/memory.md) from the claude code docs disagree, so ill stick to this way

.claude/skills/parse-cypher-with-antlr/SKILL.md
anderson4j: Lets look at if we want the script tooling you mentioned in pairing in a separate PR, if so maybe we shift some of this there

PR #710, opened 2026-07-17
Relint codemirror on schema update

anderson4j:
Adds a "needsRefresh" check to react-codemirror linter, that gets triggered when schema is updated. (Checks by object equality, not deep equality)

Has a corresponding PR [in upx](https://github.com/neo4j/upx/pull/12023) that will update codemirror to use this change, and prevent the passing of a new schema object when the old schema is identical to the new

Closes LS-203

### DIFF
--- a/packages/react-codemirror/src/CypherEditor.tsx
+++ b/packages/react-codemirror/src/CypherEditor.tsx
@@ -38,7 +38,7 @@ import {
   replMode as historyNavigation,
 } from './historyNavigation';
 import { cypher, CypherConfig } from './lang-cypher/langCypher';
-import { cleanupWorkers } from './lang-cypher/syntaxValidation';
+import { cleanupWorkers, schemaUpdated } from './lang-cypher/syntaxValidation';
 import { basicNeo4jSetup } from './neo4jSetup';
 import { getThemeExtension } from './themes';
 import { richClipboardCopier } from './richClipboardCopier';
@@ -134,6 +134,10 @@ export interface CypherEditorProps {
   };
   /**
    * The schema to use for autocompletion and linting.
+   *
+   * Compared by reference to decide when the document needs relinting:
+   * Avoid passing a new object when nothing changed
+   * as that would relint on every poll for no reason.
    *
    * @type {DbSchema}
    */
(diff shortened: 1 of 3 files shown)


packages/react-codemirror/src/lang-cypher/syntaxValidation.ts
```diff
@@ -14,62 +14,74 @@ const pool = workerpool.pool(WorkerURL, {
...
-    return [];
-  });
+      return [];
+    },
+    {
+      needsRefresh: (update) =>
+        update.transactions.some((tr) =>
+          tr.effects.some((effect) => effect.is(schemaUpdated)),
+        ),
+    },
```
anderson4j, line 83: this is the new check added to the existing cypherLinter extension of the codemirror to retrigger lint on schema update.

packages/react-codemirror/src/lang-cypher/syntaxValidation.ts
```diff
@@ -14,62 +14,74 @@ const pool = workerpool.pool(WorkerURL, {
   workerTerminateTimeout: 2000,
 });
 
-export const cypherLinter: (config: CypherConfig) => Extension = (config) =>
```
anderson4j, line 17: Lots of "diff" inside `cypherLinter`, but actually the second argument commented on below is the only change, the rest is formatting

packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -135,6 +135,10 @@ export interface CypherEditorProps {
   /**
    * The schema to use for autocompletion and linting.
    *
+   * Compared by reference to decide when the document needs relinting:
+   * Avoid passing a new object when nothing changed
+   * as that would relint on every poll for no reason.
+   *
    * @type {DbSchema}
    */
```
OskarDamkjaer, line 143: The way I've seen this handled in the past, is that you add a version number to the schema. So you say `version: 1`. Then when you've fetched the server, you do the deep compare once, if there's any difference, you increment the verion number. This means there's no risk or re-running more often than our polling interval.

Everything downstream just checks what the last schema version used was

PR #714, opened 2026-07-23
normalize CRLF line endings

nnecla:
### Description

Codemirror expects LF line endings and collapses `\r\n` into a single break, so a value containing CRLF is longer than the resulting document. When the editor was auto-focused, we set the cursor to `value.length`, which points past the end of the  normalised doc(shorter) and threw selection points outside of document.

This PR is to:
- Clamp the cursor position to the doc length
- Apply `normalizeLineEndings` consistently and not just on `setValueAndFocus`

**Similar to:** https://github.com/neo4j/cypher-language-support/pull/539

### DIFF
--- a/packages/react-codemirror/src/CypherEditor.tsx
+++ b/packages/react-codemirror/src/CypherEditor.tsx
@@ -46,6 +46,19 @@ import { LintWorker } from '@neo4j-cypher/lint-worker';
 import workerpool from 'workerpool';
 
 type DomEventHandlers = Parameters<typeof EditorView.domEventHandlers>[0];
+
+/**
+ * Normalize CRLF line endings to LF, which is what CodeMirror expects.
+ * CodeMirror collapses `\r\n` into a single line break, so a raw `value`
+ * string with CRLF endings is longer than the resulting document. Normalizing
+ * up front keeps the document length in sync with the value length used for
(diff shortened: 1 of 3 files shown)


packages/react-codemirror/src/CypherEditor.tsx
```diff
@@ -46,6 +46,19 @@ import { LintWorker } from '@neo4j-cypher/lint-worker';
...
+ * Normalize CRLF line endings to LF, which is what CodeMirror expects.
+ * CodeMirror collapses `\r\n` into a single line break, so a raw `value`
+ * string with CRLF endings is longer than the resulting document. Normalizing
+ * up front keeps the document length in sync with the value length used for
+ * cursor placement, and avoids "Selection points outside of document" errors.
+ * https://codemirror.net/docs/ref/#state.EditorState^lineSeparator
+ */
+function normalizeLineEndings<T extends string | undefined>(value: T): T {
+  return value?.replace(/\r\n/g, '\n') as T;
+}
```
angrykoala: Is there a case in which we want to return `undefined` from this function?

Maybe this one could just accept and return a string, and use `?? ''` in the input, rather than the output

angrykoala (approved):

PR #720, opened 2026-08-03
Add Knip

angrykoala:
Close LS-177


package.json
```diff
@@ -27,7 +27,8 @@
     "lint-fix": "pnpm lint --fix",
     "format": "oxfmt",
     "format:check": "oxfmt --check",
-    "benchmark": "pnpm --recursive benchmark --concurrency=1"
+    "benchmark": "pnpm --recursive benchmark --concurrency=1",
+    "knip": "knip"
```
OskarDamkjaer: I think `pnpm` let you run `pnpm knip` without adding this script. @claude could you double check the docs

OskarDamkjaer: I just double checked myself, it works without this script, it's an `npm` thing to require it. Feel free to leave it in if you prefer it this way

knip.json
```diff
@@ -0,0 +1,63 @@
...
+        "@wdio/spec-reporter",
+        "ovsx",
+        "wdio-wait-for",
+        "expect-webdriverio"
+      ]
+    }
+  },
+  "ignore": ["vendor/**/*"],
+  "ignoreBinaries": ["package", "dev", "vsce", "ovsx"],
+  "ignoreDependencies": ["ts-node"]
```
OskarDamkjaer, line 62: I don't think we need `ts-node` anymore, as far as I can see we use the node native way `node --experimental-strip-types` everywhere instead

.github/workflows/ci.yaml
```diff
@@ -51,6 +51,9 @@ jobs:
       - name: Check formatting
         run: pnpm format:check
 
+      - name: knip
```
OskarDamkjaer: nit:  
```suggestion
      - name: Detect unused exports and dependencies
```

knip.json
angrykoala: This pipeline fails: https://github.com/neo4j/cypher-language-support/actions/runs/31587828387/job/94086303701?pr=720

Seems to be a peer dependency that will need some changes before getting rid of it

OskarDamkjaer (approved):

PR #722, opened 2026-08-10
Functions and Procedures info on hover

angrykoala:
This PR adds functions and procedures information on hover

Close LS-132


packages/language-support/src/hover.ts
```diff
@@ -0,0 +1,83 @@
...
+export function getHoverInfo({
+  caretPosition,
+  dbSchema,
+  parsingResult,
+}: {
+  caretPosition: number;
+  dbSchema: DbSchema;
+  parsingResult: ParsingResult;
+}): HoverInfo | undefined {
+  const result = findCaret(parsingResult, caretPosition);
```
anderson4j: result = findCaret(parsingResult,...) sounds a bit confusing to me
At other findCaret calls we use
caret = findCaret(...)
Makes more sense from what the method name is, but we dont really get the caret from findCaret imo (rather the caret position is an input)

I propose we rewrite both calls and method name like
result = findCaret -> caretResult = findResultAtCaret
caret= findCaret -> caretResult = findResultAtCaret

anderson4j: Actually, I think it would be nice to merge a lot of the logic we have here with that of `getSignatureInfo`. Should be doable to call that method and `createHoverInfoObject `from the result, with some small modification to `getSignatureInfo`

(That would remove our call of findCaret, but I still think the namechange in other places is worthwhile)

packages/language-support/src/types.ts
```diff
@@ -90,3 +90,23 @@ export type SymbolsInfo = {
   query: string;
   symbolTables: SymbolTable[];
 };
+
+export type HoverInfo = {
```
anderson4j: This name sounds like it's generic hover info. I think SignatureHoverInfo is more accurate

packages/language-support/src/types.ts
```diff
@@ -90,3 +90,23 @@ export type SymbolsInfo = {
...
+  signature: string;
+  description: string;
+  returnDescription:
+    | string
+    | Array<{
+        name: string;
+        description: string;
+        isDeprecated: boolean;
+        type: string;
```
anderson4j: This is the existing type `ReturnDescription`

packages/language-support/src/types.ts
```diff
@@ -90,3 +90,23 @@ export type SymbolsInfo = {
...
+      }>;
+  isDeprecated: boolean;
+  params: Array<{
+    name: string;
+    description: string;
+    isDeprecated: boolean;
+    type: string;
```
anderson4j: This is also `ReturnDescription` although maybe the `ArgumentDescription` type has a more fitting name (it has an optional extra field, but otherwise its the same)

packages/vscode-extension/tests/specs/api/hover/hoverFunctions.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { testData } from '@neo4j-cypher/language-support';
+import * as assert from 'assert';
+import * as vscode from 'vscode';
+import { eventually, newUntitledFileWithContent } from '../../helpers';
+
+suite('Hover spec', () => {
+  test('Hover works for functions', async () => {
```
anderson4j, line 20: Nice! Think it would be good to expand tests to show what happens when hovering on incomplete function/procedure calls, for example without parenthesis and when missing parameters. Especially the "incomplete" case - we already get the errors on hover and the info overlaps - do we want to avoid providing the new hoverinfo in this case?

packages/language-support/src/hover.ts
```diff
@@ -0,0 +1,83 @@
...
+    return undefined;
+  }
+
+  const methodName = method.methodName;
+
+  if (method.methodType === 'procedure') {
+    if (!dbSchema.procedures) {
+      return undefined;
+    }
+    const fn = dbSchema.procedures?.[cypherVersion]?.[methodName];
```
anderson4j: If we decide not to split - a reminder that these names could be better (ex. fn => function in function case and fn => procedure in procedure case)

anderson4j (commented):
Good stuff! Had some suggestions for improvement

anderson4j:
Oh also, it needs a changeset

packages/language-support/src/types.ts
angrykoala: Good point

packages/language-support/src/types.ts
angrykoala: ArgumentDescription seems like a better fit, even if the types are the same, conceptually the return is different from parameters and I'd rather keep the separate

packages/language-support/src/signatureHelp.ts
```diff
@@ -69,11 +69,11 @@ export function toSignatureInformation(
...
 
 function toSignatureHelp(
-  methodSignatures: Record<string, Neo4jFunction | Neo4jProcedure> = {},
+  method: Neo4jFunction | Neo4jProcedure | undefined,
   parsedMethod: ParsedMethod,
 ): SignatureHelp {
-  const methodName = parsedMethod.methodName;
-  const method = methodSignatures[methodName];
+  // const methodName = parsedMethod.methodName;
+  // const method = methodSignatures[methodName];
```
anderson4j: Forgotten comment-instead-of-delete

packages/language-support/src/signatureHelp.ts
```diff
@@ -172,23 +172,27 @@ export class SignatureHelper extends CypherCmdParserListener {
...
-    consoleCommandsEnabled?: boolean;
-  } = {},
-): SignatureHelp {
+export function getMethodSignature({
+  query,
+  dbSchema,
+  caretPosition,
+  consoleCommandsEnabled = true,
+}: {
+  query: string | ParsingResult;
```
anderson4j: assigning value of type `ParsingResult` to variable `query` looks off to me. Think its better we just take 
`parsingResult: ParsingResult` and avoid resolving below + switch `getSignatureInfo` (which is the only call where we have "query or parsingresult") to pass:
`parsingResult: parsingResult ?? createParsingResult(query, { consoleCommandsEnabled }),`

Then we also dont have to pass `consoleCommandsEnabled` to `getMethodSignature`

Also, is there any point to wrapping the inputs in an object here? In cases like `getSignatureInfo` I know it lets us call the function with any combination of the optional inputs, but here we only have the one (and after my proposed change, none).

packages/language-support/src/signatureHelp.ts
```diff
@@ -211,25 +215,56 @@ export function getSignatureInfo(
...
-            dbSchema.procedures?.[cypherVersion] ?? {},
-            method,
-          );
-        }
+      let signatures: Record<string, Neo4jFunction | Neo4jProcedure> = {};
+      if (method.methodType === MethodType.function) {
+        signatures = dbSchema.functions?.[cypherVersion] ?? {};
+      } else {
+        signatures = dbSchema.procedures?.[cypherVersion] ?? {};
       }
```
anderson4j: These arent signatures, theyre the entire Neo4jFunction | Neo4jProcedure objects. Think method/function is more accurate. Id go with neo4jMethods or maybe schemaMethods if you want to distinguish from the parsed method

packages/language-support/src/signatureHelp.ts
```diff
@@ -211,25 +215,56 @@ export function getSignatureInfo(
...
+      const methodName = method.methodName;
+
+      return {
+        parsedMethod: signatureHelper.result,
```
anderson4j: this is the `method` variable above. I'd even think it's nice if we just rename it as `parsedMethod` from the get go

anderson4j: Actually, the Neo4jFunction | Neo4jProcedure already has name. All we need from parsedMethod is `activeParameter` so lets just pass that

packages/language-support/src/hover.ts
```diff
@@ -13,59 +11,29 @@ export function getHoverInfo({
...
-    const fn = dbSchema.functions?.[cypherVersion]?.[methodName];
-    if (!fn) {
-      return undefined;
-    }
-
-    const isDeprecated = fn.isDeprecated;
-
-    return createHoverInfoObject(fn, isDeprecated);
-  }
+  return createHoverInfoObject(signature, false);
```
anderson4j: signature is not actually signature (as mentioned in another comment, it's (Neo4jFunction | Neo4jProcedure)

These actually contain the true value for "isDeprecated" - while this row incorrectly passes false every time.

packages/language-server/src/server.ts
```diff
@@ -238,6 +246,37 @@ connection.onNotification(
...
+    caretPosition: offset,
+    dbSchema: neo4jSchemaPoller.metadata?.dbSchema ?? {},
+  });
+
+  if (!hoverInfo) {
+    return null;
+  }
+
+  return {
+    contents: {
```
anderson4j, line 264: We store "isDeprecated" in `SignatureHoverInfo` but dont actually use it anywhere. Maybe we should add it here?
In vscode the typescript lang-supp does it with a `@deprecated` tag
Image

Almost feels a bit over-the-top with linting also showing deprecations, but I suppose at times (like in firefox, or for very big queries) linting is not available, so imo there's still a small benefit to showing it here. 

(If we decide against this, we should remove `isDeprecated` from `SignatureHoverInfo`, as it is unused right now)

anderson4j: We store "isDeprecated" in `SignatureHoverInfo` but dont actually use it anywhere. Maybe we should add it here?
In vscode the typescript lang-supp does it with a `@deprecated`
Image

Almost feels a bit over-the-top with linting also showing deprecations, but I suppose at times (like in firefox, or for very big queries) linting is not available, so imo there's still a small benefit to showing it here. 

(If we decide against this, we should at least remove isDeprecated, as it is unused right now)

anderson4j (changes requested):
I think you forgot the request for more tests in my last review. Also, the tests you made use a new schema. I think it would be good to switch so they use `testData.ts`'s `testData.mockSchema`. That one has a lot of functions/procedures that exist on neo4j.

Also while playing around with some different functions, I noticed we have certain functions defined in the grammar, and these lack proper handling when the parser "properly" parses them (often it starts of parsing as `functionInvocation` for these, which we handle, but in a complete clause they get their own special rule. Anywho, I made a [branch](https://github.com/neo4j/cypher-language-support/pull/740) off from this one with a fix

anderson4j:
Oh and one more thing I noticed! We show the function hover info on the entire functionInvocation, but other language-support (ex. for JS in vscode, or Java in Intellij) only provides the function info when hovering the name. (Hovering the params section would instead give info on the param, which we will get even more of soon with hover for variables)

hovering blank space inside function ex.  `MATCH (n) RETURN abs(       -1)` would then give no info 

Something else worth adding tests for - Hovering function in param of function, ex.  `MATCH (n) RETURN abs(count(n))` (where we hover count),

packages/vscode-extension/tests/specs/api/hover/hoverFunctions.spec.ts
angrykoala: mmm, you mean the errors (e.g "insufficient parameters")? That one is not a problem, will add a test for that case

The one that overlaps, that we may want to tweak, is the autocomplete information

packages/language-support/src/hover.ts
angrykoala: No longer relevant after unifying signatureHelper

packages/language-support/src/hover.ts
angrykoala: Not quite, we should discuss this point, as isDeprecated is only available in Neo4jFunction, ideally we may want to change how this schema methods are stored

packages/language-support/src/signatureHelp.ts
angrykoala: We also need the methodType, I'll review if we can get rid of the name

packages/language-support/src/signatureHelp.ts
```diff
@@ -252,19 +244,23 @@ export function getSignatureInfo(
     consoleCommandsEnabled?: boolean;
   } = {},
 ): SignatureHelp {
+  const resolvedParsingResult = parsingResult
+    ? parsingResult
+    : createParsingResult(query, { consoleCommandsEnabled });
+
+  // Always pass parsingResult
```
anderson4j: Think you forgot to delete

packages/vscode-extension/tests/specs/api/hover/hoverFunctions.spec.ts
```diff
@@ -4,10 +4,77 @@ import * as vscode from 'vscode';
...
+  });
+
+  test('Hover works for incomplete functions parameters along with signature help', async () => {
+    const document = await newUntitledFileWithContent('RETURN abs(');
+    const position = new vscode.Position(0, 8);
+
+    await eventually(async () => {
+      const signatureHelp: vscode.SignatureHelp =
+        await vscode.commands.executeCommand(
+          'vscode.executeSignatureHelpProvider',
```
anderson4j: Here you have `executeSignatureHelpProvider`

Should actually be important for the test to be valid to use `executeHoverProvider` like above, otherwise we don't actually check the hover->trigger signatureInfo part.

packages/vscode-extension/tests/specs/api/hover/hoverFunctions.spec.ts
```diff
@@ -4,10 +4,77 @@ import * as vscode from 'vscode';
 import { eventually, newUntitledFileWithContent } from '../../helpers';
 
 suite('Hover spec', () => {
+  const absFunction = testData.mockSchema.functions['CYPHER 5']['abs'];
+  const expectedHoverInfo = [
```
anderson4j, line 8: I didnt think of it before, but would be good to also include procedures in these integration tests.
(so maybe rename this expectedFunctionHoverInfo, and add equiv tests for procedures)

anderson4j (commented):
Some final minor things

packages/vscode-extension/tests/specs/api/hover/hoverFunctions.spec.ts
angrykoala: This is actually testing that both, the signature and hover work at the same time, the test case use `executeHoverProvider` a bit below, I'm not 100% sure if this is adding that much coverage, as signature should be already tested elsewhere, so I'm happy to remove this check if it doesn't add that much value

packages/vscode-extension/tests/specs/api/hover/hoverFunctions.spec.ts
angrykoala: :thinking:  I'm not sure, extension tests are slow and bit fragile, and it is only testing the integration with the language server, so it may not really add much, as procedures are tested there

packages/vscode-extension/tests/specs/api/hover/hoverFunctions.spec.ts
anderson4j: Yeah lets remove it then

anderson4j (approved):

PR #730, opened 2026-08-21
Finish antlr ng migration

OskarDamkjaer:
Below is details description of what's needed to change and why, in removal of our vendored dependency. It's quite a large change, and I've done many review loops on my own (manual and ai assisted) to get a clean, correct and minimal changset.

I know the PR description below is very long, i've iterated on it to make sure it covers all the changes along with all the rationale, so it should be worth the read. Resolves LS-68.

------------------------------------------------------------------------

# antlr-ng migration — what changed and why

**Background:** this branch (`finish-antlr-ng-migration` vs `main`) swaps the parser toolchain behind Neo4j's Cypher language support — Java-based ANTLR codegen + the old `antlr4` runtime → `antlr-ng` + `antlr4ng` — and replaces the vendored antlr4-c3 completion fork with the published upstream package (history in 1.3). ~400 lines of product code changed, ~7k lines of vendor deleted; unit tests, the formatting corpus, and a clean build are green. Below: every non-self-explanatory change, grouped by kind, with the alternative that was (or could still be) considered.

**Why it's worth doing:** the old `antlr4` JS runtime is effectively dormant — its last publish (4.13.2) was Aug 2024, and it lags mainline ANTLR: it never implemented `exitRule`'s EOF handling (see 2.2). [antlr4ng](https://github.com/mike-lischke/antlr4ng) is a TypeScript-native rewrite that per its readme "implements everything the Java runtime does", developed as one coherent stack with antlr-ng and antlr4-c3 (releases into 2025). Its stricter, honest types are what let us delete the `@ts-ignore` default-export hack and the `isErrorNode()` workarounds (see 3.5). On performance we did not measure our own workload; upstream's [public benchmark suite](https://github.com/mike-lischke/antlr4ng#tests-and-benchmarks) (MySQL grammar, monitored for regressions) shows antlr4ng beating the old antlr4 JS runtime by 9–35% — a likely indicator the switch is not a regression for us.

**Contents**

1.  [The big moves](#1-the-big-moves) — codegen swap, runtime swap, unvendoring antlr4-c3
2.  [Semantic adaptations](#2-semantic-adaptations-where-behavior-could-have-changed) — places where behavior could have changed
3.  [Mechanical renames](#3-mechanical-renames-no-behavior-change) — safe API idioms, no behavior change
4.  [Build, CI & deleted files](#4-build-ci--deleted-files) — incl. the nodenext switch (4.3)

### 1.1 Parser codegen: antlr4-tools (Java jar) → antlr-ng

**What:** `gen-parser` no longer downloads/runs the Java-based ANTLR 4.13.2 jar via `antlr4-tools`; it uses `antlr-ng`, a TypeScript-native port of the ANTLR generator. The CI step that curled `antlr-4.13.0-complete.jar` is deleted; Java and the `antlr4-tools` pip package are gone from CONTRIBUTING.md prerequisites.

**Why:** no Java/Python prerequisite for contributors or CI; the generator and the runtime (below) come from the same author and move together.

**Alternative:** stay on antlr4-tools + the official Java generator.

Zero migration risk, but it chains the repo to the old `antlr4` JS runtime (which upstream barely maintains) and — critically — blocks unvendoring antlr4-c3, since upstream antlr4-c3 3.x is built on antlr4ng. Keeping the old toolchain means keeping the fork forever.

packages/language-support/package.json · package.json · .github/actions/setup-and-build/action.yaml · CONTRIBUTING.md · AGENTS.md

### 1.2 Runtime: `antlr4@4.13.2` → `antlr4ng@3.0.16`

**What:** the parser runtime the generated code runs against. Nearly every other change in this document is a consequence of API differences between these two runtimes.

**Why:** antlr4ng is the runtime antlr-ng generates code for; it is actively maintained and is what upstream antlr4-c3 targets. Exact-pinned (`3.0.16`, not a range).

**Alternative:** none that keeps 1.1 and 1.3 — the three moves are one decision. The real alternative was "don't migrate", i.e. keep the old runtime + Java codegen + vendored fork as a set.

packages/language-support/package.json

### 1.3 Unvendor antlr4-c3: fork in vendor/ → published `antlr4-c3@3.4.4`

**What:** the entire `vendor/antlr4-c3` directory (~7k lines, including the fork's patch, upstream's test suite, and grammar fixtures) is deleted; the completion engine is now a normal dependency. The build no longer pre-builds the vendor package or copies its dist into `dist/esm/vendor/`.

**Why it can go now:** the fork addressed compatibility with the old `antlr4` runtime and later gained an optional follow-up completion feature. Upstream antlr4-c3 now targets antlr4ng directly, removing the original reason to vendor it.

**What the fork added:** it computed follow-up token chains for every completion candidate and marked them *optional*, letting the completer offer both `style` and `style reset` at `:`. Upstream only attaches follow-ups when the caret's follow set is a single token (`hasTokenSequence`), so with 14 console-command candidates at `:`, `style` now gets an empty follow-up list.

**Measured impact:** the only product-facing loss in the suite is the combined `style reset` entry. The command still parses, highlights, and completes in two steps (`:style ` → `reset`). A probe over ~7,300 completion positions found other merged keyword pairs, but their bare keyword remains offered and valid; no legal continuation disappears.

**Decision (oskar):** drop the extra feature — two-step completion is the accepted behavior going forward, not an open question. The alternatives were considered and set aside:

- **Restore without a fork:** subclass `CodeCompletionCore` and override `getFollowingTokens` (~40 lines against the antlr4ng ATN) plus an "optional" flag in `completionCoreCompletions.ts`. Feasible, but re-commits us to maintaining fork-only behavior indefinitely.
- **Keep the fork:** re-port the patch onto antlr4-c3 3.4.4 and keep vendoring. Highest fidelity, highest maintenance — defeats the purpose of the migration.
- **Propose upstream:** PR the optionals feature to antlr4-c3. Clean end state, slowest, not in our control.

vendor/antlr4-c3/\*\* (deleted) · completionCoreCompletions.ts · package.json scripts · pnpm-workspace.yaml · vitest.workspace.ts · knip.json

### 1.4 Dependency hygiene: one runtime copy only

**What:** `antlr4-c3@3.4.4` exact-pins `antlr4ng@3.0.16` — the same copy language-support depends on, so there is no dual-runtime `instanceof` hazard. The `antlr4ng@3.0.15` in the lockfile comes from `stringtemplate4ts` (used by antlr-ng at codegen time only, never at runtime). `antlr4@4.13.2` survives only via lint-worker's deliberate pin of old published language-support versions (needed to lint against older Neo4j versions).

**Alternative:** none needed — but worth guarding: a renovate rule now disables automatic bumps of `antlr4ng`/`antlr4-c3`/`antlr-ng`, since antlr4ng must move in lockstep with antlr4-c3's internal pin (two runtime copies would break `instanceof` checks everywhere).

pnpm-lock.yaml · packages/language-support/package.json · renovate.json5

### 2.1 ErrorTrackingStrategy replaces `ctx.exception`

**What changed:** the old runtime attached each parser exception to the rule context that caught it:

    catch (re) {
        if (re instanceof RecognitionException) {
            _localctx.exception = re;
            this._errHandler.reportError(this, re);
            this._errHandler.recover(this, re);
        } else { throw re; }
    }

antlr4ng removed `ctx.exception`. Recording in `SyntaxErrorsListener` is not equivalent: `DefaultErrorStrategy` suppresses listener notifications while recovering, whereas generated catch blocks always call `reportError`. `ErrorTrackingStrategy` therefore overrides that public hook, records `recognizer.context → exception`, then delegates to the default strategy. Recovery behavior remains inherited and unchanged. Cypher has no left-recursive generated rules, so `recognizer.context` is the context whose catch block fired.

**Who consumes it:** four old sites, in two shapes:

- **"Was this pattern element valid?"** — `child.exception === null` → `!errorTracker.hasError(child)`: `schemaBasedCompletions.ts` (×2 — finding the last *valid* node/relationship pattern so label/rel-type completion still works after a broken segment) and the snippet path in `completionCoreCompletions.ts`. Purely boolean checks.
- **"Where did parsing break?"** — formatting's `getParseTreeAndTokens` used `tree.exception.offendingToken` to locate unparsable input and format around it; now `errorTracker.offendingTokenAt(tree)`.

The tracker is fresh per statement parse and is exposed on `ParsedStatement.errorTracker`; formatting creates its own.

**Nuances & impact:**

- **First vs. last:** the old runtime *overwrote* `_localctx.exception` on each catch, so `tree.exception` was the *last* exception a context caught; our Map is insert-if-absent, so we keep the *first*. Irrelevant for the boolean checks; for formatting's offending-token lookup, "first error on the root context" is arguably the better answer to "where does the broken region start" — and the 11,933-query formatting corpus shows no divergence either way.
- **Cost and API:** one Map insert per caught error; `ParsedStatement` gains a required field, but no downstream code constructs it manually.

**Alternatives considered:**

- **Post-hoc tree walk for ErrorNode children** — different semantics: an error node under a context is not the same as "this rule's catch fired", and it would force every consumer to re-walk.
- **Match old last-wins semantics** — a one-word change (drop the insert-if-absent guard) if we ever want bit-exact fidelity; first-wins reads better for the formatting use case.

errorTrackingStrategy.ts (new) · cypherLanguageService.ts · formattingHelpers.ts · schemaBasedCompletions.ts · completionCoreCompletions.ts

### 2.2 parseToCommand: EOF stop via CommonTokenStream.LB

**Before this PR:** `parseToCommand` extracts each statement's original text with `inputStream.getTextFromRange(stmtStart.start, stop.stop)`, where `stop` came from the root rule context. The old `antlr4@4.13.2` JS runtime always assigned `LT(-1)` when leaving a rule, so after matching EOF the context happened to stop on the last real default-channel token. The extraction code relied on that behavior without needing to say so.

**After this PR:** antlr4ng follows mainline ANTLR and records the matched EOF token as the root context's stop. After the rule returns, the `CommonTokenStream` cursor is still positioned on EOF, so `tokenStream.LB(1)` retrieves the preceding on-channel token through antlr4ng's public API. `CommonTokenStream` defines `LB` as channel-aware, meaning whitespace and comments on the hidden channel are skipped automatically. Empty input returns `null` and retains EOF; a semicolon result continues through the pre-existing semicolon adjustment.

This matches the Java runtime's `Parser.exitRule`; the old JS runtime was the outlier. The migration makes that dependency explicit using the new runtime's intended token-stream API. For example, `RETURN 1 // note` still extracts as `RETURN 1`, because `LB(1)` skips the hidden space and comment. The same call works for both the real final EOF and the per-statement EOF fabricated by `ListTokenSource`.

**Who feels it:** the extracted `cmd.statement` is what `wrappedSemanticAnalysis` receives per statement (`syntaxValidation.ts:462`) — an over-long extraction would hand Neo4j's semantic analyzer trailing comments or, worse, text past the statement boundary, shifting its diagnostics. The corrected `stop` also flows into `ParsedCommand`'s position tokens used for statement-at-caret mapping. Covered by the console-commands suite (multi-statement, trailing-comment, and empty-input cases) and the semantic-validation suite.

cypherLanguageService.ts

### 2.3 Statement splitting: ListTokenSource — no internals

**Why we touch `tokens` at all:** `splitIntoStatements` lexes the whole file once (`tokenStream.fill()`), splits the token array at SEMICOLON/EOF boundaries, and clones each token through antlr4ng's public `CommonToken.fromToken` factory. Each clone gets a statement-local index while preserving its original character offsets and input source — the property 2.2's text extraction relies on.

**What:** in the old runtime the stream's `tokens` field was public; antlr4ng makes it *protected* and offers no setter. The replacement is fully public API: antlr4ng ships `ListTokenSource`, a `TokenSource` wrapper around a token array, so each chunk simply gets `new CommonTokenStream(new ListTokenSource(chunk))`. No lexer is involved when a statement stream is parsed anymore (previously all statement streams shared one lexer as a fallback token source).

**Separation of responsibilities:** the splitter now returns only `Token[][]`. Parsing callers wrap each chunk in `ListTokenSource` and `CommonTokenStream` when they actually need a parser; lexer-only callers consume the arrays directly. This keeps tokenization/splitting separate from parser construction and avoids carrying a stream and its backing array as duplicate state.

**One behavioral nuance:** for a semicolon-terminated (non-final) statement the parser previously over-read into the shared lexer's whole-file EOF; it now gets an EOF fabricated by `ListTokenSource`, positioned at the end of the chunk. The `LB(1)` lookup in 2.2 returns the token before either kind of EOF, so statement extraction is identical; error diagnostics for a broken mid-file statement now point at the statement's end (e.g. the `;` in `MATCH (n; RETURN 1`) rather than at end of input — arguably the better position. Covered by the console-commands and semantic-validation suites.

**Alternatives considered:**

- **Re-lex each statement chunk** with a fresh lexer over the statement text — also public API, but re-tokenizes (slower) and produces different token objects/indices than the whole-file pass.
- **Read and write the protected `tokens` field** via typed casts — works, but depends on antlr4ng internals; `ListTokenSource` achieves the same through public API.

helpers.ts · cypherLanguageService.ts · formatting/formattingHelpers.ts

### 2.4 parseStatementsStrs: split without parsing

**What:** it used to fully *parse* every statement — full parser run with error strategy and all collector listeners attached — just to return the statement strings. Now it calls the pure lexer-level `getStatementTokenChunks` helper and joins each chunk's token text. It neither runs nor constructs a parser, and does not create per-statement token streams. Same output, strictly less work.

**Why the old version parsed at all:** it simply reused `parse()`, but statement boundaries are a lexer-level question (semicolons outside strings/comments), so the parse was overkill from the start — the rewrite fell out naturally from the scaffolding refactor.

**Cross-check with Neo4j's cypher-shell:** cypher-shell makes the same architectural choice in `ShellStatementParser.readCypher`: it scans input while tracking quote and comment delimiters, and ends a statement only on a semicolon outside those regions; it does not invoke the Cypher parser to find boundaries. This implementation delegates the delimiter state machine to the generated lexer instead of maintaining a second handwritten Cypher-aware scanner. The output contracts remain intentionally different: cypher-shell strips leading whitespace/comments and the terminating semicolon and records whether a statement was complete, while `parseStatementsStrs` preserves the existing source chunks used by the VS Code execution path.

**Who calls it:** the VS Code extension's `cypherRunner.ts` — the "run query" path splits the editor text into statements to execute them one by one. So this is a user-facing path, and it now skips a full parse per run. Output equivalence is covered by the existing unit tests.

**Alternative:** keep the parse-based version — works, but pays a full parse per statement for a lexer-level question.

cypherLanguageService.ts:181 · vscode-extension/src/cypherRunner.ts

### 2.5 Remove the antlrUtils compatibility wrapper

**What:** `antlrUtils.tree.Trees` is gone. The language-support package now re-exports antlr4ng's named `Trees` value and `ParseTree` type directly, and the CodeMirror playground uses those upstream APIs.

**Why the wrapper existed:** it was glue code from the start. The old `antlr4` runtime exposed `Trees` only through an untyped default-export shape, so the repo used `@ts-ignore`, a hand-written nested interface, and a double cast merely to make two tree helper calls. It was introduced for the playground's parse-tree viewer rather than as a designed language-support abstraction. The declared interface was also misleading: `getChildren` was typed as returning only `ParserRuleContext[]`, even though parse trees also contain terminal nodes.

**Why removal is better now:** antlr4ng exports `Trees` and `ParseTree` as first-class, typed symbols. Re-exporting them keeps consumers on the same runtime copy as the generated parser while removing the fake antlr4 namespace, the ignored type error, the unused third parser argument to `getNodeText`, and the incorrect rule-context-only child type. The playground now walks the honest `ParseTree` union, including terminal nodes.

**Compatibility check:** the only in-repo/upstream consumer was the CodeMirror playground. A search of the adjacent `~/work/upx` monorepo found no use, and public GitHub searches found no other Neo4j or external consumer. The symbol was publicly exported, so an unknown private consumer cannot be ruled out, but it was undocumented and the available evidence says the removal is low risk.

**Alternative considered:** retain `antlrUtils` as a deprecated adapter. That would preserve the old import name, but also keep a repo-specific API shape whose only known consumer has already moved to the upstream API. Removing it makes the migration's end state smaller and clearer.

packages/language-support/src/helpers.ts · packages/language-support/src/index.ts · packages/react-codemirror-playground/src/treeUtil.ts

### 3.1 Default exports → named exports

**What:** antlr-ng generates named exports only, so `import CypherLexer from …` became `import { CypherCmdLexer as CypherLexer } from …` throughout. `index.ts` keeps the public names (`CypherLexer`, `CypherParser`, `CypherParserListener`, `CypherParserVisitor`) via aliases, so downstream packages are untouched.

### 3.2 List accessors: `ctx.expression_list()` → `ctx.expression()`

**What:** antlr-ng names repeated-rule accessors after the rule (overloaded: `ctx.expression()` returns all, `ctx.expression(i)` returns one); same for token lists (`COMMA_list()` → `COMMA()`). This is the bulk of the formatting.ts diff — dozens of call sites, all mechanical, verified by the full 11,933-query formatting integrity corpus.

### 3.3 Context/plumbing renames

- `ctx.parentCtx` → `ctx.parent`
- `parser._ctx` → `parser.context` (also removes a `@ts-expect-error` on `ruleIndex`)
- `parser._parseListeners = […]` → `removeParseListeners()` + one `addParseListener()` per listener (antlr4ng encapsulates the list)
- `CharStreams.fromString` → `CharStream.fromString`
- `inputStream.getText(start, stop)` → `getTextFromRange(start, stop)`
- `tokenStream.tokens` reads → public `getTokens()`, or the chunk arrays handed back directly by `splitIntoStatements` (see 2.3)
- `VariableCollector` now receives the `CommonTokenStream` via constructor instead of reaching through `ctx.parser.getTokenStream()`

### 3.4 Listeners: `extends ParseTreeListener` → `implements`

**What:** in antlr4ng `ParseTreeListener` is a pure interface, so the collector classes implement it with explicit no-op methods (`enterEveryRule`, `visitTerminal`, `visitErrorNode`) instead of inheriting them.

### 3.5 `isErrorNode()` hack → `instanceof ErrorNode`

**What:** the old runtime had no real `ErrorNode` class in its types, so formatting did a `@ts-expect-error` + runtime `node.isErrorNode()` dance. antlr4ng has the class, so both sites are now plain `instanceof` checks — strictly cleaner.

formatting.ts (2 sites)

### 3.6 Error-listener signature update

**What:** `SyntaxErrorsListener.syntaxError` now matches antlr4ng's signature: `offendingSymbol` may be `null`, the recognizer may be a lexer or parser, and two extra parameters (`msg`, `e`) arrive. The listener now guards the symbol and narrows the recognizer with `instanceof CypherParser` before using parser-specific state.

syntaxValidationHelpers.ts

### 3.7 Grammar action: `this.getModeStack()` → `this.modeStack`

**What:** the embedded lexer action in `CypherCmdLexer.g4` (the `RCURLY` override that makes interpolated strings lex correctly). antlr4ng exposes `modeStack` as a getter property; there is no `getModeStack()` method, so the old form would not compile. Not a gratuitous hand edit.

**Clobber risk:** none — the monthly artifact bot only overwrites `Cypher25Lexer/Parser.g4`, `semanticAnalysis.js` and the lint-worker version tag; `CypherCmd*.g4` are repo-local (now stated explicitly in AGENTS.md, and the .g4 comment names antlr4ng).

packages/language-support/src/antlr-grammar/CypherCmdLexer.g4:32

### 4.1 Deleted: vendor/antlr4-c3 and its tests

**What:** ~7k lines: the forked completion core, its build config, and two test suites — `CodeCompletionCore.spec.ts` (13 cases, upstream's own engine suite run against the fork; that coverage now lives in upstream antlr4-c3's CI) and `CodeCompletionOptionals.spec.ts` (8 cases testing the fork-only feature, plus its `Optionals.g4` fixture — moot now that the feature is dropped, see 1.3). No product test was deleted; the only expectation change in `packages/` is the `style reset` entry (1.3) plus pure API renames in `lexer.test.ts`.

### 4.2 Workspace & config cleanup

- `pnpm-workspace.yaml`: dropped `vendor/*`
- `vitest.workspace.ts`: dropped the vendor project
- `knip.json`: dropped the vendor ignore and the `antlr4` binary ignore
- `check-imports`: vendor paths removed; it no longer scans language-support, where tsc now enforces import extensions itself (see 4.3)
- `packages/language-support/tsconfig.json`: `include` no longer reaches into the vendor dist
- CI: the "Setup antlr" jar-download step is gone from `setup-and-build`

**Also cleaned up:** stale vendor globs in `.lintstagedrc.json` and `.oxfmtrc.json`, the dead commented project reference in the root `tsconfig.json`, the "from us and vendor" cache comment, and the obsolete renovate `antlr4` rule (replaced with an `antlr4ng`/`antlr4-c3` lockstep rule).

### 4.3 language-support moves to moduleResolution nodenext

**What:** language-support now uses `module/moduleResolution: "nodenext"`, so TypeScript enforces `.js` on relative imports. The apparent blocker was not antlr4's legacy declarations but our own direct antlr4 type imports; after migration, the remaining legacy dependency is covered by `skipLibCheck`. The conflicting `--module esnext` build overrides were removed, and a rebuild verified unchanged ESM output apart from intended source edits.

**Scope:** `check-imports` no longer scans language-support. Query-tools and lint-worker retain it because they are not `"type": "module"` and have different CJS/ESM bundling constraints.

packages/language-support/tsconfig.json · packages/language-support/package.json · package.json · .github/workflows/ci.yaml · AGENTS.md

Generated from a full read of `git diff main` on branch `finish-antlr-ng-migration`. Verification status: 1,115 language-support tests and the formatting integrity corpus are green; type-aware oxlint, oxfmt, knip, and check-imports are clean; language-support rebuilt under nodenext with byte-identical ESM output apart from the source edits above.


.github/workflows/ci.yaml
```diff
@@ -40,10 +40,10 @@ jobs:
...
       - name: Check relative imports have .js extensions
         # Ideally we'd use moduleResolution: "nodenext" which enforces .js extensions at the
-        # TypeScript level, but antlr4@4.13.2's type declarations (.d.cts) use extensionless
-        # imports that break nodenext resolution. Once we switch to depending on antlr4-ng,
-        # both directly and in our (currently vendored) antlr4-c3 dependency, we can switch
-        # to nodenext and remove this grep-based check.
+        # TypeScript level, but the legacy language-support version pinned in lint-worker still
+        # pulls in antlr4@4.13.2 whose type declarations (.d.cts) use extensionless imports that
+        # break nodenext resolution. Once that legacy pin is removed, we can switch to nodenext
+        # and drop this grep-based check.
```
OskarDamkjaer: todo investigate if this is still true

packages/language-support/src/syntaxValidation/syntaxValidationHelpers.ts
```diff
@@ -28,20 +29,22 @@ export class SyntaxErrorsListener implements ANTLRErrorListener<CommonToken> {
     this.consoleCommandsEnabled = consoleCommandsEnabled;
   }
 
-  public syntaxError<T extends Token>(
-    recognizer: Recognizer<T>,
-    offendingSymbol: T,
+  public syntaxError<S extends Token, T extends ATNSimulator>(
+    recognizer: import('antlr4ng').Recognizer<T>,
```
OskarDamkjaer: todo avoid inline import

packages/language-support/src/helpers.ts
```diff
@@ -141,31 +145,50 @@ export function findCaret(
...
 }
 
+/**
+ * Helper to access the internal tokens array from a CommonTokenStream.
+ * In antlr4ng, the tokens property is protected, but we need direct access
+ * for performance and compatibility with existing patterns.
+ */
+export function getStreamTokens(tokenStream: CommonTokenStream): Token[] {
+  return (tokenStream as unknown as { tokens: Token[] }).tokens;
+}
```
OskarDamkjaer: verify downstream usage, to confirm if we actually need this direct access and what refactors would be needed

packages/language-support/src/helpers.ts
```diff
@@ -141,31 +145,50 @@ export function findCaret(
...
+
+/**
+ * Helper to set the internal tokens array on a CommonTokenStream.
+ */
+function setStreamTokens(
```
OskarDamkjaer: It seems like bas practise to set this internal field. Why must we do it, what are we trying to accomplish and could we doi it in another way?

.github/workflows/ci.yaml
OskarDamkjaer: Investigated — no longer true. `tsc` passes under `module`/`moduleResolution: "nodenext"` for all three checked packages. The original blocker was our own source importing antlr4's types; post-migration, antlr4 is only reachable through the legacy pin's `.d.ts` (lint-worker uses just `DbSchema` from it), and `skipLibCheck` covers the rest.

Went ahead and switched language-support in 3ac0d1a (`dist/esm` emit diffed byte-identical modulo intended edits), so `check-imports` no longer scans it. query-tools and lint-worker stay on the grep: they aren't `"type": "module"`, so nodenext would detect their files as CJS and not enforce extensions anyway. The comment here now records that reasoning.

packages/language-support/src/syntaxValidation/syntaxValidationHelpers.ts
OskarDamkjaer: Done in 8134c5ac — hoisted to the top-level `import type { ATNSimulator, Recognizer } from 'antlr4ng'`.

packages/language-support/src/helpers.ts
OskarDamkjaer: Verified: three call sites, all inside language-support, never re-exported from `index.ts`. Removed in 8134c5ac — `splitIntoStatements` now returns each chunk array alongside its stream (so the scaffolding doesn't read tokens back out of the stream at all), `parseStatementsStrs` uses `statement.tokens` directly, and formatting uses the public `getTokens()` after `fill()`.

packages/language-support/src/helpers.ts
OskarDamkjaer: Agreed — it's gone in 8134c5ac. What we need is per-statement parsing of pre-lexed token chunks without re-lexing. antlr4ng's public `ListTokenSource` wraps a token array as a `TokenSource`, so `splitIntoStatements` now builds `new CommonTokenStream(new ListTokenSource(chunk))` — no internals touched, and the statement streams no longer share one lexer as a fallback token source.

One nuance: non-final statements now see an EOF fabricated by `ListTokenSource`, positioned at the chunk's end, instead of the shared lexer's whole-file EOF. Statement extraction is unaffected (it walks back from EOF to the last real token), and errors on broken mid-file statements now point at the statement's end — e.g. the `;` in `MATCH (n; RETURN 1` — arguably the better position. Full unit suite and the 11,933-query formatting corpus are green.

package.json
```diff
@@ -32,6 +32,7 @@
   "devDependencies": {
     "@changesets/cli": "^2.28.1",
     "@types/node": "^24.10.1",
+    "antlr-ng": "^1.0.10",
```
anderson4j: TODO: Check why we need this when we didnt need old antlr at top-level

Maybe to generate parsers now that java-gen is gone?

.github/workflows/ci.yaml
```diff
@@ -39,11 +39,14 @@ jobs:
...
-        # TypeScript level, but antlr4@4.13.2's type declarations (.d.cts) use extensionless
-        # imports that break nodenext resolution. Once we switch to depending on antlr4-ng,
-        # both directly and in our (currently vendored) antlr4-c3 dependency, we can switch
-        # to nodenext and remove this grep-based check.
+        # language-support enforces .js extensions at the TypeScript level instead
+        # (moduleResolution: "nodenext" — the antlr-ng migration removed antlr4 from our own
+        # type surface, and skipLibCheck covers the antlr4 .d.cts files still reachable through
+        # lint-worker's legacy language-support pin, so nodenext type-checks clean).
+        # This grep-based check remains for query-tools and lint-worker: they are not
+        # "type": "module" packages, so nodenext would detect their files as CJS and not
```
anderson4j, line 47: We can skip this entire thing if we shift lint-worker and query-tools to nodenext too - ive a PR up for it.

Github is advertising something called "stacking" PRs, which would let us chain the 2 (that PR is branched off from this one)

packages/language-support/src/antlr-grammar/CypherCmdLexer.g4
```diff
@@ -27,7 +27,8 @@ ACCESSMODE: A C C E S S '-' M O D E;
 HELP: H E L P;
 // Overrides RCURLY from Cypher25Lexer, whose embedded action is Java
 // (that grammar is generated for the Java target in the Neo4j monorepo).
-// This is the equivalent action for the TypeScript runtime.
+// This is the equivalent action for the antlr4ng TypeScript runtime
+// (`modeStack` is an antlr4ng getter; there is no `getModeStack()`).
```
anderson4j: I think this is more of a "explanation because we're changing from antlr to antlr-ng" comment rather than one explaining how the code works. The old comment is still a fine explanation to me

packages/language-support/src/autocompletion/completionCoreCompletions.ts
anderson4j: looks like we get indexes right away as followUpList now - might as well remove the rename below and use `followUpIndexes` here

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -273,13 +271,6 @@ function getTokenCompletions(
           ' ' +
           (isConsoleCommand ? followUpString.toLowerCase() : followUpString);
 
-        if (followUpList.optional) {
```
anderson4j: TODO: This doesnt change behaviour?

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -489,7 +480,7 @@ export function completionCoreCompletion(
   // If the previous token is an identifier, we don't count it as "finished" so we move the caret back one token
   // The identifier is finished when the last token is a SPACE or dot etc. etc.
   // this allows us to give completions that replace the current text => for example `RET` <- it's parsed as an identifier
-  // The need for this caret movement is outlined in the documentation of vendor/antlr4-c3 in the section about caret position
+  // The need for this caret movement is outlined in the documentation of antlr4-c3 in the section about caret position
```
anderson4j: A bit wary of our comments pointing to the documentation of a repo we dont own... I suppose we could take the risk - or just make a local copy of that readme section - ill leave it to your judgement

packages/language-support/src/antlr-grammar/CypherCmdLexer.g4
angrykoala: Is this last comment addition needed? not much context left of `getModeStack()` is after this change

packages/language-support/src/autocompletion/completionCoreCompletions.ts
```diff
@@ -251,7 +249,7 @@ function getTokenCompletions(
         ? tokenNames[tokenNumber].toLowerCase()
         : tokenNames[tokenNumber];
 
-      const followUpIndexes = followUpList.indexes;
+      const followUpIndexes = followUpList;
```
angrykoala: **nitpick** Maybe remove this extra variable?

.github/workflows/ci.yaml
```diff
@@ -39,11 +39,14 @@ jobs:
...
+        # enforce extensions (lint-worker also only ships esbuild bundles, where source
+        # import specifiers don't survive into the output).
         # We also can't use oxlint's import/extensions rule because it doesn't support
```
angrykoala, line 50: I'm not sure this comment is too clear

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -122,18 +124,22 @@ function isSymbolicName(node: TerminalNode): boolean {
...
   let unParseable: string | undefined;
   let firstUnParseableToken: Token | undefined;
-  if (tree.exception) {
-    const idx = tree.exception.offendingToken.tokenIndex;
-    const errorTokens = tokens.tokens.slice(idx);
+  const rootOffendingToken = errorTracker.offendingTokenAt(tree);
+  if (rootOffendingToken) {
+    const idx = rootOffendingToken.tokenIndex;
+    const allTokens = tokens.getTokens();
+    const errorTokens = allTokens.slice(idx);
```
anderson4j: doesnt seem to me like there's much of a point to making this 2 lines

packages/language-support/src/formatting/formattingHelpers.ts
```diff
@@ -143,7 +149,7 @@ export function getParseTreeAndTokens(query: string) {
         .slice(0, -1)
         .map((t) => t.text)
         .join('');
-    firstUnParseableToken = tree.exception.offendingToken;
+    firstUnParseableToken = rootOffendingToken;
```
anderson4j: Another line that is a just renaming an existing var. Think we should just pick the correct name to begin with (either rename to `firstUnParseableToken` above, or return `rootOffendingToken`)

packages/language-support/src/syntaxValidation/syntaxValidationHelpers.ts
```diff
@@ -28,20 +29,26 @@ export class SyntaxErrorsListener implements ANTLRErrorListener<CommonToken> {
...
 
-  public syntaxError<T extends Token>(
+  public syntaxError<S extends Token, T extends ATNSimulator>(
     recognizer: Recognizer<T>,
-    offendingSymbol: T,
+    offendingSymbol: S | null,
     line: number,
     charPositionInLine: number,
+    _msg: string,
+    _e: RecognitionException | null,
```
anderson4j: Why are we passing but not using these? For debugging?

TODO: Or to ensure we implement a required method of ANTLRErrorListener ?

packages/language-support/src/syntaxValidation/syntaxValidationHelpers.ts
```diff
@@ -28,20 +29,26 @@ export class SyntaxErrorsListener implements ANTLRErrorListener<CommonToken> {
     this.consoleCommandsEnabled = consoleCommandsEnabled;
   }
 
-  public syntaxError<T extends Token>(
+  public syntaxError<S extends Token, T extends ATNSimulator>(
     recognizer: Recognizer<T>,
-    offendingSymbol: T,
+    offendingSymbol: S | null,
```
anderson4j: TODO: Figure out - Why do we accept null here? It seems we just skip on null - so maybe dont accept it and check before calling instead? (Or maybe "syntaxError" is called in the parse loop outside our control - sometimes with nullish offendingSymbol...)

packages/language-support/src/lexerSymbols.ts
```diff
@@ -1,4 +1,4 @@
-import CypherLexer from './generated-parser/CypherCmdLexer.js';
+import { CypherCmdLexer as CypherLexer } from './generated-parser/CypherCmdLexer.js';
```
angrykoala, line 1: Question: Maybe it would make sense to use the autogenerated names instead? It is quite hard to maintain consistency on renamed imports, as most tooling will just automatically import the original name and over time, we may end up with some places using `CypherCmd*` convention, and `Cypher*` in others.

In fact, `helpers.ts` is already using the other convention

packages/language-support/src/errorTrackingStrategy.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Error strategy that tracks which parser rule contexts caught a
+ * RecognitionException during parsing.
+ *
+ * In antlr4ng 3.x, ParserRuleContext.exception was removed. In the old
+ * antlr4 runtime, the generated parser's per-rule catch blocks attached the
+ * exception to that rule's context (`_localctx.exception = re`). Those catch
+ * blocks call `errorHandler.reportError(this, re)` before recovering, so
+ * overriding reportError captures exactly the same contexts — including
+ * reports the default strategy suppresses while in error recovery mode.
```
angrykoala: Is it relevant the context of the "old" runtime?

After this migration there isn't really any old runtime to worry about right?

packages/language-support/src/errorTrackingStrategy.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export class ErrorTrackingStrategy extends DefaultErrorStrategy {
+  /** First exception caught per rule context, in catch order. */
+  readonly caughtExceptions = new Map<
+    ParserRuleContext,
+    RecognitionException
+  >();
+
+  override reportError(recognizer: Parser, e: RecognitionException): void {
```
angrykoala, line 25: I see that `override` is a typescript feature that was not used anywhere in this project, just double checking if we want to add it for this

angrykoala:
LGTM, just a few questions and small comments

packages/language-support/src/tests/consoleCommands.test.ts
```diff
@@ -969,6 +975,16 @@ describe('command parser also handles cypher', () => {
...
+  test('can split statement strings without parsing them', () => {
+    expect(parseStatementsStrs("RETURN ';'; // comment\n RETURN 2;  ")).toEqual(
+      ["RETURN ';';", ' // comment\n RETURN 2;'],
+    );
+    expect(parseStatementsStrs('MATCH (n; RETURN 1')).toEqual([
+      'MATCH (n;',
+      ' RETURN 1',
+    ]);
+  });
+
```
anderson4j: Seems like useful tests, but they dont test console commands. The existing one uses `expectParsedCommands` so I guess that's why it was here. Still, maybe we should have a different file, something like `statementParsing.test.ts`?

packages/language-support/src/cypherLanguageService.ts
```diff
@@ -88,6 +92,7 @@ interface ParsingScaffolding {
 
 interface StatementParsingScaffolding {
   parser: CypherParser;
+  tokenStream: CommonTokenStream;
```
anderson4j: TODO: Figure why we need this now

packages/language-support/src/cypherLanguageService.ts
```diff
@@ -227,23 +238,29 @@ export function createParsingResult(
...
-      parser._parseListeners = [
-        labelsCollector,
-        parameterFinder,
-        variableFinder,
-        methodsFinder,
-        cypherVersionCollector,
-        propertiesFinder,
-        readPatternElementsCollector,
-      ];
+      const errorTracker = new ErrorTrackingStrategy();
```
anderson4j, line 241: Why is this removal necessary? And why not in the `addErrorListener` case then?

packages/language-support/src/cypherLanguageService.ts
```diff
@@ -227,23 +238,29 @@ export function createParsingResult(
...
+      parser.addParseListener(readPatternElementsCollector);
       parser.addErrorListener(errorListener);
       const ctx = parser.statementsOrCommands();
       // The statement is empty if we cannot find anything that is not EOF or a space
       const isEmptyStatement =
         tokens.find(
           (t) => t.text !== '<EOF>' && t.type !== CypherLexer.SPACE,
         ) === undefined;
-      const collectedCommand = parseToCommand(ctx, tokens, isEmptyStatement);
+      const collectedCommand = parseToCommand(
```
anderson4j: TODO: now tokenStream too.. why

anderson4j: Seems before we could get tokenStream like ctx.parser?.getTokenStream()
I guess ctx doesnt have parser anymore or parser doesnt have getTokenStream

packages/language-support/src/cypherLanguageService.ts
```diff
@@ -528,12 +562,10 @@ class VariableCollector implements ParseTreeListener {
...
       const nextTokenIsEOF =
         nextTokenIndex !== undefined &&
-        ctx.parser?.getTokenStream().get(nextTokenIndex + 1)?.type ===
-          CypherParser.EOF;
+        this.tokenStream?.get(nextTokenIndex + 1)?.type === CypherParser.EOF;
 
       const definesVariable = rulesDefiningOrUsingVariables.includes(
-        // @ts-expect-error the antlr4 types don't include ruleIndex but it is there, fix as the official types are improved
-        ctx.parentCtx?.ruleIndex as unknown as number,
+        ctx.parent?.ruleIndex as number,
```
anderson4j: TODO: Double check if ? is needed

packages/language-support/src/cypherLanguageService.ts
```diff
@@ -725,25 +752,34 @@ type ParsedCommand = ParsedCommandNoPosition & RuleTokens;
...
-  const stmt = stmts.statementOrCommand_list().at(0);
+  const stmt = stmts.statementOrCommand().at(0);
 
   if (stmt) {
     const start = stmts.start;
     let stop = stmts.stop;
 
+    // Matching EOF sets the context stop to EOF. CommonTokenStream.LB() is
+    // channel-aware, so this recovers the preceding default-channel token.
+    if (stop && stop.type === Token.EOF) {
```
anderson4j: TODO: Check how this affects something like
`:help; `
(compare old vs new parse)

packages/language-support/src/errorTrackingStrategy.ts
```diff
@@ -0,0 +1,44 @@
...
+    const ctx = recognizer.context;
+    if (ctx && !this.caughtExceptions.has(ctx)) {
```
anderson4j: This `!this.caughtExceptions.has(ctx)` bit seems a bit strange to me. Could we throw this "failed to parse" exception twice on the same rule (context)?

packages/language-support/src/helpers.ts
```diff
@@ -21,16 +21,16 @@ import { CypherVersion } from './types.js';
...
            / \
           /   \
-TerminalNode   RuleContext
+ TerminalNode   RuleContext
                 \
-                ParserRuleContext                 
+                ParserRuleContext
 
-Both TerminalNode and RuleContext have parentCtx, but ParseTree doesn't
+Both TerminalNode and RuleContext have parent, but ParseTree doesn't
```
anderson4j: Looks like this isnt true anymore, so `EnrichedParseTree` -> `ParseTree` should be fine

packages/language-support/src/helpers.ts
```diff
@@ -1,14 +1,14 @@
...
   CommonTokenStream,
   ParserRuleContext,
   ParseTree,
   Token,
-} from 'antlr4';
+} from 'antlr4ng';
 import { DbSchema } from './dbSchema.js';
-import CypherLexer from './generated-parser/CypherCmdLexer.js';
-import CypherParser, {
+import { CypherCmdLexer } from './generated-parser/CypherCmdLexer.js';
```
anderson4j: I think the decision made in other places to use the same naming convention as for the old default exports (`CypherCmdLexer as CypherLexer`) was a good one. Let's do the same here?

At least for this PR I think it simplifies an already quite comprehensive diff. Regarding Andres' point about default auto-imports being different maybe that's enough to switch it everywhere, though im happy to have it in a cleaning PR instead of this one

packages/language-support/src/helpers.ts
```diff
@@ -141,33 +127,25 @@ export function findCaret(
   return result;
 }
 
-export function splitIntoStatements(
-  tokenStream: CommonTokenStream,
-  lexer: CypherLexer,
-): CommonTokenStream[] {
+export function splitIntoStatements(tokenStream: CommonTokenStream): Token[][] {
```
anderson4j: TODO: Figure out why we switch return type

anderson4j (commented):
Good stuff, not easy to replace something so central to the project! I had a few comments, and a small PR to further simplify that we could stack on if you like it

(Think Andres is reviewing as I write, so we might get some overlap)

anderson4j:
Oh also, I confirmed that the new gen-parser command works fine on windows, and that executing multiple statements in vscode with the new lexer-based `parseStatementsStrs` (used in query execution on vscode) worked fine for some different multi-statement cases

packages/language-support/src/autocompletion/completionCoreCompletions.ts
OskarDamkjaer: i considered it but didn't do it, happy to update :+1:

packages/language-support/src/lexerSymbols.ts
OskarDamkjaer: good point, let's avoid the rename

packages/language-support/src/errorTrackingStrategy.ts
OskarDamkjaer: true, these "how it worked before" have a tendancy to sneak in. I got rid of a bunch but not all it seems (my bad), thanks for the catch I'll clean it up

packages/language-support/src/errorTrackingStrategy.ts
OskarDamkjaer: I think it'd be nice to add `noImplicitOverride` and use the override consistentlty, no strong pref

.github/workflows/ci.yaml
OskarDamkjaer: Cool! I approved that one!

packages/language-support/src/autocompletion/completionCoreCompletions.ts
OskarDamkjaer: Hmm, maybe a link is better? than it can at least be "way-back machined"

packages/language-support/src/formatting/formattingHelpers.ts
OskarDamkjaer: makes sense

packages/language-support/src/formatting/formattingHelpers.ts
OskarDamkjaer: ah yes it could easily be one

packages/language-support/src/tests/consoleCommands.test.ts
OskarDamkjaer: different file is a good call

packages/language-support/src/helpers.ts
OskarDamkjaer: I'm thinking we can do it as a follow up point. I'll still udpate the ones form andres

packages/language-support/src/helpers.ts
OskarDamkjaer: thanks! good catch

packages/language-support/src/cypherLanguageService.ts
OskarDamkjaer: I think the _parseListeners is not settable any more directly

packages/language-support/src/errorTrackingStrategy.ts
OskarDamkjaer: I think it shouldn't happen and that we're fine without the guard

.github/workflows/ci.yaml
OskarDamkjaer: I think this comment gets removed as we do isaks change to using nodenext :+1:

packages/language-support/src/autocompletion/completionCoreCompletions.ts
anderson4j: Good idea!

packages/language-support/src/tests/consoleCommands.test.ts
anderson4j: Ah it seems I forgot (or failed) to mark the tests i meant in the comment. I was thinking
`excludes trailing hidden-channel tokens from a statement` and `can split cypher into statements` are more about statement parsing too than about console commands, so they would fit in the new file too.

They do use the "parse to command" logic though, so I'll leave it to your judgement

packages/language-support/src/cypherLanguageService.ts
anderson4j: Yeah makes sense, I was just thinking that the old logic was "set because setting is easier than pushing each new listener" rather than the removal being something we explicitly wanted. I see you removed the `removeParseListeners` call, so I guess you landed in the same place

anderson4j (approved):
LGTM now. One small comment on a test file you can change if you want

angrykoala (approved):

Issue #717, opened 2026-07-29
Support defining connections via workspace settings (env var references)


goldsam:
## Description
The Neo4j for VS Code extension currently only supports adding/editing database connections through the UI (Connection pane or `Neo4j: Create New Connection` command). Connection details are stored in the extension's internal/global storage and can't be defined declaratively.

## Use case
Teams often manage local dev credentials (e.g. Neo4j username/password) via environment variables or .env files shared across a devcontainer/docker-compose setup, so onboarding a new contributor shouldn't require manually re-entering connection details.

## Feature request
Allow connections to optionally be defined in `.vscode/settings.json`, with support for referencing environment variables for sensitive fields, e.g.:

```jsonc
"neo4j.connections": [
  {
    "name": "Local dev",
    "scheme": "neo4j",
    "host": "localhost",
    "port": 7687,
    "username": "${env:NEO4J_USERNAME}",
    "password": "${env:NEO4J_PASSWORD}"
  }
]
```

This would let workspace-level connection definitions be version-controlled (referencing env vars for secrets) rather than requiring each contributor to manually recreate the connection via the UI.

## Alternatives considered
- Manually creating the connection via the UI works but isn't reproducible across clones/onboarding

anderson4j:
Made some minor changes, but merged in the feature pretty much as you requested today. Thanks for contributing with ideas!

Will be brought into our next release, or tomorrow's pre-release if you're eager to try it early

PR #731, opened 2026-08-21
Allow connections via settings.json

anderson4j:
Allows users to setup connections via settings.json like suggested [here](https://github.com/neo4j/cypher-language-support/issues/717)

These connections appear in the same connection tree as regular connections, but with a little `(settings.json)` tag at the end. They can't be deleted/edited from the UI. Some things however - the state -  (active/inactive, current db, approved) are stored in the globalstate of the extension like we do with full "normal" connections. 

Here's an example of my setup of a db using env variables for user/password
image

The clover bot made a good point about a security concern. if a user with this feature were to clone a repo containing a malicious settings.json, they could receive connections pointing to their own dbs, causing the user to pass secrets stored in env vars when connecting. 

To avoid this we've done 2 things. 
A: When reading these connections from the settings.json, we don't ping for a valid connection like we do when adding normal connections. Nothing is passed to the address until a user clicks to connect. 

B: When first connecting to a new server address (scheme+host+port) we prompt the user if they trust the address, like
image


packages/vscode-extension/package.json
```diff
@@ -512,6 +512,60 @@
...
+        "neo4j.connections": {
+          "type": "array",
+          "default": [],
+          "markdownDescription": "Connections to Neo4j databases. Connections defined here appear in the Connections view alongside connections created via the UI, but are managed entirely through this setting. You will be asked to confirm that you trust a server the first time you connect to it. All values except the scheme may reference environment variables with `${env:VAR}`, resolved from the environment VS Code was launched with (a full restart of VS Code is needed to pick up changes). Note that literal passwords are stored in plain text in your settings file — reference an environment variable, or prefer connections created via the UI, for sensitive credentials.",
+          "items": {
+            "type": "object",
+            "required": [
+              "scheme",
+              "host",
+              "user"
```
anderson4j, line 524: In the issue proposing the feature this was "username" but I made it user since that's what the connection panel uses.

packages/vscode-extension/src/uiUtils.ts
```diff
@@ -56,14 +56,36 @@ export async function displaySaveConnectionAnywayPrompt(
...
     `Are you sure you want to delete connection ${connectionItem.label}?`,
     { modal: true },
     'Yes',
   );
 }
 
+/**
+ * Utility function to prompt the user to confirm connecting to a server
+ * defined in the neo4j.connections setting for the first time. The approval
+ * covers the server address, so the user is not asked again for other
```
angrykoala: This part here of "The approval covers..." seems a bit misleading, as this logic is not handled by this helper

packages/vscode-extension/src/uiUtils.ts
```diff
@@ -56,14 +56,36 @@ export async function displaySaveConnectionAnywayPrompt(
...
+ * connections to the same server.
+ * @param connection The config Connection whose server to confirm.
+ * @returns A promise that resolves with the result of the prompt ("Yes" or undefined).
+ */
+export async function displayConfirmSettingConnectionPrompt(
+  connection: Connection,
+): Promise<string | undefined> {
+  const address = connection.port
+    ? `${connection.scheme}://${connection.host}:${connection.port}`
+    : `${connection.scheme}://${connection.host}`;
```
angrykoala: **nitpick** Maybe worth having a separate variable for the scheme-host path? Just to make it explicit that these are the same, regardless of it having a port or not

packages/vscode-extension/src/treeviews/connectionTreeDataProvider.ts
```diff
@@ -169,6 +179,9 @@ export const connectionTreeDataProvider = new ConnectionTreeDataProvider();
  * @param state The state of the connection, used to set the description.
  * @param label The label of the connection, used to set the label and tooltip.
  * @param collapsibleState The collapsible state of the connection.
+ * @param isConfigConnection Whether the connection comes from the neo4j.connections setting.
```
angrykoala: Just for consistency, maybe we could keep the convention of using `connection-settings` to align with the ui

packages/vscode-extension/src/commandHandlers/connection.ts
```diff
@@ -132,6 +137,20 @@ export async function toggleConnectionItemsConnectionState(
...
   const connectionToToggle = getConnectionByKey(connectionItem.key);
+
+  if (
+    connectionToToggle?.source === 'config' &&
+    connectionToToggle.state === 'inactive' &&
+    !isConfigConnectionApproved(connectionToToggle.key)
+  ) {
+    const answer =
+      await displayConfirmSettingConnectionPrompt(connectionToToggle);
+    if (answer !== 'Yes') {
```
angrykoala: **Nitpick** Maybe `Yes` could be a global constant? This makes the match between this ui handling and the modal itself explicit

packages/vscode-extension/package.json
```diff
@@ -512,6 +512,60 @@
     "configuration": {
       "title": "Neo4j for VS Code",
       "properties": {
+        "neo4j.connections": {
+          "type": "array",
+          "default": [],
+          "markdownDescription": "Connections to Neo4j databases. Connections defined here appear in the Connections view alongside connections created via the UI, but are managed entirely through this setting. You will be asked to confirm that you trust a server the first time you connect to it. All values except the scheme may reference environment variables with `${env:VAR}`, resolved from the environment VS Code was launched with (a full restart of VS Code is needed to pick up changes). Note that literal passwords are stored in plain text in your settings file — reference an environment variable, or prefer connections created via the UI, for sensitive credentials.",
```
angrykoala, line 518: Maybe this, or something like this, should also be in the README, so it is documented how to add these connection settings

packages/vscode-extension/src/connectionService.ts
```diff
@@ -46,6 +49,30 @@ export type Connections = {
 };
 
 const CONNECTIONS_KEY: string = 'connections';
+const APPROVED_CONFIG_SERVERS_KEY: string = 'approvedConfigServers';
```
angrykoala: I'm not sure this casting to string is really necessary here

packages/vscode-extension/src/connectionService.ts
```diff
@@ -159,20 +190,40 @@ export async function saveConnection(connection: Connection): Promise<void> {
...
   }
 
+  if (connection.source === 'config') {
+    const states = getConfigConnectionStates();
+    states[connection.key] = {
+      state: connection.state,
+      database: connection.database,
+    };
+    await saveConfigConnectionStates(states);
+    await commands.executeCommand(
```
angrykoala, line 198: nitpick: It is a bit weird that this command is handled here, while it is executed inside the `saveConnections` function in the normal connection path.
Could it make sense to remove this from `saveConnections` and just executed it here for both paths?

packages/vscode-extension/src/connectionService.ts
```diff
@@ -183,29 +234,36 @@ export async function getPasswordForConnection(
...
- * Gets all Connections from the global state as an array of Connection objects.
+ * Gets all Connections, both from the global state and from the
+ * neo4j.connections setting, as an array of Connection objects.
  * @returns An array of all Connection objects.
  */
 export function getAllConnections(): Connection[] {
-  const connections = Object.values(getConnections());
-  return connections.length ? connections : [];
+  const connections = Object.values(getConnections()).filter(
+    (connection): connection is Connection => !!connection,
```
angrykoala: Why this filter?

In which case does getConnection return an undefined value?

packages/vscode-extension/src/connectionService.ts
```diff
@@ -183,29 +234,36 @@ export async function getPasswordForConnection(
...
+  );
+  return [...connections, ...getConfigConnections()];
```
angrykoala: mm, I see that `getConnections()` return a `Record<string, Connection>`, whilst `getConfigConnections` returns an array. For what I see where this is used, it feels like it would make sense to keep it consistent with getConnections()

packages/vscode-extension/src/connectionService.ts
```diff
@@ -519,6 +577,17 @@ async function disconnectAllDatabaseConnections(): Promise<void> {
     }
   }
 
+  const states = getConfigConnectionStates();
+  for (const key in states) {
+    if (states[key].state !== 'inactive') {
+      states[key] = { ...states[key], state: 'inactive' };
```
angrykoala, line 578: Any reason for doing destructuring instead of modifying the object?


```suggestion
      states[key].state = 'inactive';
```

packages/vscode-extension/src/connectionService.ts
```diff
@@ -531,6 +600,290 @@ export function getConnections(): Connections {
...
+ * environment VS Code was launched with, so changes to environment variables
+ * require a full restart of VS Code to be picked up.
+ * References to unset variables are left as-is, so they are visible in the
+ * connections view rather than silently becoming empty strings.
+ * @param value The string to substitute environment variable references in.
+ * @returns The string with environment variable references substituted.
+ */
+function substituteEnvVariables(value: string): string {
+  return value.replace(
+    /\$\{env:([^}]+)\}/g,
```
angrykoala, line 646: Is this standard anywhere else in vscode? If not we should probably document this explicitly somewhere

packages/vscode-extension/src/connectionService.ts
```diff
@@ -531,6 +600,290 @@ export function getConnections(): Connections {
...
+ * Substitutes ${env:VAR} references in a string with the value of the
+ * environment variable VAR. Note that process.env is a copy of the
+ * environment VS Code was launched with, so changes to environment variables
+ * require a full restart of VS Code to be picked up.
+ * References to unset variables are left as-is, so they are visible in the
+ * connections view rather than silently becoming empty strings.
+ * @param value The string to substitute environment variable references in.
+ * @returns The string with environment variable references substituted.
+ */
+function substituteEnvVariables(value: string): string {
```
angrykoala: This may be worth having some tests to validate the regex logic, either of this function or of one using this

angrykoala (changes requested):
changeset missing

packages/vscode-extension/src/connectionService.ts
anderson4j: To me there's no harm in having it (or not), what sticks out more to is if we have `: string` specified like in CONNECTIONS_KEY above and then dont do it here (like we do for section below) - I'll remove it, also for CONNECTIONS_KEY, and then we're both happy

packages/vscode-extension/src/connectionService.ts
anderson4j: we've 3 use cases for `saveConnections`, think it's a bit off to move it out the refresh on all of them when they all need it. I've renamed it to `saveConnectionsAndRefresh` to signify the difference between the 2 use cases. (no shuffling about of the command seemed very clean for all 3 cases)

packages/vscode-extension/src/connectionService.ts
anderson4j: The existing `getConnections` returns a map of [key: string]: Connection | null. Think it's there to note that if we use an invalid key we get null. Not quite sure why we dont use Record<string, Connection>. Got a warning without the check. 

I've moved the type to record, which got rid of the warning too

packages/vscode-extension/src/connectionService.ts
anderson4j: Ive heard before it's a bad idea in general to modify objects like this, since they might be (or get in the future) retrieved somewhere else. (and thus modifying the existing object could have uninteded consequences)

packages/vscode-extension/src/connectionService.ts
anderson4j: Added as part of the readme

anderson4j:
We dont use changeset for the vscode-extension package - that one has its own release flow

Adjusted according to/responded to the other comments now

packages/vscode-extension/src/connectionService.ts
angrykoala: mm, in some cases it may, if the same object is stored somewhere else (which I don't think it is in this case) although we are already modifying the parent object with `states[key].state` anyway. This was more of a nitpick/question so I'm happy to leave it as is if there is the potential of causing issues

angrykoala (approved):

PR #734, opened 2026-08-25
Remove check imports script

anderson4j:
Oskar's PR removed the need for the "check import file extension" (.js) script on the language-support package. Thought maybe we could do the same for our other packages covered by the script. Claude managed it pretty quickly with some minor changes.


OskarDamkjaer (approved):
nice!

PR #740, opened 2026-08-28
Handle grammar-defined functions

anderson4j:
Expands the handling of grammar-defined functions like `allReduce` and `trim` to work properly with syntax highlighting and hover information.

Previously the signature help and syntax highlighting only worked for the standard case that went under the `functionInvocation` rule. We would sometimes get lucky and parse the special functions as `functionInvocation` until they were fully written, so completions with hover info mostly worked, and highlighting didnt flip until functions were finished.  Now we also listen for the special functions that have their own grammar rules, so our functionalities also work when the "real" parsing happens.
Examples:
Before (picture doesnt show cursor but I was hovering `exists`)
image

After
image

Closes LS-212


packages/language-support/src/signatureHelp.ts
```diff
@@ -84,14 +105,220 @@ function toSignatureHelp(
...
-  result: ParsedMethod;
+  result?: ParsedMethod;
   constructor(
     private tokens: Token[],
     private caretToken: Token,
   ) {
     super();
   }
 
+  shouldGiveSignatureHelp(ctx: MethodContext): boolean {
```
angrykoala: This should probably be `private`

packages/language-support/src/signatureHelp.ts
```diff
@@ -84,14 +105,220 @@ function toSignatureHelp(
...
+    return (
+      isDefined(ctx.start) &&
+      isDefined(ctx.stop) &&
+      ctx.start.start <= this.caretToken.start &&
+      this.caretToken.stop <= ctx.stop.stop &&
+      isDefined(openingToken)
+    );
+  }
+
+  handleAllReduce = (
```
angrykoala: Maybe a slight rename, to make more explicit that `AllReduce` refer to the expression

```suggestion
  handleAllReduceExpression = (
```

angrykoala (approved):

PR #742, opened 2026-09-02
Upgrade to vite 8

danieladugyan:
Closes LS-180
Closes #727

### Summary
- Upgrade Vite 4 → 8, with matching `@vitejs/plugin-react` 6, Vitest 4.1, and Playwright CT 1.62 (Vite 8 peers).
- Migrate Vitest workspace config to `test.projects`, adapt CodeMirror Vitest/Playwright configs for Vitest 4 and Rolldown, and add Node to `types` where builds rely on `vitest/globals`.


anderson4j (commented):
Think it would be nice to modify the react-codemirror tests `new props.value should cancel onChange` and `new props.value set to same value should cancel onChange` to something that isnt doing "test.fails" (ex. switching to "expect(...).**not**.toBe(...)" as the current state makes the whole test run of the package look as if it failed in the UI.

Image

Probably worth adding some explanation comment there too - I think in person you said something like "Documenting odd behaviour"

danieladugyan:
> ...as the current state makes the whole test run of the package look as if it failed in the UI.

@anderson4j I'm fairly sure that's due to a bug in the vitest VS Code extension since it works as expected when using the CLI. I've created an issue on their end: https://github.com/vitest-dev/vscode/issues/812.

But for now, I've made some small changes to the tests clarifying what's going on and inverting the assertions so that the tests are no longer tagged with `fails`.

package.json
```diff
@@ -46,7 +46,8 @@
     "ts-node": "10.9.1",
     "tsx": "^4.19.4",
     "typescript": "5.8.3",
-    "vitest": "^2.1.9"
+    "vite": "^8.2.2",
```
angrykoala: I'm not sure why we need `vite` as a dependency of the top level package.json now

danieladugyan: Oops sorry, good catch. I've removed it.

anderson4j (approved):
LGTM

PR #745, opened 2026-09-03
Make tsconfig consistent across packages and fix type errors

angrykoala:
This PR sets the baseline for a more consistent `tsconfig` setup, changes have been kept minimal, but some existing type errors are fixed, so we can begin work on enabling more strict settings

### DIFF
--- a/packages/query-tools/src/cypher-execution/query-result.ts
+++ b/packages/query-tools/src/cypher-execution/query-result.ts
@@ -99,7 +99,8 @@ export function serializeTypeAnnotations(
     return item.map((i) => serializeTypeAnnotations(i));
   }
   if (item === null || item === undefined) {
-    return item;
+    // This is needed due to the method overloading types
+    return item as any;
   }
 
   if (item instanceof types.Record) {
--- a/packages/query-tools/src/cypher-execution/query-result.ts
+++ b/packages/query-tools/src/cypher-execution/query-result.ts
@@ -99,8 +99,7 @@ export function serializeTypeAnnotations(
     return item.map((i) => serializeTypeAnnotations(i));
   }
   if (item === null || item === undefined) {
-    // This is needed due to the method overloading types
-    return item as any;
+    return item;
   }
 
   if (item instanceof types.Record) {
--- a/packages/query-tools/tsconfig.json
+++ b/packages/query-tools/tsconfig.json
@@ -6,6 +6,7 @@
     "moduleResolution": "nodenext",
     "resolveJsonModule": true,
     "types": ["vitest/globals"],
-    "strictNullChecks": true
+    "strict": true, // TODO: Move to tsconfig.base.json
+    "strictNullChecks": true // TODO: Move to tsconfig.base.json
   }
 }
--- a/packages/query-tools/tsconfig.json
+++ b/packages/query-tools/tsconfig.json
@@ -2,10 +2,10 @@
   "extends": "../../tsconfig.base.json",
   "include": ["src", "package.json", "vitest.config.mts"],
   "compilerOptions": {
-    "strict": true,
     "module": "nodenext",
     "moduleResolution": "nodenext",
     "resolveJsonModule": true,
-    "types": ["vitest/globals"]
+    "types": ["vitest/globals"],
+    "strictNullChecks": true
   }
 }
--- a/packages/query-tools/tsconfig.json
+++ b/packages/query-tools/tsconfig.json
@@ -5,7 +5,7 @@
     "module": "nodenext",
     "moduleResolution": "nodenext",
     "resolveJsonModule": true,
-    "types": ["vitest/globals"],
+    "types": ["vitest/globals", "node"],
     "strict": true, // TODO: Move to tsconfig.base.json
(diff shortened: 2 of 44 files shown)


packages/query-tools/tsconfig.json
```diff
@@ -2,10 +2,11 @@
...
   "include": ["src", "package.json", "vitest.config.mts"],
   "compilerOptions": {
-    "strict": true,
     "module": "nodenext",
     "moduleResolution": "nodenext",
     "resolveJsonModule": true,
-    "types": ["vitest/globals"]
+    "types": ["vitest/globals"],
+    "strict": true, // TODO: Move to tsconfig.base.json
+    "strictNullChecks": true // TODO: Move to tsconfig.base.json
```
danieladugyan: TODO can be fixed now since the options have been moved to `tsconfig.base.json`.

```suggestion
    "types": ["vitest/globals"]
```

packages/react-codemirror/vite.config.ts
```diff
@@ -1,8 +1,8 @@
-/// <reference types="vitest" />
 import react from '@vitejs/plugin-react';
 import { defineConfig } from 'vite';
+import type { UserConfig } from 'vite';
 
-export default defineConfig({
+const config: UserConfig & { test: unknown } = {
```
danieladugyan: I think it's nicer to do `import { defineConfig } from 'vitest/config';`, but I couldn't get that to work on this branch without upgrading vite to at least v5. Since we're upgrading vite anyways in #742, I suppose this is fine in the meantime.

packages/react-codemirror/tsconfig.node.json
```diff
@@ -1,11 +1,10 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
```
danieladugyan, line 3: I'm getting an error in this file:
```
Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
  Visit https://aka.ms/ts6 for migration information.
```
But only in my editor, so might be fine to leave for now. It is fixed in #742.

packages/query-tools/src/cypher-execution/query-result.ts
```diff
@@ -99,7 +99,8 @@ export function serializeTypeAnnotations(
     return item.map((i) => serializeTypeAnnotations(i));
   }
   if (item === null || item === undefined) {
-    return item;
+    // This is needed due to the method overloading types
+    return item as any;
```
danieladugyan: Depending on how strict we want our linting to be, this might not be allowed after #746 since it violates [`typescript/no-explicit-any`](https://oxc.rs/docs/guide/usage/linter/rules/typescript/no-explicit-any).

danieladugyan:
```suggestion
    return item;
```

As far as I can tell, this change doesn't seem to make any difference on my machine when running `pnpm typecheck` and `pnpm build`, and depending on how strict we want our linting to be, it might not be allowed after #746 since it violates [`typescript/no-explicit-any`](https://oxc.rs/docs/guide/usage/linter/rules/typescript/no-explicit-any).

packages/query-tools/tsconfig.json
angrykoala: I don't think we can fix this in this PR, as the base options are explicitly `false`, is just that this project already has the strict to true, which is what we want. My idea was to move it to base in the follow up PRs as we enable these options

packages/react-codemirror/vite.config.ts
angrykoala: Yeah, I agree and tried `defineConfig` to work, but there had to be several typescript hacks around for it to work. If vite solves that in the other PR that's perfect

packages/query-tools/tsconfig.json
danieladugyan: Oh sorry, I must have gotten the PR's mixed up. I was thinking that they were the same.

packages/react-codemirror/tsconfig.node.json
angrykoala: I think we can leave it for now, as migrating to next typescript will probably come after we sorted out vite.

We may want, however, to have a chat about moduleResolution in general, as it would be good to try and keep it as consistent as possible in follow up changes to tsconfig

danieladugyan: Sounds good! I'm a bit shaky on how the module resolution options differ so I think a chat would be good 👍

danieladugyan (approved):
LGTM!

anderson4j (commented):
Looks like we have `noEmit=true` in every `tsconfig.node.json` - maybe this should be in an extended file in root instead? Imo we could have `allowJs=true` there too (though not all confs use it currently, I don't see any harm in doing it). If we make this in a new "tsconfig.node.base.json" file like:
```
{
  "extends": "./tsconfig.base.json",
  "compilerOptions": {
    "allowJs": true,
    "noEmit": true
  }
}
```
we can skip compileroptions completely in many files extending the base (and shorten it for others)

PR #746, opened 2026-09-03
Add oxlint rules

danieladugyan:
Updates Oxlint and enables the default plugins and a few others that seem appropriate. The remaining changes are just fixes for errors/warnings that are now reported. 

There are probably more rules that we could consider turning on in a follow up PR, but I think this is a good start.


packages/query-tools/src/integration-tests/setupTestContainer.ts
```diff
@@ -13,7 +13,7 @@ export async function createAndStartTestContainer(
   },
 ): Promise<StartedNeo4jContainer> {
   const password = 'password';
-  const env = { ...(opts.env ?? {}), NEO4J_ACCEPT_LICENSE_AGREEMENT: 'yes' };
+  const env = { ...opts.env, NEO4J_ACCEPT_LICENSE_AGREEMENT: 'yes' };
```
anderson4j, line 16: `env` is optional on `ContainerOpts` (and not present on default value) - isn't this change unsafe?

danieladugyan: Nope! 

image

packages/language-support/src/tests/formatting/linebreaks.test.ts
```diff
@@ -1326,7 +1326,7 @@ RETURN deleteNodes`;
     verifyFormatting(query, expected);
   });
 
-  test('test when reduce expression has long list', () => {
+  test('reduce expression has long list', () => {
```
anderson4j, line 1329: Sounds off now. Maybe 'reduce expression on long list'?

packages/vscode-extension/src/webviews/controllers/queryDetails.tsx
```diff
@@ -59,6 +67,7 @@ export function QueryDetails() {
           if (index !== -1) {
             newState[index] = message.result;
           }
+          openLastStatement(newState);
```
anderson4j, line 70: From my understanding, all this does is re-open the final statement whenever we update if the user opens one of the previous statements while not all statements are finished executing. Seems a bit odd to me, maybe we should fix it (by removing this call) - even if it's how things were before ?

anderson4j (approved):
LGTM, found one minor test-name fix, and an optional improvement to statement-execution in vscode since you're poking around with it anyway

packages/language-support/src/tests/formatting/linebreaks.test.ts
danieladugyan: Thanks, fixed.

packages/vscode-extension/src/webviews/controllers/queryDetails.tsx
danieladugyan: Nice catch, after trying out the change in the VS Code extension I think it makes sense to remove it 👍

PR #747, opened 2026-09-03
Set strict to true in base tsconfig

angrykoala:


packages/language-support/src/formatting/cli.ts
```diff
@@ -138,7 +138,7 @@ function main() {
   } catch (error) {
     console.error(
       'Error:',
-      error instanceof Error ? error.message : String(error),
+      error instanceof Error ? error.message : JSON.stringify(error),
```
danieladugyan, line 141: There are few changes to how errors are printed. This is required by [typescript/no-base-to-string](https://oxc.rs/docs/guide/usage/linter/rules/typescript/no-base-to-string.html) which started reported warnings when strict was enabled. I've fixed them in whatever way seemed most appropriate, but not obvious to me what's best.

angrykoala: mmm, I wonder if we may want to disable this rule (or in this instance), I don't like the idea of passing an unknown value into `JSON.stringify` as it may fail unexpectedly (for example, in circular objects). I see the rule has an option `checkUnknown` that should allow to disable this specific case

packages/react-codemirror-playground/src/App.tsx
```diff
@@ -372,7 +372,7 @@ export function App() {
                     setSchema(schema);
                     setSchemaError(null);
                   } catch (e) {
-                    setSchemaError(String(e));
+                    setSchemaError(String(e as Error));
```
angrykoala, line 375: Same a before, we may want to disable that rule for unknown values

packages/react-codemirror/src/CypherEditor.test.tsx
```diff
@@ -18,7 +18,9 @@ const onChange = vi.fn((v: string) => {
   rerender();
 });
 
-global.IS_REACT_ACT_ENVIRONMENT = true;
+(
+  globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT: boolean }
+).IS_REACT_ACT_ENVIRONMENT = true;
```
angrykoala: mmm, honestly, at this point I think it is more legible to use `any`. At the end of the day both are bypassing the type system, might as well use the easier one


```suggestion
(global as any).IS_REACT_ACT_ENVIRONMENT = true;
```

For a more elegant solution, I reckon we can define the global this in a `.d.ts` declaration file, I'm not sure if it is worth it though

packages/react-codemirror/src/lang-cypher/createCypherTheme.ts
```diff
@@ -255,8 +255,10 @@ export const createCypherTheme = ({
 
   const themeExtension = EditorView.theme(themeOptions, { dark });
 
-  const styles = Object.entries(highlightStyles).map(
-    ([token, color]: [HighlightedCypherTokenTypes, string]): TagStyle => ({
+  const styles = (
+    Object.entries(highlightStyles) as [HighlightedCypherTokenTypes, string][]
```
angrykoala: nitpick: I think the syntax `Array<[HighlightedCypherTokenTypes, string]>` is a bit more readable for nested arrays, but happy to leave it as is

packages/language-support/src/formatting/cli.ts
danieladugyan: As far as I can tell, this option doesn't seem to work. By default, `checkUnknown` is `false` so it should not report any warnings. Maybe that's why there weren't any warnings before, but for some reason enabling `strict` seems to have incorrectly triggered the warnings.

packages/react-codemirror/src/lang-cypher/createCypherTheme.ts
danieladugyan: I'm fine with changing it, there's an oxlint rule for enforcing this: [typescript/array-type](https://oxc.rs/docs/guide/usage/linter/rules/typescript/array-type.html#array-simple).

danieladugyan: It's changed in the latest commit: 4205e61251c4f95de4d90cc4ea22fad9399d0b2a

danieladugyan (approved):

PR #753, opened 2026-09-08
Add desired strict tsconfig rules

angrykoala:
This PR adds the tsconfig settings that ideally would be enabled to add more strict checks in typescript.

Some of these are not enabled yet, and will be enabled in other PRs

Close LS-34

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -11,7 +11,9 @@
     "sourceMap": true,
     "target": "ESNext",
     "outDir": "dist",
-    "incremental": true
+    "incremental": true,
+    "strict": false,
+    "strictNullChecks": false
   },
   "exclude": ["node_modules", ".vscode-test"]
 }
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -12,7 +12,7 @@
     "target": "ESNext",
     "outDir": "dist",
     "incremental": true,
-    "strict": false,
+    "strict": true,
     "strictNullChecks": false
   },
   "exclude": ["node_modules", ".vscode-test"]
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -16,7 +16,7 @@
     "noUncheckedIndexedAccess": true,
     "noPropertyAccessFromIndexSignature": false, // TODO: set to true
     "noImplicitOverride": false, // TODO: set to true
-    "noImplicitReturns": false // TODO: set to true
+    "noImplicitReturns": true
   },
   "exclude": ["node_modules", ".vscode-test"]
 }
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -13,7 +13,7 @@
 
     "strictNullChecks": false, // TODO: remove this (default is true)
     "exactOptionalPropertyTypes": false, // TODO: set to true once strictNullChecks are enabled
-    "noUncheckedIndexedAccess": false, // TODO: set to true
+    "noUncheckedIndexedAccess": true,
     "noPropertyAccessFromIndexSignature": false, // TODO: set to true
     "noImplicitOverride": false, // TODO: set to true
     "noImplicitReturns": false // TODO: set to true
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -14,7 +14,6 @@
     "strictNullChecks": false, // TODO: remove this (default is true)
     "exactOptionalPropertyTypes": false, // TODO: set to true once strictNullChecks are enabled
     "noUncheckedIndexedAccess": true,
-    "noPropertyAccessFromIndexSignature": false, // TODO: set to true
     "noImplicitOverride": false, // TODO: set to true
     "noImplicitReturns": true
   },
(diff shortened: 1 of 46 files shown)


tsconfig.base.json
```diff
@@ -1,19 +1,21 @@
...
-    "esModuleInterop": true,
     "lib": ["dom", "ESNext"],
-    "module": "commonjs",
     "moduleResolution": "node",
-    "skipLibCheck": true,
     "sourceMap": true,
     "target": "ESNext",
     "outDir": "dist",
     "incremental": true,
-    "strict": true,
```
angrykoala, line 15: This rule is not needed explicitly, as it is already in the `recommended` rules

tsconfig.base.json
```diff
@@ -1,19 +1,21 @@
...
-    "strictNullChecks": false
+
+    "strictNullChecks": false, // TODO: remove this (default is true)
```
anderson4j, line 14: What's the reasoning for removing the comment without flipping the setting to true? (For all 3 removed TODOs)

angrykoala: The goal of this PR is to map out the desired ts config, some options required small enough changes to make the straight away, but the remaining ones require a lot of changes so my plan is to enable them incrementally in separate PRs to avoid too many conflicts and keep the diff easier to read

I also want to verify this is the set of rules we want before making lots of changes

anderson4j: Yeah not activating now makes sense, but removing the reminder to activate them looks like the TODO is done. Not a big deal now that I understand your reasoning
(Edit: I misread diff)

tsconfig.base.json
```diff
@@ -1,19 +1,21 @@
...
     "composite": true,
     "allowJs": false,
     "declaration": true,
-    "esModuleInterop": true,
     "lib": ["dom", "ESNext"],
-    "module": "commonjs",
     "moduleResolution": "node",
-    "skipLibCheck": true,
     "sourceMap": true,
     "target": "ESNext",
```
danieladugyan: Maybe something more stable?

tsconfig.base.json
```diff
@@ -1,19 +1,21 @@
 {
+  "extends": "@tsconfig/recommended/tsconfig.json",
   "compilerOptions": {
```
danieladugyan, line 3: I think the following options also seem nice:
- `verbatimModuleSyntax` – _"This option forces you to use import type and export type, leading to more predictable behavior and fewer unnecessary imports."_
- `declarationMap` – _"Tells TypeScript to emit declaration maps. It's needed so that when consumers of your libraries are debugging, they can jump to the original source code using go-to-definition."_

**Sources**
- https://www.totaltypescript.com/tsconfig-cheat-sheet
- https://www.typescriptlang.org/tsconfig

tsconfig.base.json
```diff
@@ -1,19 +1,21 @@
...
     "composite": true,
     "allowJs": false,
     "declaration": true,
-    "esModuleInterop": true,
     "lib": ["dom", "ESNext"],
-    "module": "commonjs",
     "moduleResolution": "node",
```
danieladugyan, line 8:
```suggestion
    "moduleResolution": "nodenext",
```

`"node"` seems outdated: _"`'node'` for Node.js versions older than v10, which only support CommonJS require. You probably won’t need to use `node10` in modern code."_

Along with this change, it might make sense to change `"module"`? The two options are connected, but I'm not so sure how it affects us. `nodenext` is recommended by both the [docs](https://www.typescriptlang.org/tsconfig/#module) and a [popular TSConfig cheat sheet](https://www.totaltypescript.com/tsconfig-cheat-sheet), but since we're using `esbuild` maybe `esnext` or `preserve` would be more appropriate? Not sure.
- _"`NodeNext` is the best option for Node."_
- _"You very likely want `nodenext` for modern Node.js projects and `preserve` or `esnext` for code that will be bundled."_

danieladugyan (commented):
I took a look at the tsconfig and I think it looks nice. I left a few comments, but I realized after the fact that most of the things I commented out are not really relevant to the strictness options. I've left the comments anyways in case we want to take this moment to change a few other options.

tsconfig.base.json
danieladugyan: I ended up having to change this in #746 to get rid of errors.

tsconfig.base.json
angrykoala: Mmm, I'm not sure we can easily add the `verbatimModuleSyntax` option at the moment, as it breaks packages using commonjs

To achieve the "types required" we could instead rely on oxlint `consistent-type-imports` https://oxc.rs/docs/guide/usage/linter/rules/typescript/consistent-type-imports.html

angrykoala: Regarding declarationMap, as you mention that is not really about strictness, and have implications on how we want to publish/bundle the different packages, so I'd rather keep it out of scope for this PR

anderson4j (approved):
Looks like I didnt approve last look. Here ya go! 

Having a look at the other stacked next

PR #756, opened 2026-09-09
Add api extractor to language-support package

anderson4j:
Adds dependency+conf+command+CI run for api extractor. 

Report is checked in to repo, and shows current api - will also show in the future if API changes in another PR, for example making it more clear if we modify an exported function (could be breaking for dependents)

Also gives warnings if we have "forgotten imports" like for this case:
_myFile.ts_
`
export const mySchema = {...}
export function myFunction(query: string, schema: mySchema )
`
_index.ts_
`export { 'myFunction' } from ./myFile` <- missing `mySchema` export to create input for `myFunction`
Can be worked around, but is a bit painful

When adding this we ran into the issue of the antlr parser being exported in its entirety. That meant that ALL generated classes were included in the API report, making it quite unreadable. It also meant that we included in "our api" generated classes based off grammar that we didnt control (since we fetch the grammar files from the neo4j monorepo). Since we are in pre-release and don't know of any projects relying on our parser specifically, I think the best path is to remove the export (we can always re-add it, or do some other workaround, if we get complaints)

Still we were not quite free from the export of generated classes. Because methods on our `CypherLanguageService` and our `parse()` method used/returned these classes, we still had "forgotten imports" for generated classes (who reference each other, creating a chain of new "forgotten imports"). To fix this we made these methods internal, and only export methods taking/returning non-generated classes.

Generated api-extractor config files are .json but use comments. Added some settings so vscode/github doesnt lint on this

I've only added the extractor to the `language-support` package so far. I figure it's nice to limit the intial PR to this so we can decide on the setup, and then do a follow-up for other published packages.

Closes LS-170


.vscode/settings.json
```diff
@@ -1,6 +1,7 @@
 {
-  "typescript.tsc.autoDetect": "off",
-  "typescript.preferences.quoteStyle": "single",
+  "files.associations": { "*.json": "jsonc" },
+  "js/ts.tsc.autoDetect": "off",
+  "js/ts.preferences.quoteStyle": "single",
```
anderson4j, line 4: the old versions were deprecated

packages/language-support/src/autocompletion/autocompletion.ts
```diff
@@ -16,13 +28,7 @@ export function autocomplete(
...
     manual = false,
     consoleCommandsEnabled = true,
-  }: {
-    symbolsInfo?: SymbolsInfo;
-    parsingResult?: ParsingResult;
-    caretPosition?: number;
-    manual?: boolean;
-    consoleCommandsEnabled?: boolean;
-  } = {},
+  }: AutocompleteOptions & { parsingResult?: ParsingResult } = {},
```
anderson4j, line 31: The ParsingResult part is what before "pulled in" generated grammar to public api

packages/language-support/src/statementAtCaret.ts
```diff
@@ -0,0 +1,32 @@
+import { createParsingResult } from './cypherLanguageService.js';
+
+/**
+ * Returns the text of the statement containing the caret.
+ *
+ * This lives here rather than in the consumer so that ParsingResult - which exposes
+ * the generated ANTLR parser - does not need to be part of this package's public API.
+ */
+export function getStatementAtCaret(
```
anderson4j, line 9: Method that used to live in `vscode-extension`, only case that made us need `createParsingResult` export (for our own packages)

anderson4j:
We have exports of the other language-support functions independent of the `CypherLanguageService` class since before. I noticed we didnt for hover - maybe we should?

packages/language-support/src/debugTree.ts
```diff
@@ -0,0 +1,32 @@
...
+  children?: SimpleTree[];
+}
+
+/**
+ * Builds a plain, serialisable view of the ANTLR parse tree for a query.
+ *
+ * This lives here rather than in the consumer so that the generated parser does not
+ * need to be part of this package's public API.
+ */
+export function getDebugTree(cypher: string): SimpleTree {
```
anderson4j, line 16: Used to live in codemirror-playground - the one separate package we had that used the exported parser. Moved here so we can kill the generated class exports

.changeset/tired-owls-walk.md
```diff
@@ -0,0 +1,6 @@
+---
+"@neo4j-cypher/language-support": patch
+"@neo4j-cypher/react-codemirror": patch
```
anderson4j, line 3: included react-codemirror since it re-exports all of language-support

.gitattributes
```diff
@@ -1,2 +1,3 @@
+*.json  linguist-language=JSON-with-Comments
```
danieladugyan, line 1: Cool! I didn't know you could do that

packages/react-codemirror/package.json
```diff
@@ -61,15 +61,15 @@
...
+    "@playwright/experimental-ct-react": "1.62.1",
+    "@playwright/test": "1.62.1",
     "@types/lodash.debounce": "^4.0.9",
     "@types/react": "^18.0.28",
     "@types/react-dom": "^18.0.11",
     "@vitejs/plugin-react": "^6.1.1",
     "copyfiles": "^2.4.1",
     "jsdom": "^24.1.1",
-    "playwright": "^1.62.1",
+    "playwright": "1.62.1",
```
danieladugyan, line 72: question: Why the pinned playwright versions?

api-extractor-base.json
```diff
@@ -0,0 +1,429 @@
...
+  "mainEntryPointFilePath": "<unscopedPackageName>",
+
+  /**
+   * Specifies what type of newlines API Extractor should use when writing output files.  By default, the output files
+   * will be written with Windows-style newlines.  To use POSIX-style newlines, specify "lf" instead.
+   * To use the OS's default newline kind, specify "os".
+   *
+   * DEFAULT VALUE: "crlf"
+   */
+  // "newlineKind": "crlf",
```
danieladugyan, line 27:
```suggestion
  // "newlineKind": "lf",
```

nitpick: All other text files in the repo have `LF` newlines.

api-extractor-base.json
```diff
@@ -0,0 +1,429 @@
...
+     * Note: This setting will be ignored if "overrideTsconfig" is used.
+     *
+     * SUPPORTED TOKENS: <projectFolder>, <packageName>, <unscopedPackageName>
+     * DEFAULT VALUE: "<projectFolder>/tsconfig.json"
+     */
+    // Must use <projectFolder>: inherited relative paths resolve against the file that
+    // declares them (this one, at the repo root), and the root tsconfig.json is a
+    // solution file with no compilerOptions. Reading that made the analysis default to
+    // target ES5 + skipLibCheck false, producing spurious TS18028 warnings about our
+    // dependencies' .d.ts files.
```
danieladugyan: question: The setting makes sense to me, but not sure if I understand the comment.

packages/language-support/api-extractor.json
danieladugyan, line 1: nitpick: Maybe make this less file verbose by only having the options that override the base config (and perhaps without the docs)? i.e:

```json
{
  "mainEntryPointFilePath": "./dist/esm/project/language-support/src/index.d.ts",
  "docModel": {
    "projectFolderUrl": "https://github.com/neo4j/cypher-language-support/tree/main/packages/language-support"
  },
  "messages": {
    "extractorMessageReporting": {
      "ae-forgotten-export": {
        "logLevel": "warning",
        "addToApiReportFile": true
      }
    }
  }
}
```

packages/language-support/package.json
```diff
@@ -44,14 +44,16 @@
...
+    },
+    "./grammar/*.g4": "./src/antlr-grammar/*.g4"
   },
   "scripts": {
     "gen-parser": "antlr-ng -Dlanguage=TypeScript --generate-visitor --lib src/antlr-grammar -o src/generated-parser --exact-output-dir -- src/antlr-grammar/CypherCmdLexer.g4 && antlr-ng -Dlanguage=TypeScript --generate-visitor --lib src/antlr-grammar -o src/generated-parser --exact-output-dir -- src/antlr-grammar/CypherCmdParser.g4",
     "build": "pnpm gen-parser && concurrently 'npm:build-esm-and-types' 'npm:build-commonjs' 'npm:build-cypherfmt-cli'",
     "build-esm-and-types": "tsc --outDir dist/esm/project/language-support && cp src/syntaxValidation/semanticAnalysis.js dist/esm/project/language-support/src/syntaxValidation/semanticAnalysis.js",
     "build-commonjs": "esbuild ./src/index.ts --bundle --format=cjs --sourcemap --outfile=dist/cjs/index.cjs",
     "build-cypherfmt-cli": "esbuild ./src/formatting/cli.ts --bundle --format=esm --sourcemap --outfile=dist/esm/formatting/cli.mjs --external:fs --external:path",
+    "check-api-report": "pnpm build-esm-and-types && tsx checkApiReport.ts",
```
danieladugyan, line 56:
```suggestion
    "check-api-report": "tsx checkApiReport.ts",
```

suggestion: In CI, this ends up calling build-esm-and-types an extra time since it already runs in the setup and build step. But maybe that's preferrable, not sure.

packages/language-support/src/autocompletion/autocompletion.ts
danieladugyan: optional: I was thinking it could perhaps be a bit neater to keep this file exactly as is, and then re-export as follows:

```ts
// index.ts
import { autocomplete as autocompleteWithParsingResult } from './autocompletion/autocompletion.js';
export function autocomplete(
  query: string,
  dbSchema: DbSchema,
  options: AutocompleteOptions = {},
): CompletionItem[] {
  return autocompleteWithParsingResult(query, dbSchema, options);
}
```

You could also avoid the extra function call by making it a `const` instead:

```ts
// index.ts
import { autocomplete as autocompleteWithParsingResult } from './autocompletion/autocompletion.js';
export const autocomplete: (
  query: string,
  dbSchema: DbSchema,
  options?: AutocompleteOptions,
) => CompletionItem[] = autocompleteWithParsingResult;
```

(the same goes for the other `*withParsingResult`)

api-extractor-base.json
danieladugyan, line 1: The settings makes sense to me, but not sure if I understand the added comments.

danieladugyan (approved):
LGTM! But there's quite a lot of changes so might be good if someone else takes a look too.

danieladugyan:
suggestion: [The docs](https://api-extractor.com/pages/setup/invoking/#1-configure-the-typescript-compiler-for-your-project) recommend setting `"declarationMap": true` in our `tsconfig.json`.

packages/react-codemirror/package.json
anderson4j: Ah I forgot to comment! I had merge conflicts in pnpm.lock and as a "lazy fix" I thought id delete it and reinstall with `pnpm i`. Turns out this is different from running `pnpm i` with existing lockfile (which apparently doesnt update) - with no lockfile it fetches latest compatible with versions like `^1.62.1` - which in this case was now 1.63.x for 2 of the packages, while one still hadnt recieved the update, so 1.62.1 was still latest.

The versions then got out of sync, which caused errors in our testfiles where the dependencies interacted. Figure the safe way to avoid this is pinned versions. It's also just in devDependencies, so we wont have the "double dependency" issue you can get when exporting a library with pinned dependencies.

packages/language-support/package.json
anderson4j: mm maybe we should split it so we have the existing command to run manually (so you dont accidentaly generate report for last built state), and what you recommend on ci

packages/language-support/src/autocompletion/autocompletion.ts
anderson4j: Im not used to seeing this "implementation in index.ts" style stuff. Looks pretty clean, but maybe makes the "export file" a bit harder to read...

What do you think @angrykoala - worth doing?

packages/language-support/api-extractor.json
anderson4j: Yeah I was wondering about what's nicest here. On the one hand this way is a lot more clean, on the other hand, how it is (with all settings+docs) makes it easier to see what settings are available and what they're for without switching back and forth from the base file

oxlint.config.ts
```diff
@@ -43,5 +45,12 @@ export default defineConfig({
         'vitest/no-conditional-expect': 'off',
       },
     },
+    {
+      // Build scripts are CLIs: stdout is their output, not stray debugging.
```
angrykoala, line 49: What does `stray debugging` means?

We could rephrase this a bit, just mentioning that console is expected for cli tools in scripts

packages/language-support/src/autocompletion/autocompletion.ts
angrykoala: It is not necessarily a problem to have implementation in `index.ts`, but this index in particular is already a bit loaded with barrel exports so it may be a bit confusing to mix implementation vs export list in this instance.

I feel that the underlying issue here is that `autocompleteWithParsingResult` is used both in the `autocomplete` function exported in this file (and project), and in the  `autocomplete` method in `CypherLanguageService`, (which is very similar) making the surface of this part a bit unclear at the moment.  

As it is now, I would keep `autocomplete` in this file, but I don't have a strong opinion if you prefer to move it to index.ts

As a (small) sidenote declaring autocomplete as a variable with an arrow function doesn't really avoid the extra function, it is just a different syntax to declare it (which I'm not that fond of, but that is more style than anything in this case)

scripts/runApiExtractor.mts
```diff
@@ -0,0 +1,64 @@
+import { Extractor, ExtractorConfig } from '@microsoft/api-extractor';
```
angrykoala, line 1: Do we have a convention for `.mts` files vs `ts` or is there a reason it needs to be an `mts`?

packages/language-support/checkApiReport.ts
```diff
@@ -0,0 +1,4 @@
+import { runApiExtractor } from '../../scripts/runApiExtractor.mjs';
+
+// Deliberately outside src/ so it is not part of the package's tsc build.
+runApiExtractor(import.meta.dirname);
```
angrykoala, line 4: Are we intending to use `runApiExtractor` somewhere else, if we are not planning to in the short term, I would consider just putting it here, to avoid polluting the global repo

packages/language-support/src/cypherLanguageService.ts
```diff
@@ -43,7 +43,7 @@ import {
   splitIntoStatements,
 } from './helpers.js';
 import {
-  lintCypherQuery,
+  lintCypherQueryWithParsingResult,
```
angrykoala, line 46: mmm, I see this pattern of having 2, very similar functions, one for the public API and another for the internal, we may want to discuss on a convention or pattern for this, to keep this consistent in a more general way

.changeset/tired-owls-walk.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Modify exports for api-extractor in language-support. Adds "forgotten exports" and most importantly removes export of generated antlr classes (or functions referencing these). Antlr grammar is now exported as .g4 files instead - see readme for instructions on how to generate parser/lexer.
```
angrykoala, line 6: Maybe we could rephrase this changeset?

Users of the library probably don't have the context on what api-extractor is, and in this case it may be useful to list the added/removed things, as they affect the public surface of the library. Technically it is a breaking change and we may want to make the release minor instead of patch

scripts/runApiExtractor.mts
```diff
@@ -0,0 +1,64 @@
...
+import * as path from 'node:path';
+
+/**
+ * Runs API Extractor for a single workspace package.
+ *
+ * Locally this rewrites the committed API report, so a change to the public API shows
+ * up as a diff to review. In CI (or with --ci) the report is only verified: a stale
+ * report fails the build instead of being silently rewritten.
+ */
+export function runApiExtractor(packageFolder: string): void {
```
angrykoala, line 12: I'm not sure why do we need to run this as a build script, is there a limitation on the cli?

api-extractor-base.json
```diff
@@ -0,0 +1,429 @@
+/**
+ * Shared Config file for API Extractor.  For more info, please visit: https://api-extractor.com
+ */
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
```
angrykoala, line 5: Do we need a base + non-base file here? at the moment is only used in one package right?

api-extractor-base.json
```diff
@@ -0,0 +1,429 @@
...
+   * The file extension must be a declaration file (e.g. ".d.ts", ".d.mts", ".d.{extension}.ts"), not ".ts".
+   *
+   * The path is resolved relative to the folder of the config file that contains the setting; to change this,
+   * prepend a folder token such as "<projectFolder>".
+   *
+   * SUPPORTED TOKENS: <projectFolder>, <packageName>, <unscopedPackageName>
+   */
+  "mainEntryPointFilePath": "<unscopedPackageName>",
+
+  /**
```
angrykoala, line 20: I'm not sure how helpful the comments are tbh, it makes it hard to see what the config is, and most relevant information seems to already be in the schema `https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json`

angrykoala (changes requested):

oxlint.config.ts
anderson4j: People making stray `console.log` statements when debugging and forgetting to remove them is the reason we have the rule originally i figure

packages/language-support/checkApiReport.ts
anderson4j: Yeah I figure once we find an approach we're happy with here, we should use api-extractor for all the packages we publish. 

I made this PR just for language-support since I figure the config and logic will be very similar in the other packages, and changes already became quite large with just lang-supp. There's already a linear issue for the follow-up

api-extractor-base.json
anderson4j: atm yes, but the intent is to make a follow-up for the other libraries. Figured this is how we will want this package in the end anyway, so it felt counterintuitive to make a solution I would soon overwrite.

angrykoala (approved):

scripts/runApiExtractor.ts
```diff
@@ -0,0 +1,64 @@
+import { Extractor, ExtractorConfig } from '@microsoft/api-extractor';
```
anderson4j: It uses ESM-style imports, but we had no `"type": "modules"` in the root package.json. I think we'd want to use that in general though, so I moved to ts + `"type": "modules"` instead

PR #757, opened 2026-09-10
Brow 380 editor action layout

alisasanib:
The action buttons were positioned by measuring from the spacer widget inside the first line, so anything rendered above it , such as an inline panel or deleted lines, pushed them down the page, and horizontal overflow pushed them outside the editor's right edge. They're now measured against the scroller's visible box, so they stay in the corner. 
This issue was reported here:

https://github.com/neo4j/upx/pull/12530#issuecomment-5618029363

### DIFF
--- a/.changeset/nervous-poems-argue.md
+++ b/.changeset/nervous-poems-argue.md
@@ -1,0 +1,5 @@
+---
+'@neo4j-cypher/react-codemirror': patch
+---
+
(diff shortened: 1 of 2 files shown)


nnecla (commented):
Looks good for the inline panel alignment. While testing, unfortunately, I've realised that the select all of the query does not highlight the first line, even if it is selected. I'll fix it and approve this afterwards!

Image

nnecla (approved):

PR #758, opened 2026-09-11
Upgrade Needle packages to v4

danieladugyan:
NDL 4.x lists React 19 as a peer. GraphVisualization moved to @neo4j-ndl/react-graph, so webviews and the CodeMirror token copy are updated for the v4 APIs.


patches/@neo4j-ndl__base@4.19.0.patch
danieladugyan: Running `pnpm test:e2e` in `react-codemirror` causes the following error:

```
Error: Cannot find module '/repo/node_modules/.pnpm/@neo4j-ndl+base@4.19.0/node_modules/@neo4j-ndl/base/lib/esm/typings' imported from /repo/node_modules/.pnpm/@neo4j-ndl+base@4.19.0/node_modules/@neo4j-ndl/base/lib/esm/index.js
    at eval (<anonymous>:1:1)

To open last HTML report run:

  pnpm exec playwright show-report

 ELIFECYCLE  Command failed with exit code 1.
```

I spent a LONG time investigating why, but I'm still not quite sure. I was unable to reproduce the error in a standalone repro. Nevertheless, I'm fairly certain that Needle's ESM build is broken since it's missing file extensions. That's fine as long as you're using a bundler, but crashes when using Node's ESM imports (that part is easy to reproduce: `node --input-type=module -e "import('@neo4j-ndl/base')"`). I'm guessing no one is using Needle without a bundler so it's not usually a big deal.

danieladugyan, line 1: Running `pnpm test:e2e` in `react-codemirror` causes the following error:

```
Error: Cannot find module '/repo/node_modules/.pnpm/@neo4j-ndl+base@4.19.0/node_modules/@neo4j-ndl/base/lib/esm/typings' imported from /repo/node_modules/.pnpm/@neo4j-ndl+base@4.19.0/node_modules/@neo4j-ndl/base/lib/esm/index.js
    at eval (<anonymous>:1:1)

To open last HTML report run:

  pnpm exec playwright show-report

 ELIFECYCLE  Command failed with exit code 1.
```

I spent a LONG time investigating why Playwright causes this error, but I'm still not quite sure – I was unable to reproduce it. Nevertheless, I'm fairly certain that Needle's ESM build is broken since it doesn't adhere to the spec. Patching the package fixes the error if we want to be able to merge before it's fixed upstream.

PR #760, opened 2026-09-15
Introduce hover for variables and restructure hover

anderson4j:
Closes LS-133

New feature - hovering on variables now shows their type and labels if node/rel.

Ex.
image

While doing this I also saw that our initial handling of hover was a bit different from other language support functionality. In linting or autocomplete we return `SyntaxDiagnostic` (=`Diagnostic & { offsets: { start: number; end: number } }` -where `Diagnostic` is from `vscode-languageserver-type`) or for completions a `CompletionItem` type implementing the `vscode-languageserver-type` `CompletionItem`. 

Hover on the other hand returned `SignatureHoverInfo`-objects which were instead converted to a `Hover`-implementing (`vscode-languageserver-type` type) object in the server. This PR brings hover in line with the other features.

Also fixed when function hover is triggered. Before it would trigger for the entire function call, now it only does it on the name, like in other language support.

Does not show hover info for properties. These are in the symbol table when used like `n.Age` but parsing (and thus symbol table) does not know type of a property (strangely even in cases like `MATCH (n:Person) WHERE n.age IS :: INTEGER RETURN n.age` we get "could be any type")


packages/language-server/src/hoverInfo.ts
```diff
@@ -1,38 +1,20 @@
-import { SignatureHoverInfo } from '@neo4j-cypher/language-support';
+import { Hover, HoverParams, TextDocuments } from 'vscode-languageserver';
+import { TextDocument } from 'vscode-languageserver-textdocument';
+import { languageService } from './server.js';
+import { Neo4jSchemaPoller } from '@neo4j-cypher/query-tools';
 
-export function createParametersHoverString(
```
anderson4j, line 3: These 2 created the data for the `Hover`-implementing objects mentioned in the description. Moved to `hover.ts` in language-support in the shift to language-support returning `Hover` objects

packages/language-support/src/hoverInformation/hover.ts
anderson4j, line 1: Moved into it's own folder much like other language support functionality - but also modified to contain variable hover and full signature-> `Hover` logic

packages/language-support/src/hoverInformation/variableHover.ts
```diff
@@ -0,0 +1,65 @@
...
+export function findVariableOnCaret({
+  parsingResult,
+  caretPosition,
+  symbolsInfo,
+}: {
+  parsingResult: ParsingResult;
+  caretPosition: number;
+  symbolsInfo: SymbolsInfo;
+}): Symbol | undefined {
+  const caret = findCaret(parsingResult, caretPosition);
```
danieladugyan, line 46: > _The symbol table is computed asynchronously and can lag behind the document actually being hovered (e.g. the user kept typing while the previous query's lint job was still running). Token offsets are only meaningful relative to the text they were computed from, so matching them against a table built for a different query could silently describe the wrong variable._
> – Claude

```suggestion
  if (symbolsInfo.query !== parsingResult.query) {
    return undefined;
  }

  const caret = findCaret(parsingResult, caretPosition);
```
```ts
test('Does not use a stale symbol table computed for a different query', () => {
  const staleQuery = 'MATCH (m) RETURN m';
  const { symbolTables } = languageService.lint(staleQuery, dbSchema);
  languageService.setSymbolsInfo({
    query: staleQuery,
    symbolTables: symbolTables,
  });

  const currentQuery = 'MATCH (n) RETURN n';
  const hoverInfo = languageService.hoverInfo(currentQuery, {
    caretPosition: 'MATCH (n) RETURN '.length,
    dbSchema,
  });

  expect(hoverInfo).toEqual(undefined);
});
```

packages/language-support/src/hoverInformation/variableHover.ts
```diff
@@ -0,0 +1,65 @@
...
+    return undefined;
+  }
+
+  for (const symbolTable of symbolsInfo.symbolTables) {
+    for (const symbol of symbolTable) {
+      if (symbol.references.includes(caret.token.start)) {
+        return symbol;
+      }
+    }
+  }
```
danieladugyan, line 78: > _This is O(total number of variable-reference occurrences in the entire query) per call — not per edit, per hover request. There's no early structure to make the lookup faster; it's a brute-force scan of the whole symbol model every time. The data being scanned — `symbolsInfo.symbolTables` — is invariant across hover calls. It only changes when a new symbol table is installed via `setSymbolsInfo`, which happens at most once per completed async lint job, i.e. at most once per edit, often less. Hover requests, by contrast, fire far more often than edits._
> – Claude

```suggestion
  return getReferencesByOffset(symbolsInfo).get(caret.token.start);
```
```ts
/* symbolsInfo is replaced wholesale (a new object) whenever the symbol table
   is recomputed, so indexing it by object identity gives us a cache that is
   automatically invalidated on the next `setSymbolsInfo` call, without any
   manual bookkeeping. */
const referencesByOffset = new WeakMap<SymbolsInfo, Map<number, Symbol>>();

function getReferencesByOffset(symbolsInfo: SymbolsInfo): Map<number, Symbol> {
  const cached = referencesByOffset.get(symbolsInfo);
  if (cached) {
    return cached;
  }

  const index = new Map<number, Symbol>();
  for (const symbolTable of symbolsInfo.symbolTables) {
    for (const symbol of symbolTable) {
      for (const reference of symbol.references) {
        index.set(reference, symbol);
      }
    }
  }
  referencesByOffset.set(symbolsInfo, index);
  return index;
}
```

danieladugyan (approved):
LGTM! Left a couple of suggestions from Claude that make sense to me.

packages/language-support/src/hoverInformation/variableHover.ts
anderson4j: Like mentioned in standup, for completions we also faced this issue. We reasoned that the best we could do to avoid waiting on the sometimes slow semantic analysis, was to use symbol tables from previous queries. Since we usually edit a query by typing new symbols into the old one, (except the copy-paste case) the "slightly outdated query"-symbol table is often mostly the same as the current.

packages/language-support/src/hoverInformation/variableHover.ts
anderson4j: accepted, with a small rename

PR #762, opened 2026-09-16
enable consistent-type-imports in oxlint

angrykoala:
Follow up from a discussion from #753, this seem to be a desired rule in oxlint that requires importing using `type` where possible


danieladugyan (approved):

packages/query-tools/src/integration-tests/setupTestContainer.ts
```diff
@@ -1,4 +1,5 @@
-import { Neo4jContainer, StartedNeo4jContainer } from '@testcontainers/neo4j';
+import type { StartedNeo4jContainer } from '@testcontainers/neo4j';
+import { Neo4jContainer } from '@testcontainers/neo4j';
```
anderson4j, line 2: Could use imports like this for imports where we have both type and non-type imports. Think in cases where the type imports are few it's a bit more clean, but probably not if you have a bunch.
```suggestion
import {
  Neo4jContainer,
  type StartedNeo4jContainer,
} from '@testcontainers/neo4j';
```

anderson4j (approved):
I think it looks a bit ugly when we have imports that are from the same file but are now split up. Suggested an alternative, but I could see it not being very viable for files with many type imports. Think most cases it will be nicer though - I'll leave it to you if it's worth changing and approve for now

packages/query-tools/src/integration-tests/setupTestContainer.ts
angrykoala: It is possible to change this with the option `https://oxc.rs/docs/guide/usage/linter/rules/typescript/consistent-type-imports.html#examples-with-fixstyle-inline-type-imports`. I don't have a strong opinion, but I slightly prefer having them separate (plus, seems to be the default)

PR #764, opened 2026-09-18
Prism codemirror demo fix

anderson4j:
Adds a fix to ensure prism (missing on demo built from main) is imported correctly after vite 8 update.

Also adds a smoke-test that would fail without the fix to detect similar issues in the future

closes LS-224


packages/react-codemirror-playground/package.json
```diff
@@ -14,6 +14,7 @@
     "dev": "vite --open",
     "build": "tsc --noEmit && vite build",
     "preview": "vite preview",
+    "test:e2e": "npx playwright test --trace on",
```
danieladugyan, line 17: nitpick: for consistency with `react-codemirror` package
```suggestion
    "test:e2e": "playwright test --trace on",
```

.changeset/rude-birds-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"@neo4j-cypher/react-codemirror-playground": patch
```
danieladugyan, line 2: Shouldn't we bump react-codemirror too? Feels like since the fix was in made this package, it's probably broken right now and in need of an update, even if we've only noticed it in the playground package.
```suggestion
"@neo4j-cypher/react-codemirror": patch
"@neo4j-cypher/react-codemirror-playground": patch
```

packages/react-codemirror/src/lang-cypher/prism-global.ts
```diff
@@ -0,0 +1,7 @@
+import Prism from 'prismjs';
+
+// `prismjs/components/prism-cypher` (used in parser-adapter) is a plain script that mutates the global
+// `Prism`. Bundlers keep prismjs' CJS core lazy and may evaluate the component
+// before the core has run, so force the core to initialise (and publish the
+// global) from a module that is imported first.
```
danieladugyan, line 6: I didn't quite understand this comment, but after doing some more research I now understand why this file is needed.

danieladugyan (approved):

